PR #148, opened 2014-07-08
Script to get nl_NL address.countries

mdxs:
As discussed in #128, here is an initial tool/script to obtain the country names as used in `faker/providers/nl_NL/address.py` (is localized)


fcurella:
not applicable anymore

PR #149, opened 2014-07-08
Script to get date_time.countries

mdxs:
Script to obtain a tuple/list of dictionary entries representing countries: with ISO 3166-1 alpha-2 code, English country name, and a list of (commonly used) timezones per country. As obtained from (latest) `pytz`, which needs to be locally installed using `pip install pytz` in development environment.

It doesn't get the `continent` nor `capital` portions seen in current `faker/providers/date_time.py` countries, but these are probably not used anyway.

### DIFF
--- a/tools/providers/get_date_time_data.py
+++ b/tools/providers/get_date_time_data.py
@@ -1,0 +1,83 @@
+# coding=utf-8
+
+"""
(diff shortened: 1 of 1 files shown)


fcurella:
not applicable anymore

PR #169, opened 2014-10-03
slovak faker: company, internet

viktormaruna:


fcurella:
@viktormaruna could you convert this to localized providers?

PR #177, opened 2014-11-10
Remove duplicate words from Lorem provider

jeffwidman:
The docs say:
> Each call to method faker.name() yealds a different (random) result. This is because faker uses __getattr__ magic, and forwards faker.Genarator.method_name()' calls tofaker.Generator.format(method_name)`.

So I assumed that by "different" it meant the words provider would yield unique results, and just wasted a bunch of time trying to debug my code when I kept hitting duplicate key constraints in my database. 

Updated the words list to be de-duplicated and sorted.


fcurella:
The lorem provider is working as intended. It generates a different text every time. Feel free to file a feature request or submit a PR for a WordProvider.

jeffwidman:
>It generates a different text every time. 

It does not generate different text *every* time. It returns random results, but because the lorem provider includes a number of duplicate words, sometimes those results are the same. The sample data I was filling was about 30 items--plenty of room for uniqueness--and yet I found about 5% of the time two of the values were duplicated.

jeffwidman:
@fcurella Would you be open to re-evaluating this? It's still an annoyance... the Lorem provider has multiple duplicate words, and so calling the generator will return a duplicate value long before the generator runs out of potential values. All my PR does is take the Lorem words and ran them through a set function so that all the values available to the generator are unique.

fcurella:
@jeffwidman Looks like I underestimated the impact of this bug, so I'm ok with your PR. Do you mind rebasing against master?

fcurella:
@jeffwidman I've merged this in for you, thank you very much fro your contribution! :sparkles:

jeffwidman:
Thanks! I really appreciate it.

Issue #150, opened 2014-07-09
fake.date_time_this_month()


soobrosa:
All date_time fakers generate within the last given time period (now month), not in 'this' time period (aka current month.)

justinwp:
it makes sense to me that 

```
fake.date_time_this_month()
```
would include potentially all days in the current month, some options...
```
fake.date_time_this_month(after_today=True)
fake.date_time_this_month(before_today=True)
```

justinwp:
For year, decade, etc. could have the following:

```
fake.date_time_last_year() # better name for current implementation which is the last 12 months
fake.date_time_this_year(before_now=True, after_now=True) # current calendar year
fake.date_time_next_year() # next 12 months
```

PR #196, opened 2015-01-20
Date Time This [Period] Only Returns DateTime in Period

justinwp:
It makes sense that a function named 'date_time_this_year' should return a DateTime in **this** year. 

##### Previous
```
fake.date_time_this_year() # any DateTime in the last 12 months 
```
##### Now
```
fake.date_time_this_year() # any DateTime in the current year
fake.date_time_between(start_date='-1y', end_date='now') # same as prior functionality
fake.date_time_this_year(after_now=False) # any DateTime in the current year before datetime.now()
fake.date_time_this_year(before_now=False) # any DateTime in the current year after datetime.now()
```
#150

### DIFF
--- a/faker/providers/date_time.py
+++ b/faker/providers/date_time.py
@@ -321,33 +321,112 @@ class Provider(BaseProvider):
         return datetime.fromtimestamp(timestamp)
 
     @classmethod
-    def date_time_this_century(cls):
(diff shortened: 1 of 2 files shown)


justinwp:
failed to install a package with pip for 2.7... everything else passed

PR #204, opened 2015-02-22
Faker is not actually zip safe.

DRMacIver:
The code in the [loading](https://github.com/joke2k/faker/blob/master/faker/utils/loading.py) module will error if called when Faker is installed as a zip file because it's trying to look in a directory that doesn't exist.

I noticed this because faker works when installed from pip but if I do setup.py install on a package with faker as a dependency it will install in a broken state that errors when I try to load the module the first time.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,9 +12,6 @@ NEWS = io.open(os.path.join(here, 'CHANGELOG.rst'), encoding="utf8").read()
 
 version = '0.5.0'
 
-on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
-zip_safe = not on_rtd
-
 setup(
     name='fake-factory',
     version=version,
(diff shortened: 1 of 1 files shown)


fcurella:
Thanks for the report @DRMacIver

fcurella:
@joke2k can we release a new version on PyPI?

joke2k:
@fcurella yes, tomorrow ill do that

DRMacIver:
Yay. Thank you. :-)

fcurella:
@joke2k ping

DRMacIver:
Any news?

fcurella:
@joke2k are you available tomorrow?

joke2k:
yes, ill see u on irc tomorrow

PR #205, opened 2015-03-04
Updates person names for es_ES (male, female)

dnmellen:
We needed separated names for males and females but in `es_ES` locale all names were together, so I split male and female names using the most common names for spanish people.


fcurella:
Thanks!

PR #207, opened 2015-03-10
Update person names for ru_RU (male, female)

mousebaiker:
Russian language requires separate names for male and female. I split the names which were bunched up together and added common female names and prefixies.


fcurella:
thank you!

PR #211, opened 2015-03-13
Add requirements file + py2.6 handled in travis

gtnx:
Handle issue #211

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ branches:
   only:
     - master
 python:
+  - 2.6
   - 2.7
   - 3.3
   - 3.4
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,1 @@
+importlib
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,0 @@
-importlib
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@
 
 import os
 import io
+import sys
 from setuptools import setup, find_packages
 
 here = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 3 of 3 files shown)


fcurella:
wouldn't this install the `importlib` package also for python 2.7+? Or does `importlib`'s own `setup.py` switch to a no-op in that case?

gtnx:
There's a test on the version in the [setup.py](https://github.com/brettcannon/importlib/blob/master/setup.py)

fcurella:
Thanks @gtnx ,

I just checked importlib under py2.7. Looks like it doesn't actually install any python code under those circumstances, but I still don't like having `pip` fetch it for nothing.

I'm ok including it as a requirement, but I'd rather have a list declared in `setup.py` rather than use a `requirements.txt`.

Can you amend your PR to have `setup.py` check the python version and build a requirement list accordingly?

gtnx:
Are you sure?
In my opinion, it makes more sense to let the dependency package (ie importlib) decide what to install. 
Plus, having a requirements.txt file is a common and helpful practice since you just need to do `pip install -r requirements.txt` for setting up your environment.

fcurella:
Yeah, I like the idea of keeping other people's build fast by not fetching. I don't see why let them incur a cost that can be avoided.

Re: `requirements.txt`, my opinion (mostly taken from my interpretation of https://caremad.io/2013/07/setup-vs-requirement/) is that `requirements.txt` is for deployments (eg: a website, or an app), while `install_requires` is for libraries. You can still setup the environment by running `pip install -e .`

gtnx:
Ok. done

fcurella:
Looks good! Thanks!

Just waiting for the CI to pass and I'll merge it.

gtnx:
I can't figure out why the CI failed on appveyor. I don't think it's related to my modifications

PR #212, opened 2015-03-18
Make faker zipsafe.

kecaps:
For loading.find_available_locales, use pkgutil.iter_modules (which works with the zipimporter) to locate the locale sub-modules.

### DIFF
--- a/faker/utils/loading.py
+++ b/faker/utils/loading.py
@@ -1,16 +1,13 @@
 import os
 from importlib import import_module
+import pkgutil
 
 
 def list_module(module):
     path = os.path.dirname(module.__file__)
-    try:
-        import pkgutil
-        modules = [name for finder, name, is_pkg in pkgutil.iter_modules([path]) if is_pkg]
-        if len(modules) > 0:
-            return modules
-    except (ImportError, AttributeError):
-        pass
+    modules = [name for finder, name, is_pkg in pkgutil.iter_modules([path]) if is_pkg]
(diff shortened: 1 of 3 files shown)


faker/utils/loading.py
```diff
@@ -4,6 +4,13 @@
 
 def list_module(module):
     path = os.path.dirname(module.__file__)
+    try:
+        import pkgutil
+        modules = [name for finder, name, is_pkg in pkgutil.iter_modules([path]) if is_pkg]
+        if len(modules) > 0:
+            return modules
+    except (ImportError, AttributeError):
+        pass
```
fcurella: I'd rather avoid an `except ...: pass` pattern. Does `pkgutil.iter_modules` work for non-zip packages as well? If it does and it's available on all Python versions that we support, I'd rather just use that as the one single way to list modules.

kecaps: pkgutil.iter_modules does work for non-zip packages.  But there does seem
to be a bug with it in python 3.2.5 where the bootstrapping File Loader is
missing an implementation. So, the fallback code is still required to get
the correct behavior on python 3.2.5 when using the standard file loader.

domtes:
+1

faker/utils/loading.py
fcurella: I don't think we necessarily care for Python 3.2, but since the fallabck is just one line of code, LGTM

fcurella:
Thanks!

PR #213, opened 2015-03-18
Fix gender-specific prefix/suffix handling

akx:


faker/tests/__init__.py
```diff
@@ -153,16 +159,27 @@ def test_magic_call_calls_format_with_arguments(self):
 
     def test_documentor(self):
         from faker.cli import print_doc
-        print_doc()
-        print_doc('address')
-        print_doc('faker.providers.it_IT.person')
+        from StringIO import StringIO
```
fcurella: you can move this import to the top of the file so it doesn't have to be imported twice

fcurella:
Good Work! Thanks!

PR #215, opened 2015-03-30
Add localized providers ja_JP

ta2xeo:
I added ja_JP(Japanese) provider, and tests.


fcurella:
Nice Work! Domo arigato gozaimasu :)

PR #216, opened 2015-03-30
Added Full Language-Country Name

GheloAce:
Added the full language-country name of the Localized Provider list section.  Good for those who do not know what do those ISO codes mean.


fcurella:
Thanks! Very much appreciated

PR #228, opened 2015-04-16
Update __init__.py

GLMeece:
Changed the following state names to correctly have spaces in them:
* New Hampshire
* New Jersey
* New Mexico
* New York
* North Carolina
* North Dakota
* Rhode Island
* South Carolina
* South Dakota
* West Virginia


fcurella:
Thank you very much! 🇺🇸

PR #231, opened 2015-04-24
Added timedelta to date_time provider

Brobin:
As the title states, this pull adds a method `time_delta` to the date_time provider.

This method generates a random timedelta in seconds from 0 to the current time.


fcurella:
Looks good! Thanks!

Brobin:
Wow, that was quick! I'm glad I could help.

PR #233, opened 2015-04-25
Added file_extension and file_name methods to file provider

Brobin:
This pull request adds the `file_name` and `file_extenstion` methods to the file provider. It adds tuples of common file extensions and a dictionary of the categories. This allows the user to either select a category or a random one will be chosen. 

The `file_name` method uses the `lorem.Provider.word()` method to get a filename, and joins that with a random extension. Additionally, the `file_name` method can take an extenstion to use. If none is provided, a random one will be chosen.


fcurella:
Looks very useful. Thank you!

PR #234, opened 2015-04-26
Update ja_JP address.

ohbarye:
Changed address formats.

Added the following method:

 - prefecture, city, town, chome, ban, gou, building_name, zipcode


fcurella:
Thank you!

PR #252, opened 2015-06-01
Adds uuid support

jaredculp:
Simple wrapper around the built in python library because I was a bit surprised that the following code didn't work.

```python
from faker import Factory

fake = Factory.create()
fake.uuid()
```

### DIFF
--- a/faker/providers/misc/__init__.py
+++ b/faker/providers/misc/__init__.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 import hashlib
 import random
 import string
+import uuid
 
 from faker.providers.date_time import Provider as DatetimeProvider
 
@@ -68,6 +69,13 @@ class Provider(BaseProvider):
     @classmethod
     def language_code(cls):
         return cls.random_element(cls.language_codes)
+
+    @classmethod
+    def uuid4(cls):
+        """
+        Generates a random UUID4 string.
+        """
+        return str(uuid.uuid4())
 
     @classmethod
     def password(cls, length=10, special_chars=True, digits=True, upper_case=True, lower_case=True):


faker/providers/misc/__init__.py
```diff
@@ -70,6 +71,13 @@ def language_code(cls):
         return cls.random_element(cls.language_codes)
 
     @classmethod
+    def uuid(cls):
```
fcurella: can you rename the moethod to be `uuid4`, so it's clear what kind of uuid is returning?

jaredculp: Sure! Should I add methods for the other uuid types?

fcurella: uuid3 and uuid5 both require some argument, and uuid1 is very insecure, so
I wouldn't bother. But it's really up to you and what you think it's useful
:)

fcurella: > Or perhaps just add a parameter to this method, and have it default to uuid4

I'd rather have it in the name method, makes it more explicit.

jaredculp: I've only ever used `uuid4` so I think that is sufficient. Do I need to add any tests/docs?

faker/providers/misc/__init__.py
```diff
@@ -70,6 +71,13 @@ def language_code(cls):
         return cls.random_element(cls.language_codes)
 
     @classmethod
+    def uuid4j(cls):
```
fcurella: is the `j` a typo?

faker/providers/misc/__init__.py
fcurella: On Tue, Jun 2, 2015 at 9:56 AM, Jared Culp <notifications@github.com> wrote:

> Do I need to add any tests/docs?


that'd be great, thanks!

jaredculp:
@fcurella I'm not sure what to add for tests/docs so I think this is ready to merge if you're okay with it

fcurella:
Thank you very much! :sparkles:

fcurella:
that's okay, i think the test suite automatically builds a smoke test,
which is pretty much what you would've written anyway.

PR #254, opened 2015-06-02
Fixed switched JCB 15 and JCB 16 prefixes

rfsbraz:


fcurella:
Thanks!

I have no idea why such change would drop coverage so much. But just in case, do you have some time to add some tests?

rfsbraz:
@fcurella The coverage also got me by surprise. Even had me re-checking the changes to make sure I didn't touch anywhere else. I'll check coveralls details and see if there is anything I can do, but I'm not even sure what to test.

PR #256, opened 2015-06-06
Fix CVV Code Generation

kevinastone:
Mastercard and American Express were flipped.  AmEx has 4 digit CVV codes and Mastercard has 3 like most of the other cards.

http://en.wikipedia.org/wiki/Card_security_code

### DIFF
--- a/faker/providers/credit_card/__init__.py
+++ b/faker/providers/credit_card/__init__.py
@@ -65,10 +65,10 @@ class Provider(BaseProvider):
     voyager_prefix_list = [['8', '6', '9', '9']]
 
     credit_card_types = {
-        'mastercard':   CreditCard('Mastercard',        mastercard_prefix_list, 16, 'CVV', 4),
+        'mastercard':   CreditCard('Mastercard',        mastercard_prefix_list, 16, security_code='CVV'),
         'visa16':       CreditCard('VISA 16 digit',     visa_prefix_list),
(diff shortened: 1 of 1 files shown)


faker/providers/credit_card/__init__.py
```diff
@@ -118,7 +118,7 @@ def credit_card_full(self, card_type=None, validate=False, max_check=10):
 
     @classmethod
     def credit_card_security_code(cls, card_type=None):
-        return cls.random_number(cls._credit_card_type(card_type).security_code_length)
+        return ''.join(str(cls.random_digit()) for _ in xrange(cls._credit_card_type(card_type).security_code_length))
```
fcurella: can you use `range` instead of `xrange` for compatibility with Python 3.x?

kevinastone:
@fcurella good spot.  fixed.

fcurella:
@kevinastone I don't think we need to worry about providing a compatibility layer. We're `range`-ing for `cls._credit_card_type(card_type).security_code_length`, which is either `3` or `4`. Compared to `range, `xrange` won't improve performance that much on Python 2.

I'd rather keep it simple and just use `range` across all Python versions.

kevinastone:
@fcurella fair enough, pushed/reverted to `range`.

fcurella:
Thanks @kevinastone , really appreciated it! :)

PR #258, opened 2015-06-17
Generate Finnish VAT id and business id (y-tunnus)

### LINKED ISSUE #239: Faker should provide company VAT numbers
It would be nice if faker could provide company VAT numbers. More information about VAT codes is in Wikipedia: https://en.wikipedia.org/wiki/VAT_identification_number

kivipe:
Fixes #239


fcurella:
Thank you!

joke2k:
thank you @kivipe 
maybe the name should be `vat()` without the company prefix?

fcurella:
all other methods on the provider are already `company_*`, so I thought `company_vat` would be at least consistent with the current convention

Issue #14, opened 2014-01-22
Enhance random generator repeatibility


deinspanjer:
The docs mention being able to call the seed() method so you can use a generated dataset as part of a unit test.

Due to the way Faker uses the random module, this usecase is a bit fragile.  Any modification to the data requested, or any outside uses of the random module during generation will diverge the dataset.

Here is a quick script demonstrating the problem along with a couple of potential solutions:
```python
import random
from faker import Faker
fake = Faker()

# initial run
fake.seed(1234)
print fake.name()
print fake.name()
print fake.name()

# repeated run with same data
fake.seed(1234)
print fake.name()
print fake.name()
print fake.name()

# adding new fake calls prevent us from getting the same names we had originally
fake.seed(1234)
print fake.name(), fake.email()
print fake.name(), fake.email()
print fake.name(), fake.email()

# One way is to implement a preserve/restore mechanism so that the user can get back to the previous trail of data
fake.seed(1234)
print fake.name()
r = random.getstate()
print fake.email()
random.setstate(r)
print fake.name()
r = random.getstate()
print fake.email()
random.setstate(r)
print fake.name()
r = random.getstate()
print fake.email()
random.setstate(r)

# A similar problem arises if the program using faker happens to use a non-instance random call during generation.
# The best way to prevent this issue is to have faker use an instance of random rather than the module version.

# If faker used an instance version of random, you could also resolve the original problem by using different faker instances
fake.seed(1234)
fake2 = Faker()
fake2.seed(1234)
print fake.name(), fake2.email()
print fake.name(), fake2.email()
print fake.name(), fake2.email()
```

fcurella:
Fixed in #259

PR #259, opened 2015-06-17
seed a Random instance instead of the module

amygdalama:
Seeding the random module will seed that module globally --
this affects calls to random outside the faker package.
Seeding an instance of the Random class will not affect
other calls to the random module.
This addresses one concern in issue #14.

Let me know if you want me to change anything!


faker/tests/requirements.txt
```diff
@@ -0,0 +1 @@
+mock==1.0.1
```
amygdalama: I used `mock` for the test I added, but I'm not sure how y'all would like to specify this requirement. It only needs to be installed for python < 3.3.

Adding a requirement here means in order to run tests, you'd need to run
```
$ pip install -r tests/requirements.txt
```
This is how django handles this problem: https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/unit-tests/#running-unit-tests-dependencies

Is there another way you'd like to handle this? If we stick with this route, I should probably add the requirement to the documentation somewhere, but I wanted to run this by y'all first to see if this is how you'd like to handle it.

fcurella: I think `tests/requirements.txt` is the way to go.

The various `travis.yml` and co. will need to be updated to run `pip install -r tests/requirements.txt`

amygdalama: Great. I updated the `*.yml`s (I needed to add one for CircleCI) and updated the documentation in the README for running tests.

Any other changes you'd like me to make while waiting for tests to finish?

fcurella: Nope, :sparkles:looks perfect :sparkles:!

Thank you so much!

amygdalama: You're welcome! Could you cut a release with this change? We're waiting on it to merge some changes for our test suite.

fcurella: Releases will have to be coordinated with the original author. I can bump up version numbers but I can't upload to PyPI. Would that work for you?

I could ping the author about uploaidng a release, but he's currently busy working on his Master's Thesis and I'd like let him focus on that :)

amygdalama: Bumping the version works too! Thanks :smile:

fcurella: Done :)

amygdalama:
FYI I released a nose plugin that will seed faker based on the test name before each test: https://github.com/venmo/nose-seed-faker :smile:

fcurella:
Thanks @amygdalama ! That's a great idea!

@joke2k can we upload a release to PyPI?

PR #262, opened 2015-07-09
Fix U.S. SSN generation

### LINKED ISSUE #260: en_US SSN provider generates invalid SSNs
There's a few limitations on United States' SSNs that prevent it from being completely random.

* No group can be all 0s
* The SSN cannot start with 666
* The SSN cannot start with a number >= 900

See http://www.ssa.gov/employer/randomization.html

Could you modify the SSN generator to avoid these issues?

jschaf:
Certain numbers are invalid for U.S. SSNs. The area (first 3 digits)
cannot be 666 or 900-999. The group number (middle digits) cannot be 00.
The id (last 4 digits) cannot be 0000.

See http://www.socialsecurity.gov/employer/randomizationfaqs.html#a0=5
for details.

Fixes #260

### DIFF
--- a/faker/providers/ssn/en_US/__init__.py
+++ b/faker/providers/ssn/en_US/__init__.py
@@ -4,4 +4,17 @@ from .. import Provider as BaseProvider
 
 
 class Provider(BaseProvider):
-    pass
+
+    @classmethod
+    def ssn(cls):
+        # Certain numbers are invalid for U.S. SSNs. The area (first 3 digits)
+        # cannot be 666 or 900-999. The group number (middle digits) cannot be
+        # 00. The serial (last 4 digits) cannot be 0000
+        area = BaseProvider.random_int(min=1, max=899)
+        if area == 666:
+            area += 1
+        group = BaseProvider.random_int(1, 99)
+        serial = BaseProvider.random_int(1, 9999)
+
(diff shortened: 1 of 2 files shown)


faker/providers/ssn/en_US/__init__.py
```diff
@@ -4,4 +4,17 @@
...
+        # Certain numbers are invalid for U.S. SSNs. The area (first 3 digits)
+        # cannot be 666 or 900-999. The group number (middle digits) cannot be
+        # 00. The id (last 4 digits) cannot be 0000
+        area = BaseProvider.random_int(min=1, max=899)
+        if area == 666:
+            area += 1
+        group = BaseProvider.random_int(1, 99)
+        serial = BaseProvider.random_int(1, 9999)
+
+        ssn = "{:03d}-{:02d}-{:04d}".format(area, group, serial)
```
fcurella: We're still supporting python 2.6, which doesn't support this kind of string formatting.

You can rewrite this as `"{0:03d}-{1:02d}-{2:04d}".format(area, group, serial)` or ` "%03d-%02d-%04d" % (area, group, serial)`

jschaf:
Should be fixed now.

fcurella:
Thanks! Really appreciated!

Issue #267, opened 2015-07-15
Python Provider does not support Python 3


derekjamescurtis:
The python provider uses the python 2.x basestring builtin, which is gone in 3.x.

https://docs.python.org/3/whatsnew/3.0.html

https://github.com/joke2k/faker/blob/master/faker/providers/python/__init__.py#L63

Something like the following will trigger it on python 3

``` python

from faker import Factory
f = Factory.create()
f.pylist() # works!
f.pylist(10, True, int) # does not!

```

fcurella:
Closed in #268

PR #268, opened 2015-07-15
Removed basestring from python provider

derekjamescurtis:
Fixes support for Python 3 in the python provider (retains python 2 support).


derekjamescurtis:
Resolves issue #267

fcurella:
good catch! Thank you!

PR #270, opened 2015-07-20
Close #269. Fixed en_GB postcodes to be more realistic

### LINKED ISSUE #269: some generated UK postcodes are invalid
UK postcodes follow strict rules and there are a limited set of codes for each part of the postcode. Faker does not know about these rules and generates postcodes such as: `XC9E 1FL` and `U93 2ZU` which are invalid. See e.g. https://github.com/hamstah/ukpostcodeparser for more info.

fcurella:
Per #269.

Postcodes for UK are generated according to Royal Mail specs. Uses https://github.com/hamstah/ukpostcodeparser/ for validation in tests.

Thanks @mapleoin for the report.


fcurella:
@mapleoin I haven't included `GIR 0AA` for semplicity.

mapleoin:
Looks great. Thanks a lot, @fcurella !

PR #271, opened 2015-07-27
improved Luhn checksum-validaton execution time

0x000:
* Executon time of _validate_luhn_checksum method reduced by ~50%
 * Reduced int function calls
 * Reduced iterations in for loop
 * Added a small lookup dict


fcurella:
Thank you!

PR #272, opened 2015-07-28
Improved credit_card provider

0x000:
* In `_generate_number` method:
 * Improved calculation using a lookup dict
 * Fixed the calculation of the check digit
* In `credit_card_number` method:
 * Removed a bruteforce-like `for` loop
 * Removed useless arguments: `validate` and `max_check`
 * Removed validation checks. `_generate_number` returns always a valid credit card number


fcurella:
:sparkles: This is just amazing! :sparkles:

Thank you!

PR #274, opened 2015-07-29
Minor changes

0x000:
* Fixed some docstrings
* Renamed some attributes
* In `_generate_number` method and `credit_card_security_code` method:
 * Using `BaseProvider.numerify()` for generating random digits


fcurella:
those are _awesome_ changes! Thank you!

PR #276, opened 2015-08-01
added support for brazilian SSN (called CPF)

ericchaves:
Added support for valid CPF (brazilian SSN).


ericchaves:
closing again due to ci failures

Issue #253, opened 2015-06-02
faker.geo_coordinate() should accept Decimals as arguments


mrabbitt:
`faker.geo_coordinate()` returns a `Decimal`, but it doesn't work if you attempt to use `Decimal`'s as arguments:

```
>>> from faker import Factory
>>> faker = Factory.create()
>>> from decimal import Decimal
>>> faker.geo_coordinate(50.5, 0.1)
Decimal('50.5866123539')
>>> faker.geo_coordinate(Decimal(50.5), Decimal(0.1))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mrabbitt/.virtualenvs/data_migration/lib/python2.7/site-packages/faker/providers/address/__init__.py", line 91, in geo_coordinate
    geo = random.uniform(center - radius, center + radius)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/random.py", line 360, in uniform
    return a + (b-a) * self.random()
TypeError: unsupported operand type(s) for *: 'Decimal' and 'float'
```

wybiral:
Does #280 close this issue?

fcurella:
Yes, this should be fixed in d08fd3d12d9bd819555ee36d21e261208e6203fe. Thanks @wybiral :)

PR #280, opened 2015-08-24
Fixed geo_coordinate Decimal issue #253

wybiral:
Fixes issue #253 and makes the precision the same whether or not a center/radius is supplied.

### DIFF
--- a/faker/providers/address/__init__.py
+++ b/faker/providers/address/__init__.py
@@ -93,8 +93,10 @@ class Provider(BaseProvider):
         if not center:
             return Decimal(str(random.randint(-180000000, 180000000) / 1000000.0)).quantize(Decimal('.000001'))
         else:
+            center = float(center)
+            radius = float(radius)
             geo = random.uniform(center - radius, center + radius)
-            return Decimal(str(geo))
+            return Decimal(str(geo)).quantize(Decimal('.000001'))
 
     @classmethod
     def latitude(cls):


faker/providers/address/__init__.py
```diff
@@ -93,8 +93,10 @@ def geo_coordinate(cls, center=None, radius=0.001):
         if not center:
             return Decimal(str(random.randint(-180000000, 180000000) / 1000000.0)).quantize(Decimal('.000001'))
         else:
+            center = float(center)
+            radius = float(radius)
             geo = random.uniform(center - radius, center + radius)
```
fcurella: I'd rather cast `random.uniform` arguments than `center` and `radius`, to minimize precision loss:
```python
geo = random.uniform(float(center - radius), float(center + radius))
```

wybiral: Since radius defaults to floating point 0.001 in the function arguments it will break if all they supply is the center as a Decimal.

fcurella: Oh I see, good point.

This LGTM, Thank you! :)

fcurella:
Thanks! :sparkles:

faker/providers/address/__init__.py
wybiral: No problem.

It might also be a good idea to replace the check from ```if not center:``` to ```if center is None:``` in the event that someone supplies a center of 0 (otherwise it'll get randomized).

fcurella: Good call. Would you mind submit another PR?

wybiral: Sure thing, it's PR #281 

This is probably a bad place to ask (I'm new to Github), but does this method need some tests added since this seems to decrease the coverage/coveralls?

fcurella: We have some _basic_ testing, where every provider gets called at least once with the default arguments, but that obviously doesn't cover most cases.

New per-method tests would be a nice addition and are really appreciated.

PR #281, opened 2015-08-24
Fixed condition for None in geo_coordinate

wybiral:
The original check to see if nothing was supplied as a center was for falseness, which would cause random results if they intended to supply a center of 0.


fcurella:
Thanks!

PR #285, opened 2015-09-18
Adds en_AU provider, adapted from PHP Faker

xfxf:
Hello,

I've submitted a few accepted patches for the en_AU provider for the PHP Faker implementation recently; as a 'native' Python developer it only makes sense to contribute a provider back to this library.

Let me know if OK.

Reference implementation this is based from at https://github.com/fzaninotto/Faker/tree/master/src/Faker/Provider/en_AU

### DIFF
--- a/faker/providers/address/en_AU/__init__.py
+++ b/faker/providers/address/en_AU/__init__.py
@@ -1,0 +1,109 @@
+from __future__ import unicode_literals
+
+from ..en import Provider as AddressProvider
+from faker.generator import random
+
+class Provider(AddressProvider):
+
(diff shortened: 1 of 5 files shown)


faker/providers/address/en_AU/__init__.py
```diff
@@ -0,0 +1,109 @@
+from __future__ import unicode_literals
+
+from ..en import Provider as AddressProvider
```
fcurella: can you extend from `faker.providers.address.en.Provider`?

```python
from ..en import Provider as AddressProvider
```

fcurella: nevermind, I've missread your line :)

fcurella:
Looks Good! Thank you for contributing back! :heart:

PR #286, opened 2015-09-22
Update de_DE first names.

fcurella:
Do these name look about right, @WarrenFaith ?


WarrenFaith:
That looks better, also nice to see where they come from :)

PR #287, opened 2015-10-09
Feature/croatian localization

mislavcimpersak:
adding Croatian localization strings for persons, internet, phone, company and address


fcurella:
Looks good to me!

thank you @mislavcimpersak :) :sparkles:

PR #293, opened 2015-10-31
Created Bosnian phone number provider

elahmo:
This adds the phone numbers for Bosnia and Herzegovina:

* All three major carriers
* Area codes for cantons in Federation of BiH
* Area codes for regions in Republic of Srpska


fcurella:
Good work :) :sparkles:

PR #294, opened 2015-10-31
Added internet provider for bs_BA

elahmo:
* added TLDs listed on nic.ba
* added two of the provider specific email address formats, along the free ones


fcurella:
Thank you! :) :sparkles:

PR #295, opened 2015-10-31
Added swiss names (de_CH)

DanEEStar:
Do I have to do other things/steps?

### DIFF
--- a/faker/providers/person/de_CH/__init__.py
+++ b/faker/providers/person/de_CH/__init__.py
@@ -1,0 +1,110 @@
+# coding=utf-8
+from __future__ import unicode_literals
+from .. import Provider as PersonProvider
+
+
+class Provider(PersonProvider):
+    formats = (
+        '{{first_name}} {{last_name}}',
+        '{{first_name}} {{last_name}}',
+        '{{first_name}} {{last_name}}',
+        '{{first_name}} {{last_name}}',
+        '{{first_name}} {{last_name}}',
+        '{{first_name}} {{last_name}}',
+        '{{first_name}} {{last_name}}',
+        '{{first_name}} {{last_name}}-{{last_name}}',
+        '{{first_name_female}} {{last_name}} {{last_name}}',
+        '{{prefix}} {{first_name}} {{last_name}}',
+    )
+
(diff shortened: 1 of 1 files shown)


fcurella:
Thank you @DanEEStar ,

I'm not sure how we should handle the Swiss case. I know Switzerland has many official languages, and the names contained in this PR sounds Italian, French and German.

I'm not familiar with common practices in Switzerland re: locales.

Is `de_CH` used as de-facto 'default'? Alternatively, we could also add `it_CH` and `fr_CH` (which could simply subclass from `de_CH`).

What do you think would be more appropriate?

Hopiu:
I'd say it generally doesn't make sense to have de_CH as the _CH default. As I saw you already added the data for fr_CH to the repo.

If you are unhappy with the names @DanEEStar provided I'd be willing to revise them on the base of the data provided by the Swiss government.
(Available at https://www.bfs.admin.ch/bfs/de/home/statistiken/bevoelkerung/geburten-todesfaelle/vornamen-schweiz.html)

DanEEStar:
@Hopiu I was also thinking about updating this pull-request recently.
What are your thoughts about the French and Italian "sounding" names in the dataset. In my opinion it is not that uncommon for a German speaking swiss person to have a French or Italian name.

fcurella:
I'd much rather have names from a governmental source. Not sure if there are separate lists for the different Swiss locales. If not, we can just use the same list for all of them.

Hopiu:
@DanEEStar Have a look at the link I provided. You'll find official excel sheets containing the most common names in Switzerland separated by sex and language region.
You will still find some French and Italian "sounding" names there.

DanEEStar:
@Hopiu The list contains only first names though. Do you know any statistical source for last names.

I have a list with about 100000 swiss persons with names and language from which I extracted and aggregated the data for the first pull request.
But the list is not officially public...

nfrei:
@fcurella Can this be merged? I think it's a decent list of Swiss names.
By the way, there is already localization for Swiss french (fr_CH) on at least jobs and addresses in faker. So it's just consistent to include de_CH as well.

PR #298, opened 2015-11-06
Add an 'office' file extensions category and extensions

j0hnsmith:
containing extensions of files commonly found within an office
environment


fcurella:
Good idea, Thank you very much! :sparkles:

PR #299, opened 2015-11-09
Add tzinfo support to datetime methods

j0hnsmith:


fcurella:
Nice work! Thank you! :sparkles:

PR #300, opened 2015-11-12
Update ja_JP for username and domain

massa142:
In order to get username and domain in ja_JP
・Add [Romanized Japanese](https://en.wikipedia.org/wiki/Romanization_of_Japanese) person name
・Add internet provider for ja_JP


fcurella:
thank you @massa142 !

Could you add some test for `username` and `domain_word` as well? I'm trying to keep the coverage up

massa142:
OK!!
I added additional test. plz check it.

fcurella:
Thank you very much! Nice Work :sparkles:

PR #301, opened 2015-11-23
Handle unexpected length in ean()

michaelcho:
I didn't realise ean lengths could only be 8 or 13 digits and tried to use `faker.ean(length=20)` and got `UnboundLocalError: local variable 'weights' referenced before assignment`.

Adding this so that it defaults to 13 digits....

### DIFF
--- a/faker/providers/barcode/__init__.py
+++ b/faker/providers/barcode/__init__.py
@@ -10,7 +10,7 @@ class Provider(BaseProvider):
         code = [self.random_digit() for i in range(length - 1)]
 
         if length not in (8, 13):
-            length = 13
+            raise AssertionError("length can only be 8 or 13")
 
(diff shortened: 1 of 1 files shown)


fcurella:
Thanks @michaelcho, you raise a valid point.

Instead of silently switching to `13` fro invlaid values, I'd rather raise a descriptive error.

Something like:

```python
assert (length in (8, 13)), "length can only be 8 or 13"
```

or, if you'd rathernot use `assert`:

```python
if length not in (8, 13):
    raise AssertionError("length can only be 8 or 13")
```

michaelcho:
@fcurella I updated the pull request....

fcurella:
Thank @michaelcho , looks good :) :sparkles:

PR #303, opened 2015-11-30
Close #302. Allow unicode in emails and domains

### LINKED ISSUE #302: zh_CN email not returning valid email addresses
When creating a fake Factory with the zh_CN or zh_TW it is not returning valid email addresses.

>>> f = Factory.create(locale='zh_CN')
>>> f.email()
u'.@yahoo.com'
>>> f.email()
u'.@gmail.com'
>>> f.email()
u'76@gmail.com'
>>> f.email()
u'.@.com'
>>> f.email()
u'.@.net'
>>> f.email()
u'.@.com'
>>> f.email()
u'.@yahoo.com'
>>> f = Factory.create(locale='zh_TW')
>>> f.email()
u'u@.com'
>>> f.email()
u'm@.com'
>>> f.email()
u'x@.org'
>>> f.email()
u'89@.net'
>>> f.email()
u'25@.com'
>>> f.email()
u'e@.info'
>>> f.email()
u'.@.com'

fcurella:


zdelagrange:
repost from https://github.com/joke2k/faker/issues/302

@fcurella That does not address the possibility for unicode domain names, for example, http://☃.net/ is a valid domain. and also does not address the problem that the user_name function will sometimes return invalid user_names such as '.', which results in invalid email addresses like '.@.com'.

I think this is a good effort to get the functions 'working' but it does not address the possibility that unicode characters can exist in these places. Working around them seems to be the wrong approach.

fcurella:
@zdelagrange does the new solution (`allow_unicode`) work for you?

zdelagrange:
I pulled your branch and tested it out, seems like there is a bug in email. 

This does address the issue of the email function not returning a valid email.

But I don't think this addresses the real localization issue, which is that unicode characters can exist in email addresses and domains. It seems that you're just changing the characters to ascii characters, which will make the module work, but from a testing perspective, this is not different than using the en_* localization, because it is not exercising the possibility of unicode characters existing in these places according to w3 & IETF standards

For example, this is a valid domain: http://はじめよう.みんな/


```
>>> from faker.factory import Factory
>>> f = Factory.create('zh-CN')
>>> f.user_name
<bound method Provider.user_name of <faker.providers.internet.zh_CN.Provider object at 0x7fcff7c69510>>
>>> f.user_name()
u'hui-guo'
>>> f.user_name()
u'jlu'
>>> f.user_name()
u'kxi'
>>> f.user_name()
u'ezuo'
>>> f.user_name()
u'lli'
>>> f.user_name()
u'rong-wu'
>>> f.user_name()
u'xiu-zhen-83'
>>> f.user_name()
u'lin-92'
>>> f.user_name()
u'xu-tao'
>>> f.email()
  File "faker/providers/internet/__init__.py", line 64, in email
    return "".join(self.generator.parse(pattern, allow_unicode=allow_unicode).split(" "))
  File "faker/generator.py", line 84, in parse
    return _re_token.sub(format_token, text)
  File "faker/generator.py", line 88, in __format_token
    formatter[1] = self.format(formatter[1], *args, **kwargs)
  File "faker/generator.py", line 63, in format
TypeError: first_name() got an unexpected keyword argument 'allow_unicode'
>>> f.domain()
AttributeError: 'Generator' object has no attribute 'domain'
>>> f.domain_name()
u'cai-hong-chuan-mei-you-xian-gong-si.biz'
>>> f.domain_name()
u'yi-dong-li-xin-xi-you-xian-gong-si.com'
>>> f.domain_name()
u'hui-lai-ji-suan-ji-wang-luo-you-xian-gong-si.org'
```
Something like this I feel would resolve the issue:

```
>>>f.email()
欢@浙大万朋传媒有限公司.cn
```

fcurella:
@zdelagrange The idea is that you can opt-in by calling `f.email(allow_unicode=True)`

fcurella:
@zdelagrange I've changed the PR to use unicode by default, give it a try

zdelagrange:
Ah I see, I do admit I glossed over the change. I think if anything an opt-out would is better than an opt-in, but not allowing an opt-in/out is probably the in the best interest of localizing the internet as a whole :)

This change looks to solve the problem.

Issue #304, opened 2015-12-01
date_time_this_month logic error


mark-love:
The logic blows up when the current month (now.month) is December (12) as I have found out. 

faker/providers/date_time/__init__.py
`next_month_start = datetime(now.year, now.month + 1 % 12, 1, tzinfo=tzinfo)`
should be 
`next_month_start = datetime(now.year, (now.month + 1) % 12, 1, tzinfo=tzinfo)`

PaulLiang1:
Had the same issue just right now...

fcurella:
Hi @mark-love , thanks for the report.

Do you mind submit a Pull Request? Really appreciate it :)

mark-love:
No problem, I can work on one right now.

fcurella:
@mark-love `now().month` is zero-based, so it needs to be compensated by doing something like:

```python
(now.month + 2) % 12
```

mark-love:
@fcurella is it? From what I can tell in the documentation and experimenting is that it is between 1 and 12. https://docs.python.org/2/library/datetime.html#datetime.datetime.month

fcurella:
@mark-love , no you're right.

The issue is that on Nov 30th, 'next month' means December, and `12 % 12` returns `0`, when we actually need it to be `12`

fcurella:
I'll have to wait for tomorrow morning to dig deeper into this

mark-love:
@fcurella ah yes that would be it `now.month % 12 + 1` should fix it.
Do we also want/need the year to increment or is that unnecessary?

rshk:
I guess the year needs to be incremented as well, "next month" in Dec '15 is Jan '16, not Jan '15..

so better be explicit, something like:

```
if now.month == 12:
    next_year = now.year + 1
    next_month = 1
else:
    next_year = now.year
    next_month = now.month + 1
```

rshk:
Btw, the current code is failing because of operator precedence (`%` has higher precedence than `+` so ``now.month + 1 % 12`` is actually ``now.month + (1 % 12)``), but anyways, ``(now.month + 1) % 12`` is broken as well, as month is one-based.

kot-behemoth:
Is there any reason to not use `relativedelta` from the `dateutil` package (https://pypi.python.org/pypi/python-dateutil)? It provides a very robust way to get "next month" and is a well-known module:

```python
import datetime
import dateutil

next_month = datetime.date.today() + dateutil.relativedelta.relativedelta(months=1)

[...]

PR #306, opened 2015-12-01
Fix an order of operations error #304

mark-love:
In the date_time_this_month() function of the date_time module there is an error that can provide 13 to the month value of datetime() when now.month is 12 `now.month + 1 % 12`.

This patch fixes this error by enforcing the correct order of operation for the function.


fcurella:
fixed in #307

PR #311, opened 2015-12-07
Correct Poland phone numbers.

fizista:
Based on the official list of telephone numbering in Poland.


fcurella:
Thanks! :sparkles:

PR #312, opened 2015-12-17
Created course provider for things related to fake courses

danhuss:
It probably isn't a big request, but created a provider to generate fake subjects, grades, and course titles. I think I have everything looking ok, let me know if there's anything I need to change.

Thanks,
Daniel


fcurella:
Hi @danhuss ,

I think this case is not generic enough to warrant being included in Faker. In other words, I don't think many people would find that useful.

You could still package your provider as a stand-alone Python package, and use it with faker.

Assuming your provider has an import path of `course_provider.Provider`:

```
from course_provider import Provider as CourseProvider
from faker import Faker

fake = Faker()
fake.add_provider(CourseProvider)

fake.grade()
```

danhuss:
understandable. Is there a wiki or something that lists other providers like this then, and if so, can I pull the provider into a separate repo and get on that list? Thanks.

fcurella:
> Is there a wiki or something that lists other providers like this

Not that I'm aware of, but we could add one :)

k1-hedayati:
I just found this repo and after going through some docs and codes saw this PR, I'm just an outsider and not much aware of project state and your plans for it but I think what `faker` needs most is a long list of providers which is available out of box.
Software projects vary too much, there is never that "many" people who find a specific provider useful.
I'm just writing a shopping center project along side a flight information system they both need a different fake data and current providers in `faker` can't help me, I can create a file for each project and throw it in my github account like @danhuss did but the next guy who wants to create the same fake data won't find my repo and will either won't use `faker` at all or have to create those files again even if he finds, in that time won't care about my repo and probably won't respond to his requests.
But if provider lies here he can simply use it and if he has any request he can create a issue here and one of 3.4k people who liked this project will respond and help him. 

If provider list gets out of control we can always push them out of repo and use a github org with a plugin system, but right now the list isn't that long and probably won't rise much maintenance issue.

Thanks for your efforts :+1:

danhuss:
@k1-hedayati Did you run across this in the "Extended Docs" (which are linked to from the readme) https://fake-factory.readthedocs.org/en/latest/providers.html? I could see how putting that in the readme in a comma separated format would be nice.

fcurella:
We can't possibly cover 100% of the cases. We will end up with instances of providers that are very similar to each other except for that one difference that matters to that one person.

Currently, you can create a custom provider (http://fake-factory.readthedocs.org/en/stable/index.html#how-to-create-a-provider). You could include it in your project or, if you want to reuse it across projects, make it its own python package and link to it.

That said, I'm interested in finding a longer term solution to this issue.

I'm willing to maintain a list of community-supported providers in the docs, but I'm -0 on putting them under a `faker` umbrella org, because that puts the burdain of those packages (taking care of code, community, governance, etc) to the maintainers of `faker` (ie: me and @joke2k).

Would be really cool if faker had its own 'packages store' where you can discover and download providers from, but that involves server expenses we don't have the resources for.

/ping @joke2k for feedback

k1-hedayati:
@danhuss yes readme is more accessible and manageable than wiki

> We can't possibly cover 100% of the cases. We will end up with instances of providers that are very similar to each other except for that one difference that matters to that one person.

Of course we can't cover all cases but we can try to cover more, a provider for product names is very different than airlines or course names.

You're seeing the big picture, I believe it is too big. Right now there is under 20 providers available and reaching to 100 takes quite some time and majority of code in providers is just some fake strings and I don't think they need much maintenance.
Anyway both of community-supported providers or `faker` org will work, my only concern is let's put a good use on the time spent by @danhuss and others who created a provider for this project.

fcurella:
moved to #319

PR #325, opened 2016-01-22
Resolves #324. Add characters classmethod to lorem provider

### LINKED ISSUE #324: Add ability to generate lorem characters without punctuation
Sometimes I want to generate a string of characters of a specific length without any punctuation or capitalization but the lorem provider currently does not allow for this.

gwongz:
Sometimes I want to generate a string of characters of a specific length
without any punctuation or capitalization. Adding a characters() method
allows for this. It defaults to a 255-character, mixed case string but
the method can also be called with params that specify length and
whether to exclude uppercase characters.

### DIFF
--- a/faker/providers/lorem/__init__.py
+++ b/faker/providers/lorem/__init__.py
@@ -6,20 +6,6 @@ from .. import BaseProvider
 
 class Provider(BaseProvider):
     @classmethod
-    def characters(cls, max_nb=255, variable_case=True):
-        """
-        Generate a random character string
-        :param max_nb number of characters the string should consist of
-        :param variable_case set to False to exclude uppercase characters
-        :example 'pmGYPlciRq'
-        """
-        characters = "".join(cls.random_letter() for i in range(max_nb))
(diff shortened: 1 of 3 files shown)


fcurella:
I don't think this should go into the `lorem` provider, as it's not providing filler text. I'd rather have it as a replacement for `pystr` inside the `Python` provider.

faker/providers/lorem/__init__.py
```diff
@@ -6,9 +6,24 @@
 
 class Provider(BaseProvider):
     @classmethod
+    def characters(cls, max_nb=255, variable_case=True):
```
fcurella: I don't think `variable_case` is necessary`. You could achieve the same by simply calling `Provider.characters().lower()` (or `.upper()`, or `.title()`)

gwongz:
Thanks for the feedback @fcurella. I moved this from `lorem` to the `Python` provider.

fcurella:
Thank you @gwongz , Nice work! :sparkles:

PR #327, opened 2016-02-18
IPv4 and IPv6 network generators.

kdeldycke:
Adds optional IPv4 and IPv6 network address generation with valid CIDR notation.

Also add corresponding unit-tests.

Usage example:
```python
>>> from faker import Factory as FakerFactory
>>> faker = FakerFactory.create()
>>> faker.ipv4()
u'94.84.244.87'
>>> faker.ipv4(network=True)
u'105.0.0.0/10'
>>> faker.ipv4(network=True)
u'131.46.122.140/30'
>>> faker.ipv6()
u'0b9e:425c:8028:ffd1:88a1:9d54:a4a0:cafb'
>>> faker.ipv6(network=True)
u'1e1b:0791:da18:0000:0000:0000:0000:0000/45'
>>> faker.ipv6(network=True)
u'79c1:84fc:53b8:c9a8:23ca:7407:a775:0180/121'
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,8 +24,8 @@ Changelog
 * Add Maestro credit card. Thanks @anthonylauzon.
 * Add ``hr_HR`` localization. Thanks @mislavcimpersak.
 * Update ``de_DE`` first names. Thanks @WarrenFaith and @mschoebel.
-* Allow generation of IPv4 and IPv6 network address with valid CIDR.
-* Unittest IPv4 and IPv6 address and network generation.
+* Allow generation of IPv4 and IPv6 network address with valid CIDR. Thanks @kdeldycke.
+* Unittest IPv4 and IPv6 address and network generation. Thanks @kdeldycke.
 
 `0.5.3 - 21-September-2015 <http://github.com/joke2k/faker/compare/v0.5.2...v0.5.3>`__
 --------------------------------------------------------------------------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,6 +24,8 @@ Changelog
 * Add Maestro credit card. Thanks @anthonylauzon.
 * Add ``hr_HR`` localization. Thanks @mislavcimpersak.
 * Update ``de_DE`` first names. Thanks @WarrenFaith and @mschoebel.
+* Allow generation of IPv4 and IPv6 network address with valid CIDR.
+* Unittest IPv4 and IPv6 address and network generation.
 
 `0.5.3 - 21-September-2015 <http://github.com/joke2k/faker/compare/v0.5.2...v0.5.3>`__
 --------------------------------------------------------------------------------------
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -9,11 +9,6 @@ from faker.generator import random
 from faker.providers.lorem.la import Provider as Lorem
 from faker.utils.decorators import slugify, slugify_unicode
 
-try:
-    unicode
-except NameError:  # pragma: no cover
-    unicode = str  # pylint: disable=W0622,C0103
-
 
 localized = True
 
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -3,6 +3,8 @@
 from __future__ import unicode_literals
 from .. import BaseProvider
 
+from ipaddress import ip_address, ip_network, IPV4LENGTH, IPV6LENGTH
+
 from faker.generator import random
 from faker.providers.lorem.la import Provider as Lorem
 from faker.utils.decorators import slugify, slugify_unicode
@@ -50,9 +52,6 @@ class Provider(BaseProvider):
      )
 
     replacements = tuple()
-
-    IPV4LENGTH = 32
-    IPV6LENGTH = 128
 
     def _to_ascii(self, string):
         for search, replace in self.replacements:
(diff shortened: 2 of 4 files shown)


CHANGELOG.rst
```diff
@@ -24,6 +24,8 @@ Changelog
 * Add Maestro credit card. Thanks @anthonylauzon.
 * Add ``hr_HR`` localization. Thanks @mislavcimpersak.
 * Update ``de_DE`` first names. Thanks @WarrenFaith and @mschoebel.
+* Allow generation of IPv4 and IPv6 network address with valid CIDR.
+* Unittest IPv4 and IPv6 address and network generation.
```
fcurella: thank you, very much appreciated! Feel free to add your name next to those entries :)

faker/providers/internet/__init__.py
```diff
@@ -99,15 +102,39 @@ def url(self):
         pattern = self.random_element(self.url_formats)
         return self.generator.parse(pattern)
 
-    def ipv4(self):
+    def ipv4(self, network=False):
+        """ Produce a random IPv4 address or network with a valid CIDR.
+
+        IP code generation inspired by ipaddress module: https://github.com/phihag/ipaddress
```
fcurella: If `phihag/ipaddress` and Python3's `ipaddress` have the same api and functionality, I think i'd rather add `phihag/ipaddress` as a requirement in `setup.py` for Python < 3.3.

kdeldycke: That's even better! Didn't went this way as I wasn't sure you'll accept more dependencies. I'll rework the generator by reusing `ipaddress` then.

CHANGELOG.rst
kdeldycke: Done in commit: https://github.com/kdeldycke/faker/commit/a9c021717cf2a45b09a9ffc00b8897cd1df31fbd

faker/providers/internet/__init__.py
kdeldycke: Done in commit: https://github.com/kdeldycke/faker/commit/d16ea3c7a76d530ec64c9ab0d3533346618be6df

faker/providers/internet/__init__.py
```diff
@@ -99,15 +101,23 @@ def url(self):
...
-    def ipv6(self):
-        res = [hex(random.randint(0, 65535))[2:].zfill(4) for i in range(0, 8)]
-        return ":".join(res)
+    def ipv4(self, network=False):
+        """ Produce a random IPv4 address or network with a valid CIDR. """
+        address = str(ip_address(random.randint(
+            0, (2 ** IPV4LENGTH) - 1)))
+        if network:
+            address += '/' + str(random.randint(0, IPV4LENGTH))
+            address = str(ip_network(unicode(address), strict=False))
```
fcurella: because of the concatenation on the line above, I would expect `address` to already be coerced to a string at this point, making `unicode()` unnecessary

kdeldycke: See for yourself how `ipaddress` expect explicit unicode string:
```python
Python 2.7.11 (default, Dec  5 2015, 14:44:53) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.1.76)] on darwin
>>> from ipaddress import ip_network
>>> ip_network(unicode('49.50.51.52/22'), strict=False)
IPv4Network(u'49.50.48.0/22')
>>> ip_network('49.50.51.52/22', strict=False)
  File "/Users/kdeldycke/virtualenvs/billing/lib/python2.7/site-packages/ipaddress-1.0.16-py2.7.egg/ipaddress.py", line 199, in ip_network
ipaddress.AddressValueError: '49.50.51.52/22' does not appear to be an IPv4 or IPv6 network. Did you pass in a bytes (str in Python 2) instead of a unicode object?
>>> 
```

kdeldycke: Ah yes, got it. Fixed in commit: https://github.com/joke2k/faker/commit/76720d7c35c436ba27638d9bfc25fb9e883ba0a7

kdeldycke:
@fcurella: now the only issue preventing a merge is making `pypy` install `ipaddress`. I don't have a clue how to specify it via `setup.py`'s `extras_require`parameter. Can you help me?

fcurella:
@kdeldycke pypy3 is Python 3.2, and `ipaddress` was introduced in Python 3.3.

I think all you need is to specify the extra requirement as

```
        ':python_version=<"3.3"': [
            'ipaddress'
        ]
```

or, alternivately:

```
        ':python_version=="3.2"': [
            'ipaddress'
        ]
```

kdeldycke:
Even better! I'll push the fix tomorrow.

kdeldycke:
This PR is ready to be merged.

faker/providers/internet/__init__.py
```diff
@@ -3,10 +3,18 @@
...
+from ipaddress import ip_address, ip_network, IPV4LENGTH, IPV6LENGTH
+
 from faker.generator import random
 from faker.providers.lorem.la import Provider as Lorem
 from faker.utils.decorators import slugify, slugify_unicode
 
+try:
+    unicode
+except NameError:  # pragma: no cover
+    unicode = str  # pylint: disable=W0622,C0103
```
fcurella: I don't see the `unicode` module used anytwhere in this diff.

Additionally, `six` is already in the requirements, so if you need this behaviour you can use `six.text_type` https://pythonhosted.org/six/#six.text_type

kdeldycke: Yes, was removed in https://github.com/kdeldycke/faker/commit/76720d7c35c436ba27638d9bfc25fb9e883ba0a7 . Just fixed that.

fcurella:
Looks good! Thanks! :sparkles:

kdeldycke:
Thanks ! :)

PR #328, opened 2016-02-19
Add a binary blob generator.

kdeldycke:


faker/providers/misc/__init__.py
```diff
@@ -27,6 +27,15 @@ def null_boolean(cls):
         }[random.randint(-1, 1)]
 
     @classmethod
+    def binary(cls, length=(1 * 1024 * 1024)):
+        """ Returns random binary blob.
+
+        Default blob size is 1 Mb.
+        """
+        return bytes(bytearray([
+            random.randrange(0, 256) for _ in range(length)]))
```
fcurella: how about using `os.urandom(length)` length instead? https://docs.python.org/2/library/os.html#os.urandom

kdeldycke: As you wish. Did the change in commit https://github.com/kdeldycke/faker/commit/7fcec3c812d71722d8501822681f27be6c4422f4

faker/tests/__init__.py
```diff
@@ -459,6 +459,16 @@ def test_date_time_this_period_with_tzinfo(self):
...
         )
 
+    def test_binary(self):
+        from faker.providers.misc import Provider
+
+        for _ in range(999):
+            length = random.randint(0, 2 ** 10)
+            binary = Provider.binary(length)
+
+            self.assertTrue(isinstance(binary, bytes))
```
fcurella: I think this will fail on Python2, since os.urandom return `str`. You should compare with `six.binary_type` instead:

`self.assertTrue(isinstance(binary, six.binary_type))`

kdeldycke: Done ! :)

kdeldycke:
There's still an issue with the CLI side of things: should we simply blacklist the binary blob generator for the CLI? As there's a hight chance of garbling the user's terminal if we output data as-is.

fcurella:
> There's still an issue with the CLI side of things

I've fixed that in master: https://github.com/joke2k/faker/commit/6c27a07fe6d9cd18616b76b510a4555ed25f744a

I think all that's left is rebasing against master. I probably just conflicting on  `tests/__init__.py`.

fcurella:
I've fix the conflicts and merged it in. Thanks! :sparkles:

kdeldycke:
I apologize. I did not realized I messed up the rebase. Sorry for the inconvenience.

fcurella:
no worries! thanks again for your contribution! :)

kdeldycke:
Any plan for a release then?

fcurella:
I'm trying to coordinate with @joke2k , which as the original author is the one responsible for uploading on PyPI

kdeldycke:
Great ! :)

PR #329, opened 2016-02-23
Fix randomly produced language codes.

kdeldycke:
First commit shows that the `cn` language code is not recognised by the `Factory` constructor.

Second commit replace `cn` by `zh`, the former being the Chinese TLD, not Chinese's language code.


fcurella:
Nice catch! Thank you! :sparkles:

kdeldycke:
That was fast. Thanks! :smile:

PR #330, opened 2016-02-27
Command improvements

cbaines:
Thanks for writing fake-factory (or faker, still a little confused about the name) and releasing it as free software, I am using it for several projects.

While writing a man page for faker, as part of packaging it for Debian, I noticed that the help given when passing --help was a little lacking. While attempting to improve this, I also noticed that the fake argument was oddly structured, so I changed that also.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -170,7 +170,7 @@ When installed, you can invoke faker from the command-line:
           [-l {bg_BG,cs_CZ,...,zh_CN,zh_TW}]
           [-r REPEAT] [-s SEP]
           [-i {module.containing.custom_provider othermodule.containing.custom_provider}]
-          [fake [fake ...]]
+          [fake] [fake argument [fake argument ...]]
 
 Where:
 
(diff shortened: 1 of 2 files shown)


fcurella:
@cbaines does splitting the `fake` argument change the way the CLI should be called? In other words, does [this section](https://github.com/joke2k/faker#command-line-usage) of the `README` need to be changed?

fcurella:
@cbaines FYI, I'm looking at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=816169 and it mentions the license as 'Expat', when in fact `faker` is released under MIT.

cbaines:
> @cbaines does splitting the fake argument change the way the CLI should be called? In other words, does this section of the README need to be changed?

It does not change the way in which it should or can be called, but its probably worth updating the README to be more consistent with the --help output (which this branch goes some way to doing). I'll add another commit improving this shortly.

cbaines:
> @cbaines FYI, I'm looking at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=816169 and it mentions the license as 'Expat', when in fact faker is released under MIT.

MIT is an ambiguous name that sometime refers to the Expat license (as it does in the case of fake-factory/faker), see https://en.wikipedia.org/wiki/MIT_License#Various_versions for more information.

fcurella:
Thanks for clarifyinf @cbaines, I wasn't aware of this subtlety. Nevermind
my comment re: the license.

cbaines:
> Thanks for clarifyinf @cbaines, I wasn't aware of this subtlety. Nevermind my comment re: the license.

No problem :) I have now updated the README. The help output and it still differ, but I think that is ok, as I think the README is more helpful in some ways (e.g. abbreviating the language code list, and using `module.containing.custom.provider`).

fcurella:
thanks! :sparkles:

PR #337, opened 2016-03-06
Geometry provider

singingwolfboy:
This is https://github.com/rbarrois/factory_boy/pull/174 applied to Faker instead.


joke2k:
Hi, thank you for contrib.
How do you think about generalize your Point provider into a Geometry provider?
We can add a lot of think like Segment, Square, Circle, Angle and other shapes.

fcurella:
I agree, a `Point` provider is too specific. I would rename it to `Geometry`, and have one unified `.point` method which can take an optional `dim` argument, defaulting to `2`

singingwolfboy:
@fcurella: The problem with an argument to specify the number of dimensions is, how do you then specify min, max, and step for each of those dimensions individually? That's why I have arguments like `min_x`, `max_y`, and so on. Do you have a suggestion for that?

fcurella:
@singingwolfboy the signature would be the same as `point_xyz`, but we addition of `dim`.

You'd have `min_*` and `max_*`. We simply ignore `min_z` and `max_z` if `dim == 2`

singingwolfboy:
I've updated the PR to respond to feedback.

singingwolfboy:
Any update on this pull request? The build failures on Travis CI and Appveyor seem unrelated to the code -- the error messages point to the instances running out of memory.

PR #338, opened 2016-03-08
Discern male and female first names in pt_BR

gabrielusvicente:
Fixed male and female first names in pt_BR


fcurella:
Thanks! :sparkles:

PR #342, opened 2016-03-27
More command line improvements

cbaines:
Print the available locales in a separate section to make the help output easier to read. Also, include the examples given in the README in the help output.


fcurella:
Good call! Thanks! :sparkles:

PR #344, opened 2016-03-31
Changed US_EN names to be more realistic.

dethpickle:
Desc: Updated first names to be reflect the most common US adult names for people 15-55 years old based on census birth records and to be used proportionally to their realistic frequency.  Last names updated to be the top 1000 US surnames weighted by frequency.

This is my first pull request and I'm not sure how to leave out the .gitignore.  I work on a mac and had to add .DS_Store.

Hey, first thanks all of you for all the awesome work on this module.  It's been a huge help to me over the years.

One of the things that's bugged me was that US names ended up being exceptional too frequently.  There are quite a lot of names and they're all used perfectly randomly.  I'm modeling sample documents for OCR purposes (checks mostly) and the high concentration of european names was a bit unrealistic.  
I wanted to make it use common names based on how common they are with a weighted dict - then to dig into it and you already HAVE it designed in.  Just no data for it.

So I took US baby names from census data from the 60s, 70s, 80s and 90s and the numbers that came with them to build weighted dicts for EN_US.  Something similar for last name, but just from current data.

https://www.ssa.gov/OACT/babynames/decades/names1960s.html
http://names.mongabay.com/data/1000.html

This fits my needs much better.  It might be of use to others.  **It might be more appropriate to build it in as a separate region (EN_USw) or something to leave the existing data.**  I'm not quite sure how to do that though.  I'd welcome suggestions if you think it'd be useful, but as an addition, not an update.

Looking through some of the open issues, I'd still like to further refine this and other providers to have some agreement in the generated data.  Even what I've just produced is a little wonky in that it's mixing cultural first names and last names randomly (though that does happen).  But a more sophisticated and realistic fake profile could start with age and generate a selection of data most appropriate to that starting data.  I've averaged away the first name popularity that came in by decade, but that's what I mean.  Older people with common names of Michael, David, Lisa and Mary and younger ones with Jacob, Joshua, Emily and Madison.  That would be neat...  This is just a first stab in that direction.


fcurella:
Thanks @dethpickle , impressive work!

re: the `.gitignore`, I think adding the .DS_Store is totally fine. I'm actually surprised it wasn't there already.

I think your changes should go into the `en_US` provider (which is currently 'empty', and we can leave the more generic `en` the way it is.

I'm not worried about changing the distribution of how names are selected for `en_US`. We never documented the specific distribution of this particular provider, so I don't expect people to have built things that rely on it.

dethpickle:
Okay, sounds good.  So, I should close this pull request, get a new fork, put it in en_US and submit a new pull request?

fcurella:
@dethpickle you can update this one. just make the necessary changes, and push a new commit

dethpickle:
Right.  Maybe a newb question, but how do I un-commit the change I made to en/\__init\__.py then and restore it to what it was in my fork?  That's what i was angling at.

Or do I change it back and commit that as well?  Is it acceptable to see an X + y - y in the history on some file?

fcurella:
@dethpickle Don't worry, you can simply change it back and commit it. I'm not too worried about having many commits

dethpickle:
Roger.  I'll try to get to it tonight.

dethpickle:
It should be ready. Let me know if there is anything further I need to do for this pull. Thanks.

fcurella:
Thank you very much @dethpickle ! :sparkles:

PR #347, opened 2016-04-05
Add method to specify a random string of minimum and maximum length.

tamarbuta:
Added a string method in the **misc** provider to generate a random string of min and max length.

### DIFF
--- a/faker/providers/misc/__init__.py
+++ b/faker/providers/misc/__init__.py
@@ -127,16 +127,3 @@ class Provider(BaseProvider):
             chars[index] = required_tokens[i]
 
         return ''.join(chars)
-
-    @classmethod
-    def string(cls, min_char=5, max_char=30):
-        """
-        Generates a random string of upper and lowercase letters.
-        :param min_char: Integer. Number of minimum characters
-        :param max_char: Integer. Number of maximum characters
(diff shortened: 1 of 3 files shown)


fcurella:
Thanks @tamarbuta 

I'd rather have this in the `pystr` provider: https://github.com/joke2k/faker/blob/master/faker/providers/python/__init__.py#L27

I think the `min_chars` should be optional and default to `None`. If it's `None`, the string will have the previous behavior: fixed length of `max_chars`

tamarbuta:
@fcurella Thanks for the feedback. Made the modifications to allow a `min_chars` specification in `pystr` and removed `string` from `misc`.

tamarbuta:
@fcurella Didn't mean to close the PR. I though "close" meant close the comment preview... Newbie mistake. :)

fcurella:
looks good!

can you add this quick test to https://github.com/joke2k/faker/blob/master/faker/tests/__init__.py#L528:

```python
characters = provider.pystr(min_chars=10, max_chars=255)
self.assertTrue((len(characters) >= 10)
```

tamarbuta:
@fcurella Added unit test.

fcurella:
Thanks @tamarbuta! Good work! ✨

PR #350, opened 2016-04-29
fix #346

### LINKED ISSUE #346: Add method to generate fake text with a minimum number of characters
Faker has methods to afford text strings with a maximum number of character, such as, `fake.text(max_nb_chars=200)`. Would like the ability to specify a minimum number of characters for a set of text or a specific length. This would allow users to easily input data into fields with minimum character input requirements. There is currently only two methods, `fake.password()` and `fake.ean()`, that allow the user to specify a length.

crf1028:
The function now has a minimum length. It is mainly composed of two parts, generator and slicer. If the generated string length is between `max_nb_chars` and `min_nb_chars`, then slicer would be skipped. Otherwise the string will be sliced to `max_nb_chars`.
The intention of a slicer is to solve the problem when `max_nb_chars - min_nb_chars` is too small, so that generated sting would fall out of scope.

### DIFF
--- a/faker/providers/lorem/__init__.py
+++ b/faker/providers/lorem/__init__.py
@@ -2,6 +2,7 @@ localized = True
 default_locale = 'la'
 
 from .. import BaseProvider
+import re
 
 
 class Provider(BaseProvider):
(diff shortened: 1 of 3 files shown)


faker/providers/lorem/__init__.py
```diff
@@ -81,50 +82,80 @@ def paragraphs(cls, nb=3):
...
     @classmethod
-    def text(cls, max_nb_chars=200):
+    def text(cls, min_nb_chars=5, max_nb_chars=200):
         """
         Generate a text string.
         Depending on the $maxNbChars, returns a string made of words, sentences, or paragraphs.
         :example 'Sapiente sunt omnis. Ut pariatur ad autem ducimus et. Voluptas rem voluptas sint modi dolorem amet.'
-        :param max_nb_chars Maximum number of characters the text should contain (minimum 5)
+        :param max_nb_chars Maximum number of characters the text should contain (inclusive; default=200)
+        :param min_nb_chars Minimum number of characters the text should contain (inclusive; default=0; minimum 5)
```
fcurella: the default is is `5` not `0`

faker/providers/lorem/__init__.py
```diff
@@ -81,50 +82,80 @@ def paragraphs(cls, nb=3):
...
+            raise ValueError('minimum length cannot exceed maximum length')
+        if min_nb_chars < 5:
             raise ValueError('text() can only generate text of at least 5 characters')
+        text_len = cls.random_int(min=min_nb_chars, max=max_nb_chars)
+        size = 0
 
-        if max_nb_chars < 25:
+        if text_len < 25:
             # join words
             while not text:
```
fcurella: i'm not sure this parent `while` is still needed. `text` will always be truth-y on the first iteration, the inner `while` loop should be all we need, unless I'm missing something

faker/tests/__init__.py
```diff
@@ -601,6 +601,31 @@ def test_ipv6(self):
...
+
+        for _ in range(999):
+            min_nb = int_provider.random_int(min=5, max=200)
+            max_nb = min_nb + int_provider.random_int(max=200)
+            lorem_text = text_provider.text(min_nb_chars=min_nb, max_nb_chars=max_nb)
+            self.assertTrue(max_nb >= len(lorem_text) >= min_nb)
+            self.assertTrue(lorem_text[-1] == '.')
+            self.assertFalse(re.findall(r'\.\s[A-Z][a-z]*[\.]*[\s]*$', lorem_text))
+            self.assertFalse(lorem_text[-2] == ' ')
+            self.assertTrue(all(map(lambda x: x[0].isupper(), lorem_text.split('. '))))
```
fcurella: those tests are a little obscure. Can you add some comments clarifying what it's checking?

fcurella:
closed for inactivity

PR #365, opened 2016-07-19
Make country code configurable

braintreeps:
# Problem

The base class `faker/providers/phone_number/__init__.py` provides a default phone number method for generating random numbers using the supplied format. Each locale `Provider` overrides the formats with those particular to that local.

We only want to use a subset of these formats; those with country codes and without extensions/

# Solution

This PR does 3 things (3 commits, best reviewed separately):

- 1. Creates new public methods for phone numbers explicitly with and without country codes, while maintaining functionality for the preexisting `phone_number` method (randomly selects with or without country code)

- 2. Removes extensions from the base formats (used in all public methods) and adds them back in only to the previously `phone_number`

- 3. Adds in missing locales


faker/providers/phone_number/bg_BG/__init__.py
```diff
@@ -4,16 +4,17 @@
 
 
 class Provider(PhoneNumberProvider):
+    country_code = '359'
```
cmonty: We lost the `+` here. We should add that back in.

cmonty: Oh nevermind, I see it's included in the concat.

PR #368, opened 2016-07-27
pydict docstring

navyad:
the docstring for method **pydict()** of python module has typo for word tuple as touple and also it is unclear, as it suggests that method will return tuple but it will always returns dict.


fcurella:
Thanks! ✨

Issue #355, opened 2016-05-18
Specify profile gender


lukemaxwell:
As profile provider names are now gender-specific, it would be great if the class accepted a parameter to allow the user to specify the gender of the profile, e.g profile = fake.profile(sex='M').

Is this within scope? I did a rudimentary test addition to providers/profile/__init__.py, (python 3) -  there was no issue.

fcurella:
That sounds reasonable and trivial to implement, but I'd like to iron out a few issues before going forward.

The first issues I see is that, at least in you description, gender and sex are conflated, which is not a reasonable assumption under some cultures. Since the providers already use 'female' and 'male', I would go with `sex` and be consistent in using `sex`, `female` and `male` throughout code and docs.

Another thing i'm not clear is if the parameter's possible values should be themselves translated to the provider language, eg: should the German provider accept M/F, or M/W? Obviously M/F for everything would be the easiest, but it would be great to have some feedback from non-English users on what _they_ prefer the API to look like.

lukemaxwell:
Thanks fcurella. I only used 'gender' arbitrarily in the issue description, that is confusing.

kmike:
A single set of constants is much better - it enables to switch language without changing source code.

fcurella:
i'm OK with having the arguments `M` or `F` regardless of language. @lukemaxwell could you submit your changes with a Pull Request?

navyad:
That is reasonable enhancement allows flexibility for creating profile. 

@lukemaxwell Have PR is created for this ?  if you are busy, then i can take this up.

navyad:
Hi @fcurella 

I check the source code suggested change is not there, so i have created the MR for that
https://github.com/joke2k/faker/pull/370

lukemaxwell:
Hey sorry I missed this. I'm travelling for the next few weeks so yes - by all means pick up :-) Otherwise I can when I'm back, late August.

L

L

navyad:
@lukemaxwell i have picked up.

fcurella:
closed in #370

PR #370, opened 2016-08-01
profile and simple_profile takes optional parameter sex

navyad:
This one is for issue https://github.com/joke2k/faker/issues/355.

Optional parameter '**sex**' added for` profile() `and `simple_profile()` allowed values are **'F'** and **'M'**

### DIFF
--- a/faker/providers/profile/__init__.py
+++ b/faker/providers/profile/__init__.py
@@ -33,7 +33,7 @@ class Provider(BaseProvider):
 
         }
 
-    def profile(self, sex=None, fields=None):
+    def profile(self, fields=None, sex=None):
         """
         Generates a complete profile.
         If "fields" is not empty, only the fields in the list will be returned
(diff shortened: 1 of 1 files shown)


navyad:
@coveralls  will add tests.

faker/providers/profile/__init__.py
```diff
@@ -32,7 +33,7 @@ def simple_profile(self):
 
         }
 
-    def profile(self, fields=None):
+    def profile(self, sex=None, fields=None):
```
fcurella: to maintain backward compatibility with code using the arguments positionally, the `sex` argument should be added _after_ `fields`

navyad:
@fcurella  following is fixed.

> to maintain backward compatibility with code using the arguments positionally, the sex argument should be added after fields

fcurella:
Thanks @navyad ! ✨

PR #371, opened 2016-08-01
Fix random_sample_unique

cecedille1:
There is a typo in random_sample_unique and it does not work.

I added a test method to ensure it works as expected


fcurella:
nice catch! Thank you very much! ✨

PR #373, opened 2016-08-04
Update RTD.org links to RTD.io

jeffwidman:
Also switched to https and terminated readme in a newline as recommended in PEP8


fcurella:
Thanks! Very much appreciated ✨

Issue #363, opened 2016-07-12
Remove randomness dependency on PYTHONHASHSEED


adamchainz:
We're using `faker` extensively in our test suite with `factory-boy` to generate dummy data. I love it! To help with repeatability in tests, I developed the pytest plugin `pytest-randomly` to allow re-using the `random` seed between test runs, so if they fail we can reuse the seed to get back the same random data from `faker` and debug the test.

Unfortunately we're moving from Python 2 to 3 and this includes activating `PYTHONHASHSEED` by default. We're currently trying to add it on Python 2 so that we're ready.

Since `PYTHONHASHSEED` means that dictionary iteration order is random, it means that anything that depends on this breaks. This is generally good as if you really depend on the iteration order, you should use a different data structure.

Unfortunately the way Faker generates _some_ random data (specifically I've found `random_element` when using a `dict`) is affected by dictionary iteration order. This means that fixing the random seed for the `random` module isn't enough, and one would need to fix `PYTHONHASHSEED` too. What a pain! It's also quite hard to do since there isn't a way inside a python process to query what the value of `PYTHONHASHSEED` is, unless it was explicitly set as an environment variable as opposed to set to `'random'`.

I think it would be better if Faker just didn't depend on dictionary iteration order, so its random data is also repeatable. This should be possible in most cases by swapping `dict` for `OrderedDict`.

fcurella:
Thank for the report @adamchainz

I think `random_element` method is good the way it currently is.

The issue is with specific providers that declare weighted elements as `dict`s instead of `OrderedDict`. We need to audit and fix them.

Additionally, I think this is something that should be mentioned in docs, including examples and docstrings.

adamchainz:
Just realized `OrderedDict` is 2.7+ but Faker still supports 2.6. Lists of tuples is probably the way to go here then? Unless you want to add a backport dependency for `OrderedDict` from somewhere.

kevin-brown:
You could always introduce a Python 2.6 dependency on the `ordereddict` package and fall back to it.

http://stackoverflow.com/q/1617078/359284

adamchainz:
Thanks @kevin-brown , probably the best solution

fcurella:
FWIW, dropping support for python 2.6 would also allows us to stop including importing `importlib` in [our requirements](https://github.com/joke2k/faker/blob/master/setup.py#L62).

jeffwidman:
[...]

PR #374, opened 2016-08-04
Drop Python 2.6 support

adamchainz:
Ref #363, required so we can use `OrderedDict` without an extra dependency, plus it's super old and insecure.


fcurella:
@joke2k If you're ok with dropping py2.6, I plan to merge this on Monday.

fcurella:
merged in 68b8ebad112a0a12f47b963ebd62a1abfef3b3ff

adamchainz:
Yay thx. I'll look into the `OrderedDict` fix now.

fcurella:
Thanks, very much appreciated :)

PR #377, opened 2016-08-08
Updated faker/providers/address/fi_FI/__init__.py

delneg:
Fixed utf8 coding issue with correct # -*- coding: utf-8 -*-


fcurella:
nice catch, Thanks! ✨

PR #379, opened 2016-08-23
Removed potentially offensive phrases

astawiarski:
Removed names that could potentially be found as offensive depending on the context.

I understand they're real names, in fact I know people having the surnames I removed in this commit. But this has caused a bit of trouble as we're assigning random names to players in our test environment.

I think those names should be removed just in case of potentially being offensive if they're used in the wrong context. Developers using this great library should not be worried about these things.


fcurella:
Thank you for bringing this up @astawiarski .

On one hand, I like the idea to have these names to potentially expose [Scunthorpe problems](https://en.wikipedia.org/wiki/Scunthorpe_problem).

On the other hand, the lists those names belongs to don't reference any source, so there isn't any factual data telling us how common they are.

Since there is no documented reason to have them included (ie no source), I'm in favor of removing them, in order to avoid any issue in the future.

PR #380, opened 2016-09-03
Fix for date_time_this_century function

JarUrb:
Function was returning datetime in current decade. Fixed it to return datetime in current century.


fcurella:
good catch, thanks! ✨

PR #389, opened 2016-09-20
Add date and time object providers

jtojnar:


fcurella:
thank you @jtojnar!

Looks good, could you add some tests?

jtojnar:
Are these tests enough? The date_time provider currently does not have any tests so I am not sure what exactly should I test.

fcurella:
@jtojnar those tests are good enough :)

thank you very much for your contribution! ✨

PR #390, opened 2016-09-20
Add unicode string provider

jtojnar:


jtojnar:
Any idea why does `test_documentor` fail?

MrCsabaToth:
Can someone tell me how to get unicode texts by the faker. Is there another provider if this one is rejected?

jtojnar:
Feel free to pick this up if this makes sense to you.

MrCsabaToth:
@jtojnar So there isn't any right now? Why this was not merged? Faker without the unicode texts is kind of surprises me.

jtojnar:
No idea. We were using fauxfactory in hamster-gtk and I wanted to use just faker https://github.com/jtojnar/hamster-gtk/commit/e1bd153b72b0eb8403b236feab383be53f3a8a9d but I no longer participate in its development.

PR #407, opened 2016-10-18
Update the PyPI version badge

illia-v:


fcurella:
Thanks! Good catch! ✨

PR #410, opened 2016-10-26
Turkish male names and Turkish surnames updated as usual usages

u-nux:
i edited because it genereates unusual Turkish names and surnames.


fcurella:
Hi @unux-git , do you have source for the new list of names?

unux-git:
i surfed on turkish name-surname websites,
scraped open data,
excluded bad words,
filtered common names,
then i generated this names and surnames

fcurella:
We need an authoritative source to point to so that we don't have to make any decision on what names get included or not. Maybe Turkey has an official census website where you can get them from?

unux-git:
official website:http://www.nvi.gov.tr (most of data not open)
open ones:
http://www.nvi.gov.tr/Files/File/Istatistikler/Isimler/en%20%C3%A7ok%20kullan%C4%B1lan%20ad.pdf
http://www.nvi.gov.tr/Files/File/Istatistikler/Isimler/en%20%C3%A7ok%20kullan%C4%B1lan%20soyad.pdf
http://www.nvi.gov.tr/Files/File/Istatistikler/Isimler/illere%20g%C3%B6re%20en%20%C3%A7ok%20kullan%C4%B1lan%20ad.pdf

another website we used:
http://www.mynet.com/annecocuk/populer-erkek-bebek-isimleri-ve-anlamlari-1181890

fcurella:
@unux-git The less we modify the data, the better.

I would stick to just using the governmental source for the first 2 links. I don't think we need to worry about the frequency of names per region, so disregard the 3rd link you posted as well.

Just get the first name female and male from the first link and the last names from the second link.

Remember to include those URLs in the code as a comment just above the lists.

unux-git:
no need to do this, i made this decisions with my two Turkish friends, and tested together. 
Before, code generates absurd Turkish names mostly and i want to help best way i do. nvi.gov.tr has no open data as many to help improve code. If you have different idea, i can try.

fcurella:
@unux-git  I'm not understanding what the issue is with the data you linked from nvi.gov.tr. I think their lists are good enough

unux-git:
I am ok with nvi's data, but i think it is not enough data to generate many names

PR #411, opened 2016-10-28
Add support to brazilian company IDs (CNPJ)

lamenezes:
This PR implments methods that provides pt-br company ids a.k.a. CNPJ [1] [2] 

[1] https://en.wikipedia.org/wiki/CNPJ
[2] https://pt.wikipedia.org/wiki/Cadastro_Nacional_da_Pessoa_Jur%C3%ADdica#Algoritmo_de_Valida.C3.A7.C3.A3o

### DIFF
--- a/faker/providers/company/pt_BR/__init__.py
+++ b/faker/providers/company/pt_BR/__init__.py
@@ -1,6 +1,5 @@
 # coding=utf-8
 from __future__ import unicode_literals
-import random
 from .. import Provider as CompanyProvider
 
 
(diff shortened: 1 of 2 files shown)


ladylovelace:
This would be very helpful.
Good collaboration @lamenezes

faker/providers/company/pt_BR/__init__.py
```diff
@@ -62,3 +80,15 @@ def catch_phrase(self):
         catch_phrase = self.generator.parse(pattern)
         catch_phrase = catch_phrase[0].upper() + catch_phrase[1:]
         return catch_phrase
+
+    @classmethod
+    def company_id(cls):
+        digits = random.sample(range(10), 8) + [0, 0, 0, 1]
```
fcurella: instead of the global `random` module, you should use the provider's own random instance, so that it can be frozen. Simply call `cls.random_sample`, instead of `random.sample`. The API is the same

fcurella:
Thank you very much! Great work! ✨

PR #414, opened 2016-10-30
Remove the warning about the `fake-factory` deprecation from `setup.py`

illia-v:
Since the package `fake-factory` is not being updated, should not the warning be removed?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,11 +15,6 @@ version = '0.7.3'
 
 package_name = 'Faker'
 
-if package_name == 'fake-factory':
-    print("""WARNING:
-    The `fake-factory` package is being deprecated on December 15th, 2016.
-    Use the `faker` package instead.""")
-
 # this module can be zip-safe if the zipimporter implements iter_modules or if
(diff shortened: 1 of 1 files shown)


fcurella:
I'm keeping `fake-factory` and `faker` both updated. They are the same code, but I'm always versioning `faker` one patch version higher to avoid confusion - ie: I can ask 'which version are you using?' and I know which package that refers to.

illia-v:
OK. But would not it be better if you put the warning only in `fake-factory`. Because it is slightly strangely that there is defining of the package name `faker` and below there is checking if the name is not `fake-factory`

fcurella:
@illia-v the warning is in both so I have one less difference between the two to remember about

fcurella:
@illia-v the warning is in both so I have one less difference between the two to remember about.
It's a little ugly, but it's temporary

fcurella:
removed the warning in 2fcae4ccd2329a1d2a97f356a0c5a0984c7611c9

PR #416, opened 2016-10-31
Added job providers for bs_BA

elahmo:
Added providers for jobs, taken from National statistics institute


fcurella:
Thank you! ✨

PR #417, opened 2016-11-01
Make the Windows build badge show info about build of the master branch

illia-v:
The badge have been showing information about builds of all contributions.
Because of it people could sometimes see deceptive information.


fcurella:
Good Catch, Thanks! ✨

PR #419, opened 2016-11-03
Secure and update some URLs

illia-v:


fcurella:
Thanks! HTTPS Everywhere! 💫

PR #420, opened 2016-11-04
Update the provider `user_agent`

illia-v:


faker/providers/user_agent/__init__.py
```diff
@@ -13,20 +13,39 @@
...
+        'Windows NT 6.2', 'Windows NT 6.3', 'Windows NT 10.0',
     )
 
     linux_processors = ('i686', 'x86_64',)
 
     mac_processors = ('Intel', 'PPC', 'U; Intel', 'U; PPC',)
 
-    langs = ('en-US', 'sl-SI', 'it-IT',)
+    # Language codes from ISO 639-1
+    langs = (
```
fcurella: from what I've seen, user agent strings usually use a locale (eg: `en-us`) rather than languages.

Maybe this can be refactored to make use of https://github.com/joke2k/faker/blob/master/faker/providers/misc/__init__.py#L71, and you can exapnd the list there?

faker/providers/user_agent/__init__.py
```diff
@@ -163,6 +182,13 @@ def internet_explorer(cls):
                             random.randint(0, 1))
 
     @classmethod
+    def microsoft_edge(cls):
+        tmplt = ('Mozilla/5.0 (Windows NT 10.0) AppleWebKit/537.36 (KHTML, '
```
fcurella: should this have some part randomized?

faker/providers/user_agent/__init__.py
fcurella: Using just language might be technically better, but I think the purpose of `Faker` is to generate data as close as possible to what the user may actually encounter in practice.

In other words, it's technically correct and considered best practice to use just the language, but how _common_  is it in practice for user agent string? Does it happen often?

For the same reason, I agree that the `locale` method needs to be fixed.

illia-v:
@fcurella, I won't add Microsoft Edge UA in the PR.

fcurella:
Thanks! ✨

Issue #401, opened 2016-10-07
pyfloat cant have 0 as digits


Benjamin-Setzer:
fake.pyfloat(0, 2, True)
-> 851420.54

But I expected: 0.XX

the code is as follows:
@classmethod
def pyfloat(cls, left_digits=None, right_digits=None, positive=False):
    left_digits = left_digits or cls.random_int(1, sys.float_info.dig)
    right_digits = right_digits or cls.random_int(0, sys.float_info.dig - left_digits)

compare to random_number:
@classmethod
def random_number(cls, digits=None):
    if digits is None:
        digits = BaseProvider.random_digit()

Basically, I expected pyfloat to behave like random_number and check if digits is None, else use the value I passed.

fcurella:
close by #428

Benjamin-Setzer:
Thank you

PR #428, opened 2016-11-20
Allow left/right_digits=0 for pyfloat (see issue #401)

mnalt:
Setting `left_digits=0` results in purely fractional numbers, allowing to
create e.g. random percentage values for test cases.

Setting `right_digits=0` results in integer values (represented as
floats), allowing to test if they are handled correctly or maybe
rounding is a problem for the use case under test.

Also see discussion for this [commit](https://github.com/illia-v/faker/commit/348a312e1a8378b9a168c7bf22fce8b5be500aad)


fcurella:
Seems like AppVeyor had some build issues on its own, but this LGTM.

Thank you!

PR #436, opened 2016-12-20
Fix no_NO postcodes.

kdeldycke:
Current implementation of postal codes for `no_NO` locale returns the schema instead of the value:
```python
>>> import faker
>>> faker.VERSION
'0.7.6'
>>> faker.Faker(locale='no_NO').postcode()
u'####'
```

This PR fix this issue.


fcurella:
Thank you for catching this! ✨

kdeldycke:
Thanks for the quick merge! :)

PR #445, opened 2017-01-09
Update __init__.py

vlad-ki:
Add new parameter in to classmethod 'random_number'. It's allow to create random number with fixed lenght.

### DIFF
--- a/faker/providers/__init__.py
+++ b/faker/providers/__init__.py
@@ -72,17 +72,24 @@ class BaseProvider(object):
             return ''
 
     @classmethod
-    def random_number(cls, digits=None):
+    def random_number(cls, digits=None, fix_len=False):
         """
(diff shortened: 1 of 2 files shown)


vlad-ki:
Hi, tell me please - where may I find the tests for classmethod 'random_number'. Sorry, I do not find it myself.

kevin-brown:
It looks like the tests are in [`faker.tests`](https://github.com/joke2k/faker/blob/c4d142734ba939ecc99e78cd46f9c34f8bca93fc/faker/tests/__init__.py) but there are not any for `random_number`. I suspect it wouldn't be too difficult to create a few though, given the fact that other similar tests exist.

vlad-ki:
Sure! Thanks for your answer.

fcurella:
Thank you! ✨

PR #451, opened 2017-01-27
Extend Factory Boy example

stephane:


stephane:
I find it painful to duplicate documentation between index.rst and README.rst.
Could you consider to reduce the size of the README.rst file (just an introduction and a quick example)?

What is the point of the FAKER title in ASCII art?
There is no requirements_dev.txt file to install sphinx package.
I would find useful to have a `requirements_dev.txt`at root of the project with tests and doc requirements.

fcurella:
> I find it painful to duplicate documentation between index.rst and README.rst.
> Could you consider to reduce the size of the README.rst file (just an introduction and a quick example)?

Sounds like a good idea

> What is the point of the FAKER title in ASCII art?

Branding, I suppose. That's a question for the original author, @joke2k 

> There is no requirements_dev.txt file to install sphinx package.
> I would find useful to have a requirements_dev.txtat root of the project with tests and doc requirements.

Sure, I don't see any reason why not :)

PR #461, opened 2017-02-16
Minor syntax update

an0o0nym:


fcurella:
Nice catch, thank you!

PR #462, opened 2017-02-17
Support for non-ascii fake data while maintaining readability for english speakers

chhabrakadabra:
**tl;dr;** I'd like to be able to run my test suite with non-ascii test data, while not using characters that most devs in the team can't read and words that most can't pronounce. The objective is to detect encoding errors in the code, while maintaining readability.

In the Bitbucket test suite, we use a [similar library](https://bitbucket.org/bitbucket/fakedata) (to faker) to generate the fake data. At the time I wrote that library, I wasn't aware of faker. One feature of that library I really find useful is its ability to optionally obfuscate strings. So the library can generate strings like "ợɼȧɳɠé", which are quite readable.

I thought it would be cool to add such a feature to faker and then move our test suite to faker (and deprecate fakedata). I'm not sure if you would be open to integrating my changes, given that this is an esoteric feature, but I'd appreciate it if you did.

To better illustrate this change, I'm pasting this Python session:

```python
>>> from faker import Factory
>>> fake = Factory.create('ma_MA')
>>> print fake.address()
50088 Ƙạṙᶖ Ṁîℓľ Ȧƥƫ. 590
Łḁǩễ Ṙȏἠἀŀɖ, ȘČ 49183-6904
>>> print fake.state()
Ӥȇᶀṙàŝķἇ
>>> print fake.name()
Ƌṛ. Ťȭñȳ 𝕄ŏřᵷåɲ
>>> print fake.url()
http://ṙổᵬɨἤşồή-ĵȭḫǹṧőἡ.óṛģ/
>>> print fake.job()
ℌáếḿầţǒɭǭǧïᵴṯ
>>> print fake.sentence()
Ċổᵯṁṏɖĩ άǖȶ ʠửåᶊ ὴơṥṱŗựḿ.
>>> print fake.bs()
ḟαɕïļȋţāŧé ᵬŗïçʞṧ-ἅňᶁ-ȼƚĩ¢ǩš ᵱᶅḁťƒöṝᶆȿ
```

Thank you for your time.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -146,6 +146,7 @@ Included localized providers:
 -  `ko\_KR <https://faker.readthedocs.io/en/master/locales/ko_KR.html>`__ - Korean
 -  `lt\_LT <https://faker.readthedocs.io/en/master/locales/lt_LT.html>`__ - Lithuanian
 -  `lv\_LV <https://faker.readthedocs.io/en/master/locales/lv_LV.html>`__ - Latvian
+-  `ma\_MA <https://faker.readthedocs.io/en/master/locales/ma_MA.html>`__ - Martian (Mars)
 -  `ne\_NP <https://faker.readthedocs.io/en/master/locales/ne_NP.html>`__ - Nepali
 -  `nl\_NL <https://faker.readthedocs.io/en/master/locales/nl_NL.html>`__ - Dutch (Netherlands)
 -  `no\_NO <https://faker.readthedocs.io/en/master/locales/no_NO.html>`__ - Norwegian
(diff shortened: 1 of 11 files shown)


chhabrakadabra:
Upon further reflection, I've decided to use a mixture of the various locales provided by faker in the codebase. I guess I wrote that feature for the fakedata library because I did not have access to high quality fake data for other locales, but with the kind of data available to this library, perhaps this approach is just not as useful. So if you show interest, I'd be happy to finish this change, but for now I won't be using this change in my own codebase.

fcurella:
Thanks @chhabrakadabra , your provider looks very cool!

My only issue is with using the non-standard `ma_MA` locale code. If there was a 'private use' locale code already standardized, we could use that. But until then, I'll have to pass on implementing providers for non-recognized languages.

For reference, I'm using [this list](https://www.loc.gov/standards/iso639-2/php/code_list.php) as source for valid local codes

davidszotten:
was this replaced by something else?

Issue #469, opened 2017-02-27
Add ISBN support


barseghyanartur:
Please, add [ISBN](https://en.wikipedia.org/wiki/International_Standard_Book_Number) support.

fcurella:
Hi @barseghyanartur ,

I think that would be a useful addition. Feel free to submit a Pull Request! :)

grantbachman:
Is it okay if I work on this one? Never done OSS work before but I've been looking for a project to which I can contribute, and faker looks like a great way to get started!

fcurella:
@grantbachman Absolutely! Thank you so much for volunteering :)

If you have any question, don't hesitate to ask :)

grantbachman:
Hey @fcurella I have basic ISBN-10/ISBN-13 number support working, meaning I can generate basic 10/13 digit numbers that meet the ISBN-10/13 check digit requirements. However things start to get complicated when we talk about hyphenating these numbers (e.g. turning `9783161484100` into `978-3-16-148410-0`). Doing this requires referencing the [ISBN rules](https://www.isbn-international.org/range_file_generation) to identify which language/region the number falls under, then looking up the hyphenation rules for that region. These rules appear to change over time.

My question becomes, how real do we want these numbers to look? I could simply code in the rules for english ISBNs and call it a day, or we could attempt to follow the rules of all of the regions by either storing them somewhere in faker or letting a separate python package (like [python-stdnum](https://github.com/arthurdejong/python-stdnum)) handle that for us. Thoughts?

fcurella:
@grantbachman Thank you! I think we can start with just `en_US`. We can always add more countries later, if any requests come in.

grantbachman:
What's the policy on closing issues? Should this be closed right away because it was merged or do we wait a period before doing so?

fcurella:
This should be closed upon merge. Thanks @grantbachman for reminding me

PR #471, opened 2017-03-09
Implement ISBN

grantbachman:
Per discussion in #469 this PR allows faker to support the generation of valid ISBNs. See [ISBN International](https://www.isbn-international.org/content/what-isbn) for an overview of the individual numbers that make up a valid ISBN.
 
An important thing to consider with ISBN is that the hyphenation rules are dynamic, numerous, and subject to change by ISBN International. This PR only implements the ruleset for English language books. If we are interested in getting fancy, we can always utilize other python projects that actually keep up-to-date with information like this.

Example usage:
```python
from faker import Factory
fake = Factory.create()

fake.isbn10()  # '1-67423-180-6'
fake.isbn13()  # '978-0-584-45740-7'
fake.isbn13(separator=' ')  # '978 1 917745 78 9'
```

This is my first ever PR to an open source project, so let me know if anything is out of order.

### DIFF
--- a/faker/providers/isbn/__init__.py
+++ b/faker/providers/isbn/__init__.py
@@ -2,7 +2,7 @@
 
 from __future__ import unicode_literals
 from .. import BaseProvider
-from .isbn import ISBN, ISBN13
+from .isbn import ISBN, ISBN10, ISBN13
 from .rules import RULES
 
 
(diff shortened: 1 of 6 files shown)


fcurella:
Nice Work! ✨ 🍰

fcurella:
@grantbachman Very nice PR! Descriptive, well documented and includes tests! Impressive :)

grantbachman:
Thanks for the kind words @fcurella!

PR #472, opened 2017-03-10
Consolidate tests into single directory

grantbachman:
This PR just moves all of the tests into the single `tests/` directory at the project root. There was a single test not following this convention.

@fcurella I hope you don't mind me firing off small little commits like this, I see a bunch of places where I think the code can be made a little cleaner.


fcurella:
@grantbachman I don't mind :) On the contrary, I appreciate your help 👍

Issue #474, opened 2017-03-11
Improve README for Google?


grantbachman:
When I search for this repo on Google, the large ASCII logo gets lumped into the description. For visibility's sake would we want to consider removing this logo, turning it into an image, or moving it down the README so it's not the very first thing listed?

My search results:
<img width="661" alt="screen shot 2017-03-10 at 9 13 36 pm" src="https://cloud.githubusercontent.com/assets/499564/23819612/8581b3e4-05d6-11e7-9de7-696e8390aa76.png">

fcurella:
Nice catch, I haven't thought of that!

I would say let's move it down, right above the CI badges (~ L17) should do :)

kevin-brown:
It looks like it was handled in #475 and can be closed.

PR #475, opened 2017-03-11
Move logo further down the README

grantbachman:
Google mangles the ASCII logo when displaying README in search results.

Refs #474


fcurella:
Thanks! ✨

PR #479, opened 2017-03-13
add provider en_TH, th_TH

ppeemk:
am i doing it right? i've added en_TH and th_TH person provider but i don't know if i'm adding to the right branch or not


faker/providers/person/th_TH/__init__.py
```diff
@@ -0,0 +1,90 @@
+# coding = 'utf8'
```
fcurella: try `# -*- coding: utf-8 -*-` instead

fcurella:
@ppeemk You targeted the wrong branch, but I've fixed the PR for you

ppeemk:
@fcurella Do you want me to adjust anything else?

fcurella:
@ppeemk No, that's perfect. Thank you so much for your contribution! ✨

fcurella:
@ppeemk do you have a source for your data?

ppeemk:
it's my friend name and i'm write it on my own. Does that count as no source ?

fcurella:
@ppeemk No. We try to use some 'official' source. If you can get your data from wikipedia or from an official government supported statistical agency, that will work.

I'm going to revert the merge until we can find a reliable source.

ppeemk:
Oh ok then, i will fixed it up and submitted to you later on. Where should i comment the source then. At the top of the file?

fcurella:
Preferably at the top of each list. If they all come from the same page, you can put a single comment under the class declaration. Example: https://github.com/joke2k/faker/blob/master/faker/providers/person/en_US/__init__.py

PR #481, opened 2017-03-14
Add french words for days and months

sblondon:
The patch provides translation to french for days and months:

```
>>> from faker import Factory
>>> fake = Factory.create('fr_FR')
>>> fake.month_name()
'Mai'
>>> fake.day_of_week()
'Mercredi'
```


fcurella:
Thanks! ✨

Issue #476, opened 2017-03-11
Break up tests by provider instead of by language


grantbachman:
Looking through the tests directory, I see that the tests are mainly broken up by language. To me, this isn't very discoverable in that I have no idea what's actually under test in each of those directories. Would it make more sense to break them up by their individual functions?

For instance, something like this:
```
tests/
    providers/
        isbn.py
        ssn.py
        postcode.py
   generator.py
   documentor.py
```

fcurella:
yes, it would make more sense to break them down. I meant to do that, but haven't had the chance yet.

grantbachman:
I can handle it. I quite enjoy mucking with tests :) I can get started tonight.

PR #482, opened 2017-03-16
Reorganize tests

grantbachman:
This PR reorganizes tests per the discussion in #476.

1.  Instead of all of the locale-based provider tests being under locale-specific directories, they are moved into their appropriate provider-based modules. This new org method more closely mimics the directory structure of the functional code so things are easier to find.

2. I moved all of the tests out of `tests/__init__.py` and into their appropriate factory/generator/utils module.

Other than moving all of the tests around, not much was changed (except a wrong test I uncovered in fe167bfd25c0c86b3c6fb5ef76eb24036ad2b6da which has been fixed)


fcurella:
Nice Job! 💯 👍 

Thank you so much! ✨

PR #485, opened 2017-03-18
Fix no_NO building numbers from address provider

nikolaik:
Fixes unparsed building number suffixes like `Drammensveien 1?`
and improves them with an ok distribution of most used suffix letters.


fcurella:
Looks good! Thank you ✨

PR #494, opened 2017-04-06
[close #493] adding_file_path_provider

### LINKED ISSUE #493: Add a file path provider
In the file providers would be nice to have a file_path provider who would return a path like ```/lorem/ipsum/lorem.pdf```.

diegommarino:
Issue #493 

Adding a file path provider in file provider.


fcurella:
Thank you!

PR #495, opened 2017-04-10
Add Japanese lorem provider

richmondwang:
Created the `sentence_punctuation` on `Provider` because Japanese period (punctuation mark) uses different symbol. And I think some languages might also have different period symbols.

Reference: https://www.tofugu.com/japanese/japanese-punctuation/#japanese-punctuation-marks


fcurella:
Thank you! ✨

PR #499, opened 2017-04-13
Added ValueError unit test for lorem

Newman101:
Hi.

File faker/providers/lorem/__init__.py had a missing unit test for checking that if ```max_nb_chars``` is below 5, ```ValueError``` is raised. 

This unit test has been now added and coverage improved. 

```python setup.py test``` executed locally - all green.

Thank you.


fcurella:
Thank you! ✨

PR #501, opened 2017-04-19
Georgian person provider

GeorgeLubaretsi:
Added fake person provider for Georgian names.

The data was obtained from here:
http://mashasada.me/en/chamotvirtva

Which is based on the voter list.


fcurella:
Thanks! ✨

GeorgeLubaretsi:
Sure! PR for Georgian addresses is on the way!

PR #503, opened 2017-04-20
Added unit test for new_date method

Newman101:
Hi. 

Method ```new_date``` in ```datetime_safe``` module was missing an unit test. This has been now added. 

```python setup.py test``` executed locally - all green.

Thanks.


fcurella:
Thanks! :Sparkl

Issue #490, opened 2017-03-23
Improve accuracy of hu_HU generators


chrisvoncsefalvay:
The Hungarian 🇭🇺 generator (`hu_HU`) exhibits some anomalies:

- Domains and e-mail addresses, which derive their content from the first and last name generators, contain accents. In general, it is uncommon for Hungarian domains to contain accents, and in most cases, these are simply outright skipped.
- Fake month names are capitalised. In Hungarian, month names are non-capitalised.
- Name in credit card generator has 'Western' naming order
- `fake.street_address()`, `fake.address()` and `fake.street_address_with_county()` return a capitalised result
- `fake.street_address_with_county()` is in a wrong format
- `fake.color_name()` returns English result
- `fake.company_suffix()`, `fake.bs()` and `fake.catch_phrase()` return English results
- `fake.currency_code()` should return `HUF`
- `fake.isbn()` should be Hungarian, not 978 (the fictitious `Bookland`)
- `fake.job()` is still in English
- `faker.providers.lorem` are still in English
- `faker.providers.profile` should generate coherent name-mail-sex-username combos!
-  telephone numbers should use the standard `(20) 123-4567`, `(1) 234-5678` or `(21) 234-567` formats only

Newman101:
Hi. Thanks for the feedback! I'm running some tests today and we'll see what can be improved. =)

Newman101:
```fake.company_suffix()``` returns now Hungarian results.

fcurella:
Closed By #504

fcurella:
reopening, all other Hungarian providers still have the mentioned issues

Newman101:
Accents have been removed from domains and e-mail addresses.

Newman101:
Months + days of the week implemented.

Newman101:
Task list: 

- [x] E-mail addresses and domains
- [x] Month names
- [ ] Credit card generator
- [ ] Capitalize street addresses
- [ ] Capitalize addresses with a country
- [ ] Localize colors
- [x] Localize fake.company_suffix()
- [ ] Localize fake.bs() and fake.catch_phrase()
- [ ] Localize fake.street_address_with_county()
- [ ] Localize fake.currency_code()
- [ ] Localize fake.isbn()
- [ ] Localize fake.job()
- [ ] Localize faker.providers.profile
- [ ] Localize telephone numbers

Newman101:
[...]

PR #504, opened 2017-04-22
Added company provider to hu_HU locale

Newman101:
Related to issue #490. 

Looks like ```hu_HU``` locale was missing the company localisation. As a consequence, ```fake.company_suffix()``` was returning English results. 

This has now been fixed with unit tests.

```python setup.py test``` executed locally. All green.

Thank you.


fcurella:
Thank you! ✨

Issue #505, opened 2017-04-23
Allow subdomains in domain_name()


thiagofigueiro:
Sometimes it's useful to test domain names that contain subdomains.  Currently, domain_name() can only return domain names composed of domain_word + . + domain_tld.

I propose to modify domain name to accept an optional parameter to specify the domain depth.  Example:
```python
        >>> domain_name()
        nichols-phillips.com
        >>> domain_name(2)
        williamson-hopkins.jackson.com
```

joke2k:
Implemented in #506

PR #506, opened 2017-04-23
Allow subdomains for domain_name

thiagofigueiro:
Implements #505 

Please let me know if you'd like tests for this specific functionality.  I had a look at the existing tests and it doesn't look like it's common practice in this project to test parameters.

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -97,8 +97,22 @@ class Provider(BaseProvider):
         )
         return username
 
-    def domain_name(self):
-        return self.domain_word() + '.' + self.tld()
+    def domain_name(self, levels=1):
+        """
+        Produce an Internet domain name with the specified number of
+        subdomain levels.
+
+        >>> domain_name()
+        nichols-phillips.com
+        >>> domain_name(2)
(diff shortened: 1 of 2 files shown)


faker/providers/internet/__init__.py
```diff
@@ -97,8 +97,20 @@ def user_name(self):
...
+        """
+        Produce an Internet domain name with the specified number of
+        subdomain levels.
+
+        >>> domain_name()
+        nichols-phillips.com
+        >>> domain_name(2)
+        williamson-hopkins.jackson.com
+        """
+        if 1 >= levels:
```
fcurella: just a stylistic note: could change this to be `if levels == 1:`

fcurella:
@thiagofigueiro more tests are always good. Feel free to add some :)

faker/providers/internet/__init__.py
thiagofigueiro: I changed and added a check to avoid an infinite recursion.

thiagofigueiro:
Sorry, @fcurella - do you mind reviewing again?  I added a couple of tests.

fcurella:
looks good! Thank you! ✨

PR #507, opened 2017-04-24
Fix typo

m-allam:
Replace unicode '×' symbol that causes encoding errors with DRF (Django Rest Framework) API


fcurella:
Thanks! ✨

PR #515, opened 2017-05-19
add cakeipsum

guinslym:
I added another [lorem](http://www.cupcakeipsum.com/#/paragraphs/5/length/long/with_love/false/start_with_cupcake/false/seed/9a16b91036d2b501499c9081a70315283ab74daba0bfce6bf1c6dff53499a025)  . I had to prototype a recipe website and I found this lorem

>>> from faker import Factory
>>> fake = Factory.create()
>>> fake.text()
'Maxime rem non suscipit temporibus asperiores quidem. Nesciunt quos molestias voluptatibus. Pariatur minus aut maiores quo consectetur expedita soluta.'

>>> fake = Factory.create('cakeipsum')
>>> fake.text()
'Powder snaps bar Ice Topping croissant croissant Cake dragée. Croissant cake beans Jelly Lollipop jujubes donut. Chups jelly Danish chocolate brownie. Marshmallow Jelly toffee biscuit.'
>>> fake.name()
'Darrell Neal'


fcurella:
Hi @guinslym , thanks for your contribution!

I don't really think this case is generic enough to be useful to other users.

I'd rather see a PR that makes the `lorem` provider configurable by accepting a list of words. Is that something you're willing to work on?

guinslym:
Hi @fcurella ! Thanks for the comment. I will work on that.

faker/providers/lorem/la/__init__.py
```diff
@@ -5,7 +5,17 @@
 
 
 class Provider(LoremProvider):
-    word_list = (
+
+    def __init__(self, ext_word_list=None):
+        #verify the type of the variable
```
kevin-brown: Wouldn't it make more sense to keep the existing `word_list` variable and then override it based on whether or not `ext_word_list` was provided?

```python
class Provider(LoremProvider):

    def __init__(self, ext_word_list=None):
        if ext_word_list is not None:
            self.word_list = ext_word_list

    word_list = (
        # existing list of words
```

guinslym: Thanks your are right @kevin-brown it's simpler.

PR #518, opened 2017-05-28
Add et_EE (Estonian) provider: names and ssn

trtd:
Created the provider for Estonian language (et_EE).
Providers implemented for names and ssn.


fcurella:
Nice work! Thank you! ✨

PR #523, opened 2017-05-30
Make static data accessible at class level in pl_PL DateTime provider

zgoda:
This cleans tests by giving direct access to dataset used to generate results


fcurella:
+1 Much cleaner.

Thank you!

Issue #513, opened 2017-05-16
Possible bug in datetime


Newman101:
Hi. 

Looks like one of the unit tests is randomly failing: 

```
======================================================================
FAIL: test_date_time_between_dates_with_tzinfo (tests.factory.FactoryTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jremes/Projects/Python/faker-2/tests/factory.py", line 284, in test_date_time_between_dates_with_tzinfo
    self.assertTrue(datetime_start <= random_date)
AssertionError: False is not true

----------------------------------------------------------------------
Ran 80 tests in 12.309s
```

If needed, I can have a look at this issue. 

Thanks.

fcurella:
Thanks for the report @Newman101 . If you have time to look into, I'd definitely appreciate it :)

fcurella:
Closed in #524

PR #524, opened 2017-05-31
Fix for issue #513 in test test_date_time_between_dates_with_tzinfo

marcustas:
Problem was in function datetime_to_timestamp from `faker/providers/date_time/__init__.py`.
`time.mktime()` assumes that the passed tuple is in local time and `calendar.timegm()` assumes it's in UTC.
 After fix in this commit there are no such issue.

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -4,11 +4,11 @@ from __future__ import unicode_literals
 
 from datetime import timedelta
 import re
-from time import time, mktime
+from time import time
 from calendar import timegm
 
 from dateutil import relativedelta
-from dateutil.tz import tzlocal
+from dateutil.tz import tzutc
 
 from faker.generator import random
 from faker.utils.datetime_safe import date, datetime, real_date, real_datetime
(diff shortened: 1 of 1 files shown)


faker/providers/date_time/__init__.py
```diff
@@ -20,6 +21,8 @@
 
 def datetime_to_timestamp(dt):
     if getattr(dt, 'tzinfo', None) is not None:
+        if getattr(dt, 'tzinfo', 'UTC'):
```
fcurella: we already checked that `dt.tzinfo` exists and is not None.

The only way L25 will not be executed is if `dt.tzinfo` is either `False` or `0`, which I is never the case, meaning L25 will always be executed and this `if` is unnecessary

marcustas:
Hi, I've checked this issue one more time. And I see error in test described in issue #513 in some cases.
I found that I see this bug for my timezone (+3) for years from 1930 till 1989. Seems it's related with "decret time"(timezone time +1 hour) which was implemented in USSR from 1930.
So instead using `mktime` I propose to convert datetime to UTC, and then use `timegm` to convert to timestamp.
If it's acceptable, I can make pull request.
Thx.

faker/providers/date_time/__init__.py
marcustas: Hi, I've checked this issue one more time. And I see error in test described in issue #513 in some cases.
I found that I see this bug for my timezone (+3) for years from 1930 till 1989. Seems it's related with "decret time"(timezone time +1 hour) which was implemented in USSR from 1930.
So instead using `mktime` I propose to convert datetime to UTC, and then use `timegm` to convert to timestamp.
If it's acceptable, I can make pull request.
Thx.

fcurella:
Thanks! Looks good ✨

PR #528, opened 2017-06-03
A configurable lorem provider

guinslym:
Following the recommendation of @fcurella, [PR#515](https://github.com/joke2k/faker/pull/515) , I am providing a way so that the user can provide it's own set of words to the **lorem** provider.  I added also a test named **test_ext_word_list** .  

In `lorem/__init__.py` I created a params call **ext_word_list** that stands for **external word list** so that the user can provide a list of words to use with the lorem provider.

As English is not my main language the maintainer can change the parameter **ext_word_list** to something else

### DIFF
--- a/faker/providers/lorem/__init__.py
+++ b/faker/providers/lorem/__init__.py
@@ -1,8 +1,6 @@
 localized = True
 #default_locale is 'en_US' in the previous State of this application
 default_locale = 'la' 
-#external provider
-external_provider = ''
 
 from .. import BaseProvider
 
@@ -42,9 +40,7 @@ class Provider(BaseProvider):
         'ext_word_list' --- a list of word you would like to have 
                  instead of 'Lorem ipsum'
         """
-        if ext_word_list:
-            return [cls.word(ext_word_list) for _ in range(0, nb)]
-        return [cls.word() for _ in range(0, nb)]
+        return [cls.word(ext_word_list) for _ in range(0, nb)]
 
     @classmethod
     def sentence(cls, nb_words=6, variable_nb_words=True, ext_word_list=None):
@@ -65,10 +61,7 @@ class Provider(BaseProvider):
         if variable_nb_words:
             nb_words = cls.randomize_nb_elements(nb_words)
 
-        if ext_word_list:
-            words = cls.words(nb_words, ext_word_list)
-        else:
-            words = cls.words(nb_words)
+        words = cls.words(nb=nb_words, ext_world_list=ext_world_list)
         words[0] = words[0].title()
 
         return cls.word_connector.join(words) + cls.sentence_punctuation
--- a/faker/providers/lorem/__init__.py
+++ b/faker/providers/lorem/__init__.py
@@ -94,9 +94,7 @@ class Provider(BaseProvider):
 
         :return list
         """
-        if ext_word_list:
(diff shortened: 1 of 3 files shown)


faker/providers/lorem/__init__.py
```diff
@@ -1,67 +1,108 @@
...
         :example array('Lorem', 'ipsum', 'dolor')
-        :param nb how many words to return
+
+        Keyword arguments:
+        nb --- how many words to return
+        'ext_word_list' --- a list of word you would like to have 
+                 instead of 'Lorem ipsum'
         """
+        if ext_word_list:
+            return [cls.word(ext_word_list) for _ in range(0, nb)]
```
fcurella: you don't need this `if` clause, as `.word()` takes care of it already.

Simply modify the line below to call `[cls.word(ext_word_list) for _ in range(0, nb)]`

faker/providers/lorem/__init__.py
```diff
@@ -1,67 +1,108 @@
...
             return ''
 
         if variable_nb_words:
             nb_words = cls.randomize_nb_elements(nb_words)
 
-        words = cls.words(nb_words)
+        if ext_word_list:
+            words = cls.words(nb_words, ext_word_list)
+        else:
+            words = cls.words(nb_words)
```
fcurella: no need to check for `ext_word_list`. Just pass everything: `words = cls.words(nb=nb_words, ext_world_list=ext_world_list)`

faker/providers/lorem/__init__.py
```diff
@@ -1,67 +1,108 @@
 localized = True
-default_locale = 'la'
+#default_locale is 'en_US' in the previous State of this application
+default_locale = 'la' 
+#external provider
+external_provider = ''
```
fcurella: what is the purpose of this line?

guinslym: Error from me I will remove it

faker/providers/lorem/__init__.py
guinslym: Thanks! Done!

faker/providers/lorem/__init__.py
guinslym: Thanks! Done

guinslym:
will open a new pull request with the changes.

fcurella:
Thank you!  ✨

shacker:
I can see that this is configurable to use an external word list, but does that mean that I cannot generate a fake sentence in English without bringing in a separate word list? Looking through this code and the docs, and trying things like:

```
fake.text(lang='en_US')
fake.text(default_locale='en')
fake.text(ext_word_list='en_US')
fake.text(default_locale='en_US')
```

Nothing seems to work. If there is / is not a way to generate text/sentences in English, using the built-in language files, can it be documented? Thanks.

fcurella:
@shacker By default it uses the word list defined in one of these localized providers: https://github.com/joke2k/faker/tree/master/faker/providers/lorem. If you didn't configure `Faker` to one of these, it will default to `la`.

In practical terms, this means that the `lorem` doesn't support `en_US` out of the box. We will someone to submit a PR that add a `lorem.en_US` provider containing a list of words to be used as default.

shacker:
Got it, thanks. I may just submit a PR sometime soon. Cheers.

Issue #488, opened 2017-03-20
Not possible to seed .binary()


J535D165:
It is not possible to seed the method ``binary`` (see also issue #484)

```
>>> from faker import Faker
>>> f1 = Faker()
>>> f1.seed(4321)
>>> print(f1.binary()[0:100])
b'm\x17 ;\xca\x88\xb6\xb1\x8c/6Qmo\x02\xfa\xc3g\xf9R\xc5\xc6\x84\n\xac\xa3\x99\xb9Xs\xc3D\xa5g\xbb\xe5YU)\x95\tt\x0fsy\xe3\xaa\xfc\xaa\xb7\xb9\xfd\xac\x88\x13\xe2\xb1!\xbc\xea\x00\xd3]"\x9c;\xf2\xecx\xee\xe1\x06\xa6_\x9by\x97\xc3\xd7\xf6\x87&\x8fM\x90{\x18\xb0/\xdf\xcf$*\xde\x13\xc1\x96&\x12\xc2'
>>> f2 = Faker()
>>> f2.seed(4321)
>>> print(f2.binary()[0:100])
b"\xc1Xf\x81\xba\xe8P\xcb\xb2\xfd\x8e\x8bz7'\xe0\x7f\xef\xdej\xf5\xb8\xf1\xe4\x97v\x7f\x9aM\xfb\xee\xabT\x17\x9c\xb7\xa6\x8f\xd1\x8c\x07l0\xc5&*\xd8\xf2B\xbf\xbe\xe8\xca\xbe\xe8\xdf\xec4\xd2\xc1\xff\xecI\xc9\xdc\xae\x92~|\xa1\x0ch\x08\\G\xd7\xd8\xf0I\xc0\xcaLR\xaa\xd6{%\xfa\x14\xd9C\x9f\x1fK\x14\xc0s?O)"
```

PR #530, opened 2017-06-07
generate blob using random instead of os.urandom, making it reproduceable

magicgoose:
fix for https://github.com/joke2k/faker/issues/488

### DIFF
--- a/faker/providers/misc/__init__.py
+++ b/faker/providers/misc/__init__.py
@@ -91,7 +91,7 @@ class Provider(BaseProvider):
 
         Default blob size is 1 Mb.
         """
-        return os.urandom(length)
+        return random.getrandbits(8 * length).to_bytes(length, byteorder='big', signed=False)
 
     @classmethod
     def md5(cls, raw_output=False):


faker/providers/misc/__init__.py
```diff
@@ -91,7 +91,7 @@ def binary(cls, length=(1 * 1024 * 1024)):
 
         Default blob size is 1 Mb.
         """
-        return os.urandom(length)
```
kevin-brown: `int` didn't gain a `to_bytes` method until Python 3, and this project still supports Python 2.7.

kevin-brown: It also fails [this test for binary](https://github.com/joke2k/faker/blob/026e0577696ff21732e0c3ac529bbce30d723f0b/tests/factory.py#L358-L366) because `os.urandom()` supports having a length of `0` (which we consider valid) while `getrandbits()` doesn't support getting `0` random bits.

magicgoose:
OK, I will come up with something, stay tuned

fcurella:
done in #536

PR #534, opened 2017-06-15
Generate proper dates before 1970

kungfu71186:
Using timestamp doesn't allow you to generate a date before 1970. In fact, you can even put in (-100y,-90y) as parameters and get a runtime error.


fcurella:
Looks good! Thank you!

PR #535, opened 2017-06-16
Generate blob

### LINKED ISSUE #488: Not possible to seed .binary()

kungfu71186:
Fixes #488 and #530

Test cases fail due to 
six.binary_type
Type for representing binary data. This is str in Python 2 and bytes in Python 3.

Where python 2 uses str. which is technically correct, but not in this case. Could use isinstance(binary,bytearray) or isinstance(binary,bytes) instead. Could possible do a try decode on it.

try
binary.decode(errors='ignore')

### DIFF
--- a/faker/providers/misc/__init__.py
+++ b/faker/providers/misc/__init__.py
@@ -5,6 +5,7 @@ import hashlib
 import string
 import uuid
 import os
+import sys
 
 from faker.generator import random
 from faker.providers.date_time import Provider as DatetimeProvider
@@ -91,7 +92,8 @@ class Provider(BaseProvider):
 
         Default blob size is 1 Mb.
         """
-        return bytes([random.randrange(256) for o in range(length)])
+        blob = [random.randrange(256) for o in range(length)]
+        return bytes(blob) if sys.version_info[0] >= 3 else bytearray(blob)
 
     @classmethod
     def md5(cls, raw_output=False):
--- a/faker/providers/misc/__init__.py
+++ b/faker/providers/misc/__init__.py
@@ -91,7 +91,7 @@ class Provider(BaseProvider):
 
         Default blob size is 1 Mb.
         """
-        return os.urandom(length)
+        return bytes([random.randrange(256) for o in range(length)])
 
     @classmethod
     def md5(cls, raw_output=False):


faker/providers/misc/__init__.py
```diff
@@ -91,7 +92,8 @@ def binary(cls, length=(1 * 1024 * 1024)):
 
         Default blob size is 1 Mb.
         """
-        return os.urandom(length)
+        blob = [random.randrange(256) for o in range(length)]
+        return bytes(blob) if sys.version_info[0] >= 3 else bytearray(blob)
```
fcurella: wouldn't it work if we'd just `return bytes(blob)` for all python versions?

kungfu71186: not really. Python 2 bytes is an alias to str. So if you get 1 byte it will generate something like '[87]' (length 4 str).

Example:
===Python 3====
>>> byteArr = [23,67,75,254]
>>> bytes(byteArr)
b'\x17CK\xfe'
>>> len(bytes(byteArr))
4

===Python 2====
>>> byteArr = [23,67,75,254]
>>> bytes(byteArr)
'[23, 67, 75, 254]'
>>> len(bytes(byteArr))
17

>>> bytearray(byteArr)
bytearray(b'\x17CK\xfe')
>>> len(bytearray(byteArr))
4

fcurella: I see. Then I think we can just fix the test by checking for `isinstance(binary, (bytes, bytesarray))`

kungfu71186: Sounds good, yeah wasn't sure if we should keep it to just bytes or not

PR #538, opened 2017-06-18
Adding an implementation of `ssn` provider for the pl_PL locale.

pdaw:
Reference: https://en.wikipedia.org/wiki/National_identification_number#Poland


fcurella:
looks good! Thank you! ✨

PR #539, opened 2017-06-18
Added pt_BR colors localization.

ppcmiranda:
I saw there were no brazilian portuguese localization for color names, so I added one. I tried to stick to the guidelines, but, if something is wrong, I can quickly fix it.


fcurella:
Looks good! Thank you! ✨

Issue #541, opened 2017-06-22
hu_HU name provider has a bug in one of the female name formats


swilcox:
When randomly generating female names, it's possible to get output that includes the text `{{last_name})}` as part of the output. The extra `)` is causing the problem.

I am preparing a PR for this issue.

fcurella:
Closed by #542

PR #542, opened 2017-06-22
fixed female name format typo (issue #541)

swilcox:
fixed typo that could result in `{{last_name})}` appearing in a random female names.

Fix for issue #541


fcurella:
Thank you! ✨

Issue #544, opened 2017-06-26
sv_SE Swedish provider does not handle gender specific first names


swilcox:
This provider does not have gender specific first names.

PR coming with properly attributed source data.

swilcox:
fixed by PR #545

PR #545, opened 2017-06-26
Feature/better swedish names

swilcox:
This PR is for Issue #544 

Source data is from:

Statistiska centralbyråns - Statistics Sweden
https://www.scb.se/en/About-us/official-statistics-of-sweden/

The web site easily provides the top 100, but I used the API to pull full datasets to do the top 200 first names for each gender and the top 200 last names and generate the percentages based on the relative counts. 1999 data was used as that more closely matches the existing data that the provider had used.

Added a fairly basic test to exercise the provider and make sure that the new gender specific calls were working.


swilcox:
I'm not very proud of the code, but just as a reference for how I got the data from Statistics Sweden you can see my calls to their API and the subsequent sorting and filtering: https://github.com/swilcox/query-swedish-data

fcurella:
Looks good! Thank you! ✨

PR #546, opened 2017-07-01
Adding an implementation of `regon` for pl_PL company provider.

pdaw:
REGON is the National Business Registry in Poland.


fcurella:
Looks great! Thank you! ✨

PR #551, opened 2017-07-10
Add `postcode()` method to locales that are missing it

zafarali:
There is some non-standard behavior in some locales when generating postcodes. The main Provider implements `Provider.postcode()` but there are some locales which do not have compatibility with this and still generate the default postcode `#####`:

```
from faker import Factory
us = Factory.create('en_US')
us.postcode()
# > '19904'
ca = Factory.create('en_CA')
ca.postcode()
# > '33900'
```

Whereas it should give:
```
ca.postalcode()
# > `X2V1J7`
```

Something similar happens in `jp_JP` and `ko_KR`. This PR should fix that.


fcurella:
This raises an interesting question.

Some providers use `postcode`, some `postal_code`, other one `zipcode`.

1. Should it be consistent across providers? They kinda are the same thing, and I think it would be convenient from a pragmatic point of view. If so, which term should it be?
2. OTOH, terms like 'postalcode' or 'zipcode' are specific to a Country. For example, 'zipcode' is US-specific. Should we decide to 'localize the term' (ie: the method name)?

zafarali:
Most providers that use `postal_code` and `zip_code` add their formats in `postcode_formats` which allows it to be used by `Provider.postcode()`, and if not, they map `postcode` to `zip_code`/`postal_code`.

I think having them localized makes sense since in a particular locale they might call it something different (example: in the USA it's common to call it `zipcode()` and therefore a programmer would do `Provider.zipcode()` vs a programmer in Canada who'd do `Provider.postal_code()`)

However, all locales should implement `postcode` so that if someone wants to generate data from multiple locales, it will result in correct post codes.

fcurella:
In this case, I would implement it a little differently. I'd rather override `postcode` and alias the localized name, rather than implement the localized name and have `postcode` be an alias.

In other words:

* `address/__ini__.py` already implements `postcode`
* localized providers will override `.postcode`, and add the localized term as an alias:
    ```
    class Provider(AddressProvider):
    
        @classmethod
        def postcode(cls):
            # custom code here...

        @classmethod
        def zipcode(cls):
            return cls.postcode()
    ```

fcurella:
This has already been merged somewhere else

PR #552, opened 2017-07-10
New Provider: Automotive

zafarali:
This PR adds license plate generators for the following locales: `en_US`, `en_CA` and `en_GB`.

### DIFF
--- a/faker/providers/automotive/__init__.py
+++ b/faker/providers/automotive/__init__.py
@@ -8,6 +8,7 @@ import re
 
 class Provider(BaseProvider):
     license_formats = ()
+    
     @classmethod
     def license_plate(cls):
         temp = re.sub(r'\?',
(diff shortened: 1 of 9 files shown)


fcurella:
Maybe I would move this to a provider called 'automotive', so that we can add more methods to it in the future (eg: driving license numbers, or VINs)

faker/providers/automotive/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+
+localized = True
+
+from .. import BaseProvider
+from string import ascii_uppercase
+import re
+
+class Provider(BaseProvider):
+    license_formats = ()
+    @classmethod
```
fcurella: one last thing: could add a blank line above `@classmethod`?

fcurella:
waiting for the CI to finish

fcurella:
Thank you! ✨

Issue #421, opened 2016-11-09
name provider is a lot slower than e.g. word provider


mfussenegger:
Not sure if this is of interested, but I've noticed that the name provider is significantly slower than other providers. Which seems kind of odd:

```
from faker import Factory
import timeit

fake = Factory.create()
print(timeit.timeit('fake.word()', globals=locals()))
print(timeit.timeit('fake.name()', globals=locals()))
```

```
python faker-perf.py
2.0924576180004806
611.8756651770018
```

jchildren:
I might be wrong, but I think this is because `faker.utils.distribution.choice_provider` takes linear time due to comprehensions for the cumulative distribution function used to normalize probabilities. If probabilities were normalized in advance (or assumed to be normalised), it would be significantly faster but I can see the advantage in not assuming this for flexibility.

It might be possible to work out the cdfs for a generator when the provider is added, but this might get complicated.

jchildren:
As a point of interest, I tried out using `random.choices` from python 3.6 in place of `choice_distribution` and found that it was a few times faster with the same benchmark. Obviously this is to be expected given that `random.choices` is implemented in c, but is it worth adding the option to use `random.choices` instead if running python 3.6?

with random.choices
```
>>> print(timeit.timeit('fake.name()', globals=locals()))
111.96085176300039
```
without
```
>>> print(timeit.timeit('fake.name()', globals=locals()))
458.3832480560013
```

I'm not sure what the best way to support this option is other than using hasattr on the instance of random.

catleeball:
@jchildren an alternative to hasattr might be using sys.version_info. You could do something like:

```python
from sys import version_info

if version_info.major >= 3 and version_info.minor >= 6:
     from random import choices
    # use choices
else:
    # use choice_distribution
```

There might be a cleaner way, but I've used this solution in the past and it worked fine.

catleeball:
Started dinking with this a bit ago, tried something like this:

```python
def choice_distribution(a, p):
    assert len(a) == len(p)

    if version_info.major >= 3 and version_info.minor >= 6:
        from random import choices
        return choices(a, cum_weights=p)[0]
    else:
        cdf = list(cumsum(p))
        normal = cdf[-1]
        cdf2 = [float(i) / float(normal) for i in cdf]
        uniform_sample = random_sample()
        idx = bisect.bisect_right(cdf2, uniform_sample)
        return a[idx]
```

But it gets caught on this unit test:

```python
[...]

PR #553, opened 2017-07-11
Use random.choices when available for better performance

catleeball:


catleeball:
PR for bug #421

faker/utils/distribution.py
```diff
@@ -17,10 +18,13 @@ def cumsum(it):
...
     assert len(a) == len(p)
 
-    cdf = list(cumsum(p))
-    normal = cdf[-1]
-    cdf2 = [float(i) / float(normal) for i in cdf]
-    uniform_sample = random_sample()
-    idx = bisect.bisect_right(cdf2, uniform_sample)
-
-    return a[idx]
+    if version_info.major >= 3 and version_info.minor >= 6:
```
fcurella: is checking against `hasattr(random, 'choices') any slower?

catleeball: Good call! Looks like hasattr is quite a bit faster, at least when testing on python 3.6 on my machine.

```
$ python3 -m timeit "import random; hasattr(random, 'choices')"
1000000 loops, best of 3: 0.265 usec per loop
$ python3 -m timeit "import sys; sys.version_info.major >= 3 and sys.version_info.minor >= 6"
1000000 loops, best of 3: 0.614 usec per loop
```

I'll update the PR soon. :)

catleeball:
Okay, updated with `hasattr()`. Unit tests seem happy. Let me know if any further edits are needed. :)

jchildren:
How many times did you run the unit tests? Success seemed to be randomized. Is it worth adding new ones that are more thorough?

faker/utils/distribution.py
```diff
@@ -1,6 +1,7 @@
 # coding=utf-8
 
 import bisect
+from sys import version_info
```
fcurella: You can just remove this import, since it's not used anymore

catleeball: Oh, good catch, I'll remove that!

catleeball:
Okay, just pushed a commit removing the unused import. :)

@jchildren: Good catch! To be sure, I ran the tests 100 times with `for i in {1..100}; do python3 setup.py test &> ~/100tests.log; done` and got all passing output: https://drive.google.com/file/d/0B4p5S5l8DBDxcWxnR3JfUy1PcDQ/view?usp=sharing

fcurella:
perfect! Thank you so much for working on this, it's deeply appreciated! 🚀 ✨

PR #554, opened 2017-07-12
Adding an implementation of `local_regon` for pl_PL company provider.

pdaw:
An extension of REGON.


fcurella:
Looks good! Thank you! ✨

PR #559, opened 2017-07-19
Taiwan/Traditional character support for internet and lorem

bearnun:
:D 

謝謝! Thanks!


fcurella:
looks good! Thank you! ✨

PR #560, opened 2017-07-19
Refactored RGB color methods

catleeball:
Simplified RGB color generation a little. Previously rgb_color piggybacked off of hex_color, where it then parsed rgb ints from the hex color string, casted them as ints, then formatted them into rgb string. This just generates ints between 0-255 and formats them into a string.

Checked with the color-tests branch and it still passes the unit tests added there. :)

### DIFF
--- a/faker/providers/color/__init__.py
+++ b/faker/providers/color/__init__.py
@@ -175,18 +175,9 @@ class Provider(BaseProvider):
         return "#{0}{0}{1}{1}{2}{2}".format(*color)
 
     @classmethod
-    def rgb_color_list(cls):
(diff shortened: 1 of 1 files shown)


catleeball:
Oh! Coveralls seems unhappy. I could rebase these commits onto the color-test branch so these changes will be with their associated unit tests in that branch, but will that reopen the old color-tests PR? Apologies for being a little unfamiliar with github.

fcurella:
Yeah, don't worry about coveralls for just a .0004% decrease.

fcurella:
LGTM 👍 ✨

PR #567, opened 2017-08-07
Close #45. More descriptive error message when a formatter is not found

### LINKED ISSUE #45: Refactor Profile to be used with locale: how
I got this idea but i'm not sure it would be the simplest: the actual profile.py becomes something like "internal_profile.py", its methods are renamed "internal_simple_profile()" and "internal_profile()", and is removed from the list of standard providers. Then we will have a standard profile.py that simply calls self.generator.internal_profile(). For each locale instead, we will be able to add more logic, for example to customize field names and eventually values. 

Do you think there would be a simpler way to do it?

fcurella:


setup.py
```diff
@@ -60,8 +60,10 @@
     test_suite='tests',
     zip_safe=zip_safe,
     install_requires=[
+        "email_validator>=1.0.2",
         "python-dateutil>=2.4",
         "six",
+        "ukpostcodeparser>=1.1.1",
```
kevin-brown: With this dependency on `ukpostcodeparser`, our internal tools started throwing red flags because it's not properly licensed.

https://github.com/hamstah/ukpostcodeparser/issues/7

It's also not clear why this is in `requires` and not `test_requires`.

fcurella: you're right, it should go in `test_requires`, since it's used only for tests

setup.py
```diff
@@ -60,8 +60,10 @@
     test_suite='tests',
     zip_safe=zip_safe,
     install_requires=[
+        "email_validator>=1.0.2",
```
kevin-brown: I believe `email_validator` can also be moved to `test_requires` as well.

fcurella: thanks, fixed in db73eb54

PR #572, opened 2017-08-17
Re-formatted fr_CH address provider

Newman101:
Hi.

Looks like fr_CH address provider needed some attention. I amended the indentation issues and too long lines, improving pylint score from -3.33 / 10 to +8.75/10. 

Thank you.


fcurella:
Thank you!

FWIW I would like to add pylint to the CI. I don't think it makes much sense to spend time on indentation if the style is not enforced or checked automatically.

Newman101:
As far as I am aware, Pylint can be added to Travis build. I can check the details how to do this.

PR #575, opened 2017-08-18
Enable city suffixes.  Add section number. Add more city names.

clarmso:
The city names are taken from https://en.wikipedia.org/wiki/List_of_cities_in_Taiwan
The reference for the address format: https://en.wikipedia.org/wiki/Postal_addresses_in_Taiwan


fcurella:
Thank you! ✨

PR #583, opened 2017-08-30
Use exact dependency versions

alexmuller:
This helps downstream projects which use `pip install --require-hashes`, because of the error:

> In --require-hashes mode, all requirements must have their versions pinned with ==.
> These do not.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,10 +60,10 @@ setup(
     test_suite='tests',
     zip_safe=zip_safe,
     install_requires=[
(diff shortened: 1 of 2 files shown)


fcurella:
I didn't know about `--require-hashes`. Thank you! ✨

alexmuller:
Me neither until a couple of weeks ago when I started looking at https://github.com/pypa/warehouse. Thanks for the merge!

PR #592, opened 2017-09-04
Revert name change of faker.generator.random

### LINKED ISSUE #591: Reseed code broke in 0.8 unnecessarily
Commit ea4f189bbff1925d7a7e2d7cbc6e91e2e8a9a3f3 changed the name of the `random.Random()` object in `faker.generator` from `random` to `mod_random`, re-exposing the `random` module as the name `faker.generator.random`. This broke code like:

```py
from faker.generator import random
random.seed(1)
``` 

which is basically what my `pytest-randomly` plugin does. At first I thought this was #586 but then I saw that every run was broken. The failure is silent because `random` is still a name in `faker.generator`, it just now points at the global `random` module 😱

I suggest just doing some renaming in the module to fix this so old code continues to work.

adamchainz:
Fixes #591. This makes third party code that re-seeds the global faker library work again, after ea4f189bbff1925d7a7e2d7cbc6e91e2e8a9a3f3 changed the name and made attempts to reseed silently fail by affecting the standard random module instead.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -296,13 +296,17 @@ used to generate the values:
     fake.random
     fake.random.getstate()
 
+By default all generators share the same instance of ``random.Random``, which
+can be accessed with ``from faker.generator import random``. Using this may
+be useful for plugins that want to affect all faker instances.
+
 Seeding the Generator
 ---------------------
 
 When using Faker for unit testing, you will often want to generate the same
 data set. For convenience, the generator also provide a ``seed()`` method, which
-seeds the random number generator. Calling the same script twice with the same
-seed produces the same results.
+seeds the shared random number generator. Calling the same methods with the
+same version of faker and seed produces the same results.
 
 .. code:: python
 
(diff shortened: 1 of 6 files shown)


fcurella:
Can you document the use of Faker.generator.faker in the Readme and in the docs?

adamchainz:
@fcurella done, also I noticed `seed_instance()` wasn't doc'd so did that. I don't get the docs layout, it seems the readme is simply duplicated at `docs/index.rst`? Could be better structured as an RST include.

fcurella:
@adamchainz Yeah, `docs/index.rst` should be changed to use an RST include. Do you mind adding that change to this PR?

adamchainz:
I think it's best to tackle it later, I'm not sure they're 100% equivalent.

fcurella:
looks good, thank you so much! ❤️

adamchainz:
🎉

Issue #601, opened 2017-09-15
Lorempixel images requested over http


swizzard:
Image urls from lorempixel are `http`, causing unnecessary security warnings when served from `https` pages.

![lorempixel](https://user-images.githubusercontent.com/1909918/30486431-23cb42dc-99f6-11e7-8307-665bdd06b207.png)

Earliest known version: 0.7.18

PR #602, opened 2017-09-15
use https for lorempixel

swizzard:
fixes https://github.com/joke2k/faker/issues/601


fcurella:
looks good, thank you! ✨

Issue #563, opened 2017-07-31
date_time issues - Windows only?


kungfu71186:
Could possibly be windows only. Need to debug further, but it seems like this is the windows bug where if a timestamp is generated within the first day after 1970, then there is an error. This happens anywhere when using timestamp. Will investigate further, but opening an issue to not forget. 

Hard to reproduce, but you can do datetime.fromtimestamp(0) and get an error, anything above 86400, is fine datetime.fromtimestamp(86400)

Also getting another error on:
timestamp = random.randint(start_date, end_date)
ValueError: empty range for randrange(). This is due to start_date being greater than end_date. Happens very rarely though.

an-oreo:
Ran into same issue for the datetime.fromtimestamp() function on windows. after poking about I think its due to the Cpython implemetation using  what looks to be a windows specific function call for the C localtime() function. 
https://github.com/python/cpython/blob/c90e96015085784df86632b26059b19c80cbfc97/Python/pytime.c#L793

MS documentation
https://msdn.microsoft.com/en-us/library/a442x3ye.aspx

linux testing works fine, and putting in values that caused the error ( [0, 86400) ) worked on my windows box using the standard C localtime function.

The crash doesnt seem to happen if I specify a timezone. In my case central time.
eg:
```python
# from repl 
from datetime import datetime, timezone, timedelta

# fails for all values [0, 86400)
for i in range(86400):
    try: 
        datetime.fromtimestamp(i)
    except OSError:
        pass

central = timezone(-timedelta(hours=6))

# works 
for i in range(86400):
    try: 
        datetime.fromtimestamp(i, central)
    except OSError:
        pass
```

In terms of a possible fix, maybe changing the call from `datetime.fromtimestamp()` to `datetime.utcfromtimestamp()`? That seems to work for me.
The alternative would be to provide an arguement do the `fake.date()` function, but I havent looked into how that'd translate into implemetation details. 

```python
# from repl 
from datetime import datetime, timezone, timedelta

# fails for all values [0, 86400)
for i in range(86400):
    try: 
        datetime.fromtimestamp(i)
    except OSError:
        pass

# works 
for i in range(86400):
    try: 
        datetime.utcfromtimestamp(i)
    except OSError:
        pass
```

I will probably poke about more later to test my theory, but thought I'd throw my two cents.

fcurella:
Closed By #604

PR #604, opened 2017-09-19
Fixes issue #563

kungfu71186:
Fix Issue #563. The performance difference between using fromtimestamp and using datetime with timedelta is negligible.


fcurella:
LGTM! Thank you!

PR #606, opened 2017-09-20
Fixes #566

### LINKED ISSUE #566: Cannot seed datetime generator
The faker lib allows you to seed Python's `random` instance, and the expected result is that all faker calls will have deterministic results every time the code is run. But the datetime generator relies on the `unix_time()` method, which drifts based on the current time:
```
@classmethod
def unix_time(cls):
    """
    Get a timestamp between January 1, 1970 and now
    :example 1061306726
    """
    return random.randint(0, int(time()))
```
The usage of time() as the end date means that datetime results change over time, even if random has been seeded.

I suggest making these methods accept an optional `end_datetime` (maybe `start_datetime` as well) parameter that would let the caller set the range of acceptable datetimes. That would allow users seeding random to get deterministic results.

kungfu71186:
Fixes #566, added end_datetime to all methods to provide consistent seeding when needed.

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -246,21 +246,23 @@ class Provider(BaseProvider):
 
     regex = re.compile(timedelta_pattern)
 
-    def unix_time(self):
+    def unix_time(self, end_datetime = None):
         """
         Get a timestamp between January 1, 1970 and now
         :example 1061306726
         """
-        return self.generator.random.randint(0, int(time()))
+        end_datetime = self._parse_end_datetime(end_datetime)
+        return self.generator.random.randint(0, end_datetime)
 
-    def time_delta(self):
+    def time_delta(self, end_datetime = None):
         """
         Get a timedelta object
         """
-        ts = self.generator.random.randint(0, int(time()))
+        end_datetime = self._parse_end_datetime(end_datetime)
+        ts = self.generator.random.randint(0, end_datetime)
         return timedelta(seconds=ts)
 
-    def date_time(self, tzinfo=None):
+    def date_time(self, tzinfo=None, end_datetime = None):
         """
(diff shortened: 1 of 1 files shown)


faker/providers/date_time/__init__.py
```diff
@@ -289,46 +292,53 @@ def date_time_ad(self, tzinfo=None):
...
+    
+    @classmethod
+    def _parse_end_datetime(self, value):
+        if value is None:
+            return int(time())
+
+        return self._parse_date_time(value)
 
     @classmethod
-    def _parse_date_string(cls, value):
```
fcurella: could you leave the first argument of `classmethod`s as `cls`?

faker/providers/date_time/__init__.py
```diff
@@ -246,21 +246,23 @@ class Provider(BaseProvider):
 
     regex = re.compile(timedelta_pattern)
 
-    def unix_time(self):
+    def unix_time(self, end_datetime = None):
```
fcurella: no space around the `=` in arguments, as per PEP8

faker/providers/date_time/__init__.py
```diff
@@ -289,46 +292,53 @@ def date_time_ad(self, tzinfo=None):
...
+    def time_object(self, end_datetime = None):
         """
         Get a time object
         :example datetime.time(15, 56, 56, 772876)
         """
-        return self.date_time().time()
+        return self.date_time(end_datetime=end_datetime).time()
+    
+    @classmethod
+    def _parse_end_datetime(self, value):
```
fcurella: `@classmethod`s shuold use `cls` instead of `self`

kungfu71186: Yeah, I thought I had changed it back after I realized this. Ill take care of it

fcurella:
Thank you! ✨

PR #609, opened 2017-09-25
Add ASCII emails, add random uuid4 emails

barseghyanartur:
A couple of additions to the existing emails:

- Added ASCII emails (100% no Unicode character occurrence).
- Added UUID4 emails (to ensure uniqueness, when needed).

The following methods added:

- ascii_email
- ascii_safe_email
- ascii_free_email
- ascii_company_email


faker/providers/internet/__init__.py
```diff
@@ -88,6 +89,39 @@ def company_email(self):
     def free_email_domain(self):
         return self.random_element(self.free_email_domains)
 
+    def uuid4_email(self):
```
fcurella: I don't really see the use case for uuid4 emails. The goal of Faker is provide plausible values, and I don't think there are many people using uuid4 for their email.

tests/providers/test_internet.py
```diff
@@ -87,3 +87,20 @@ def test_free_email_domain(self):
...
+
+    def setUp(self):
+        self.factory = Factory.create('nl_NL')
+        self.provider = self.factory.provider('faker.providers.internet')
+
+    def test_uuid4_email(self):
+        email = self.factory.uuid4_email()
+        validate_email(email, check_deliverability=False)
+
+    def test_uuid4_free_email(self):
```
fcurella: could you write tests for the `ascii_*` methods too?

faker/providers/internet/__init__.py
```diff
@@ -68,7 +69,7 @@ def _to_ascii(self, string):
         for search, replace in self.replacements:
             string = string.replace(search, replace)
 
-        return string
+        return string.encode('ascii', 'ignore').decode('utf8')
```
fcurella: 👍

tests/providers/test_internet.py
barseghyanartur: @fcurella: Sure.

faker/providers/internet/__init__.py
barseghyanartur: @fcurella:

Well, I've added that, because I needed it. I use faker as a generator of test data. My last use case was to generate 100 POST requests to an REST API. Emails had to be unique. I have solved it by now by registering a custom provider, but purely for testing, absolutely unique emails are handy.

tests/providers/test_internet.py
barseghyanartur: @fcurella:

ascii tests added.

faker/providers/internet/__init__.py
fcurella: I understand your use case, but I don't think it's common enough to be included as a convenience.
For people who needs an uuid4 email, It's still possible to generate one by simply calling `faker.uuid4() + '@' + faker.domain_name()`

faker/providers/internet/__init__.py
barseghyanartur: @fcurella:

I removed the uuid4 emails (unfortunately, our comments were gone with the change too).

fcurella:
this change will generate invalid emails for arabic locales, since it will drop all characters in a name.

We need to decide how to 'asciify' arabic names before this gets merged.

faker/providers/internet/__init__.py
fcurella: I would add a `drop_ascii = True` property on this class, check for its value in the method, and set the property to `False` for the providers in `/providers/internet/ar_*`

barseghyanartur: @fcurella:

It could be done the same way it's done for de_DE. So, something like this, but then for Arabic:

```python

# coding=utf-8

from __future__ import unicode_literals
from .. import Provider as InternetProvider

class Provider(InternetProvider):

    free_email_domains = (
        'aol.de', 'gmail.com', 'gmx.de', 'googlemail.com', 'hotmail.de',
        'web.de', 'yahoo.de',
    )
    tlds = ('com', 'com', 'com', 'net', 'org', 'de', 'de', 'de', )

    replacements = (
        ('ä', 'ae'), ('Ä', 'Ae'),
        ('ö', 'oe'), ('Ö', 'Oe'),
        ('ü', 'ue'), ('Ü', 'Ue'),
        ('é', 'e'), ('É', 'E'),
        ('à', 'a'), ('À', 'A'),
        ('ß', 'ss'),
    )
```

fcurella: that would be the best way, but I don't know how to transliterate arabic (assuming it's even possible)

barseghyanartur: @fcurella:

I've just done that. See the updated tests and the ``ar_AA module``.

barseghyanartur: @fcurella:

Question. Since you know faker better, we always transliterate **to** ASCII and never **from** ASCII, right? If so, there is a more elegant solution, using a dependency (``unidecode``), which does it pretty well for (almost) all languages.

barseghyanartur: Just a snippet of code:

```python

import unidecode

unidecode.unidecode('اصيل')
unidecode.unidecode('Ժամանակով Կատուն ճոն էր')
unidecode.unidecode('У попа была собака')
```

fcurella: the problem with using `unidecode` is that `unidecode` is not locale aware. For example, it converts `ü` to `u`, when in German it should be replaced with `ue`.

I think we can combine the two: use the replacement list if it's defined, or fallback to unidecode if it's not.

fcurella:
I've reworked it a little bit to use both `replacements` and unidecode.

Thank you for the help :)

PR #611, opened 2017-09-27
Add Provider id_ID

ceceprawiro:


faker/providers/person/id_ID/__init__.py
```diff
@@ -0,0 +1,193 @@
...
+        '{{first_name_female}} {{last_name}}',
+        '{{prefix_male}} {{first_name_male}} {{last_name}}',
+        '{{prefix_female}} {{first_name_female}} {{last_name}}',
+        '{{first_name_male}} {{last_name}}, {{suffix}}',
+        '{{first_name_female}} {{last_name}}, {{suffix}}',
+        '{{prefix_male}} {{first_name_male}} {{last_name}}, {{suffix}}',
+        '{{prefix_female}} {{first_name_female}} {{last_name}}, {{suffix}}',
+    )
+
+    # From https://github.com/fzaninotto/faker/blob/master/src/Faker/Provider/id_ID/Person.php
```
fcurella: Could you cite the original sources (ie: the sources that are in PHP Faker)? I think it will make it easier to update the lists in the future

ceceprawiro: You mean something like these?

```
# From https://github.com/fzaninotto/faker/blob/master/src/Faker/Provider/id_ID/Person.php
# (http://www.nama.web.id/search.php?gender=male&origin=Indonesia+-+Jawa&letter=&submit=Search)
```

I already mentioned some source links. The rest are my own work.

fcurella: something like that would work. Let's use this format:
```
# From http://www.nama.web.id/search.php?gender=male&origin=Indonesia+-+Jawa&letter=&submit=Search
# via https://github.com/fzaninotto/faker/blob/master/src/Faker/Provider/id_ID/Person.php
```

We only require sources for the `person` provider. Citing the source for other providers is appreciated, but not required.

fcurella:
Thank you! ✨

Issue #613, opened 2017-09-29
date_time.time_series() output does not include bounds as described 


bijanvakili:
The `date_time.time_series` provider's comments indicate [here](https://github.com/joke2k/faker/blob/master/faker/providers/date_time/__init__.py#L706) that:

```
The data points will start at ``start_date`` ...
```

However, the [generator's loop](https://github.com/joke2k/faker/blob/master/faker/providers/date_time/__init__.py#L727) skips the start interval because it increments first.

```python
>>> import pprint
>>> import faker
>>> import datetime

>>> now = datetime.datetime.now()
>>> start = now - datetime.timedelta(days=30)
>>> start
datetime.datetime(2017, 8, 30, 10, 31, 26, 253120)
>>> end = now
>>> end
datetime.datetime(2017, 9, 29, 10, 31, 26, 253120)

>>> series = f.time_series(start_date=start, end_date=end)
>>> pprint.pprint(list(series))
[(datetime.datetime(2017, 8, 31, 10, 31, 26), 13495.847561694067),  # not start!!
 (datetime.datetime(2017, 9, 1, 10, 31, 26), 16816.17939316429),
 ...
```

@fcurella

fcurella:
closed by #614

PR #614, opened 2017-09-29
Fix date_time.time_series() to ensure start and end bounds are inclusive

bijanvakili:
* Fix for issue #613
* Includes update to regression tests

@fcurella

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -726,9 +726,12 @@ class Provider(BaseProvider):
 
         datapoint = start_date
         while datapoint < end_date:
-            datapoint += precision
             dt = timestamp_to_datetime(datapoint, tzinfo)
             yield (dt, distrib(dt))
+            datapoint += precision
+
+        dt = timestamp_to_datetime(end_date, tzinfo)
+        yield (dt, distrib(dt))
 
     def am_pm(self):
         return self.date('%p')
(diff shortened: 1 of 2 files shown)


fcurella:
I don't think `end_time` should *always be included. if the last time interval is shorter than `precision`, then the last datapoint is outside of the range.

tests/providers/test_date_time.py
```diff
@@ -354,6 +354,13 @@ def test_time_series(self):
         self.assertTrue(len(series), 7)
         self.assertTrue(series[1][0] - series[0][0], timedelta(days=1))
 
+        # avoid microseconds as provider's internal parsing uses POSIX timestamps which only have second granularity
+        end = datetime.now(timezone.utc).replace(microsecond=0)
```
fcurella: use the provided `utc` instance (L37) instead of importing `timezone`

faker/providers/date_time/__init__.py
```diff
@@ -726,9 +726,9 @@ def time_series(self, start_date='-30d', end_date='now', precision=None, distrib
 
         datapoint = start_date
         while datapoint < end_date:
-            datapoint += precision
             dt = timestamp_to_datetime(datapoint, tzinfo)
             yield (dt, distrib(dt))
+            datapoint += precision
```
fcurella: can we move this one line up, just above the `yield`? mostly a stylistic preference

fcurella:
Thank you!

PR #616, opened 2017-10-02
Use a proper international format for Ukrainian phone numbers

illia-v:
The third digit should always be '0'.


fcurella:
Thank you! ✨

PR #619, opened 2017-10-10
Removed potentially offensive phrases

astawiarski:
Removing 'Sucker' from the list of surnames to avoid any concerns when results are exposed to public.

This is exactly same situation as with PR #379 and it's main intention is to make it safe for developers to use the results of this library in a publicly facing context.

I understand that this project could benefit from having some names that expose [Scunthorpe Problem](https://en.wikipedia.org/wiki/Scunthorpe_problem), but this may as well be a purpose for a completely separate library.


fcurella:
Thank you! ✨

PR #626, opened 2017-10-15
Close #622. Add en_GB cellphone formats

### LINKED ISSUE #622: Uk mobile number
It seems like the uk mobile number is not in the right format 
it's completely not valid
some examples of them: 
+44(0)9128 405119
(01414) 35336
01231052134

jacobbudin:
Fixes #622.

Primary source:
https://en.wikipedia.org/wiki/Telephone_numbers_in_the_United_Kingdom#Mobile_telephones

***

Example:

```python
import faker

from faker import Faker
fake = Faker('en_GB')
for _ in range(10):
    print(fake.cellphone_number())
```

```
07352099948
+44 7466113027
+44 7894363975
07353912887
07151 252843
07925 440819
07179 121894
07603681069
07349589776
07849352443
```

### DIFF
--- a/faker/providers/phone_number/en_GB/__init__.py
+++ b/faker/providers/phone_number/en_GB/__init__.py
@@ -3,6 +3,15 @@ from .. import Provider as PhoneNumberProvider
 
 
 class Provider(PhoneNumberProvider):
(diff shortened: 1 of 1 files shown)


fcurella:
LGTM. Thank you! ✨

PR #627, opened 2017-10-16
Fix README link to Ruby faker

ryankask:
The RubyForge site is no longer active so I've used the project's GitHub URL


fcurella:
nice catch, thank you!

PR #631, opened 2017-10-17
Fix ResourceWarning in setup.py

jdufresne:
Always close file when finished by using a context manager. Fixes `ResourceWarning` when running Python with warnings enabled.


fcurella:
good catch! Thank you so much!

PR #635, opened 2017-10-19
Add title faker

kleingeist:
Faker is missing the equivalent to real world titles. A title consists of a
number of words. The first letter of the first word is uppercase and there is
no punctuation.

For example: 'Sapiente sunt omnis'

In applications with a database the title is usually restricted to
certain number of chars. Thus the title faker allows to restrict the
title char number.

(This PR resulted from an issue with titles and email subjects which do 
not support newlines. For reference see liqd/adhocracy4#200)

### DIFF
--- a/faker/providers/lorem/__init__.py
+++ b/faker/providers/lorem/__init__.py
@@ -182,3 +182,32 @@ class Provider(BaseProvider):
                 text.pop()
 
         return "".join(text)
+
+
+    def title(self, max_nb_chars=50, ext_word_list=None):
+        """
+        Generate a title string.
+        Returns a title with at most ``max_nb_chars`` made of words.
+        :example 'Sapiente sunt omnis'
+
+        Keyword arguments:
+        :param max_nb_chars: Maximum number of characters the title should contain (minimum 5)
+        :param ext_word_list: a list of words you would like to have instead of 'Lorem ipsum'.
+
+        :rtype str
+        """
+        title = []
+        if max_nb_chars < 5:
+            raise ValueError('title() can only generate text of at least 5 characters')
+
+
+        while not title:
+            size = 0
+            while size < max_nb_chars:
+                word = (self.word_connector if size else '') + self.word(ext_word_list)
+                title.append(word)
+                size += len(word)
+            title.pop()
+
+        title[0] = title[0].title()
+        return ''.join(title)


kungfu71186:
I actually just went through a very very similar issue. 

Issues
1. Naming the function title. Python already has a function title for strings.

2. I believe there are some other providers that may be acceptable to use.

One way I overcame this was to generate words and then set a char length with slices. For example. Faker.sentence(50)[:200]. I'll have to look how I actually did it, but it was similar to that.

Now to capitalize a first letter of first word, I'd recommend just doing sentence[0].upper()

Now I do agree, there is no way to limit characters and words at the same time right now. This may be something of worth.

fcurella:
I think the best approach is to just add a `max_length` argument to `.sentence()` and use that.

fcurella:
alternatively, we could change `.sentence()` to make the punctuation optional, and leave the 'trimming' up to the user.

kungfu71186:
@fcurella I think a combination of both would be good. This would at least optimize it a bit so you won't need to generate more data than you need. I don't think there would be a performance impact by implementing this either.

kungfu71186:
@fcurella @kleingeist 

This is kinda what I'm thinking:

```
    def sentence(self, nb_words=6, variable_nb_words=True, ext_word_list=None, max_nb_chars=9999, sentence_punctuation=sentence_punctuation):
        """
        Generate a random sentence
        :example 'Lorem ipsum dolor sit amet.'
        :param nb_words: around how many words the sentence should contain
        :param variable_nb_words: set to false if you want exactly ``nb``
            words returned, otherwise the result may include a number of words
            of ``nb`` +/-40% (with a minimum of 1)
        :param ext_word_list: a list of words you would like to have instead of
            'Lorem ipsum'.
        :param max_nb_chars: Maximum number of characters the sentence should contain
        :param sentence_punctuation: End of sentence punctuation character(s)
        :rtype: str
        """
        if nb_words <= 0 or max_nb_chars <= 0:
            return ''

        if len(sentence_punctuation) >= max_nb_chars:
            return sentence_punctuation[:max_nb_chars]

        if variable_nb_words:
            nb_words = self.randomize_nb_elements(nb_words, min=1)

        if sentence_punctuation is None:
            sentence_punctuation = ''
        
        max_nb_chars -= len(sentence_punctuation)
        max_length_size = max_nb_chars
        words = []
        while max_length_size >= 0 and nb_words >= 0:
            word = (self.word_connector if words else '') + self.word(ext_word_list=ext_word_list)
            words.append(word)
            max_length_size -= len(word)
            nb_words -= 1

        if len(words) == 1:
            words[0] = words[0][:max_nb_chars]
        else:
            words.pop()
        

        words[0] = words[0].title()

        return "".join(words) + sentence_punctuation
```

This should keep the old functionality, but adds in the new length. 
Rules that I could think of:
- Punctuation is part of length. 
- You can have a size of max_nb_chars.
- If punctuation is longer than max_nb_chars, then just return the punctuation with that size.
- If you only get 1 word, then trim that one word to max_nb_chars

fcurella:
I would leave `max_nb_chars` out of the equation, and let that up to the user to slice the result. Python already has a convenient way to slice strings, let's use that.

kungfu71186:
@fcurella only reason I was thinking this way, is really due to being more efficient. I'll run some tests and see though if it would make any difference

kleingeist:
thank you for the good feedback. 

@kungfu71186 we have decided to use the `sentence` faker, too. and for the last week we did not generate strings that are to long for our database restrictions.
we are not able to trim the generated string by slicing, as we are using [faker through factory_boy](http://factoryboy.readthedocs.io/en/latest/#realistic-random-values). we may only define the faker method and the arguments to pass, as the actual string is generated later when the factory is called.

but i see the increased complexity when using both words and char restrictions. thats why i was aiming for a separate "title" faker. although i have to admit the name clash with pythons string "title" is bad.

fcurella:
I haven't considered the `factoryboy` scenario. I _think_ you should be able to use some like `factory.LazyAttribute(lambda a: Faker.sentence()[:length])`, but I also think this use case is common enough that we should make it easier.

What if we add `mode='auto'` to the `text` provider. Valid values would be:
* `'auto'`: current behavior. return words, sentences, or paragraphs depending on `max_nb_chars`
* `'words'`, `'sentences'`, `'paragraphs`: force the results to be words, sentences, or paragraphs, regardless of `max_nb_chars`

shacker:
I need this too! Before this is merged, a  bit of feedback: 

`title` is ambiguous. People and cars and other things have titles that don't look anything like news headlines. Since this faker will mostly be used for news sites and blogs (I'd imagine), how about calling this `headline` instead? Nice and clear.

I'd also recommend that this faker generate output in Title Case by default, with most words Initial Capped: "The Quick Brown Fox Jumps over the Lazy Dog", or provide Title Case as an option. This can be done with the addition of the [titlecase](https://github.com/ppannuto/python-titlecase) dependency.

stevelacey:
I am currently using `faker.sentence().title()[:-1]` it'd be nice to have `faker.title()` – anything involving kwargs for this I wouldn't use

xlotlu:
> I'd also recommend that this faker generate output in Title Case by default, with most words Initial Capped: "The Quick Brown Fox Jumps over the Lazy Dog", or provide Title Case as an option. This can be done with the addition of the [titlecase](https://github.com/ppannuto/python-titlecase) dependency.

Also note that the meaning of titlecase is language-dependent.

OJFord:
It's an old comment, but since it's frustrated me with other providers, I just wanted to point out that you can't do:
```python
factory.LazyAttribute(lambda a: Faker.sentence()[:length])
```
(NB it's `factory.Faker('sentence')`, but that's not the issue)

you need `factory.Faker('sentence').generate()`, because factoryboy won't evaluate factories inside that lambda.

a cleaner workaround is sometimes to do something like:
```python
class Params:
    _title = factory.Faker('sentence')

title = factory.LazyAttribute(lambda a: a._title[:length])
```

particularly as these get longer and nastier and you have more of them (possibly with common parts).

fcurella:
We can add `max_nb_chars` to the `sentence` provider is it's needed for factoryboy. Anybody, feel free to submit a PR :)

fcurella:
Closing for inactivity

PR #637, opened 2017-10-23
Added the ability to create college majors

herdingbats:
I needed my fake data to include field of study, so I made a module for it. It's a bit idiosyncratic (the list of fields comes from the University of Chicago) but it's a start.


fcurella:
I think this could go into the `profile` provider, rather than being its own

faker/providers/major/__init.py__
```diff
@@ -0,0 +1,72 @@
...
+		"Social Sciences",
+		"Sociology",
+		"South Asian Languages and Civilizations",
+		"Statistics",
+		"Theater and Performance Studies",
+		"Tutorial Studies",
+		"Visual Art"
+	)
+
+def major(self):
```
fcurella: I think you missed an indentation level here

herdingbats: Will fix and re-submit.

herdingbats: In the larger scheme of things, would it make sense to have a few different functions for education? One for level (from "some HS" up through "PhD") and then one for field (from "broad field" to "detailed field" to "major"--I'm pulling these categories from p. 2 of this Census report: https://www.census.gov/prod/2012pubs/acs-18.pdf.

fcurella:
closing for inactivity

PR #641, opened 2017-11-02
Fix infinite loop random sample unique

153957:
This is fixed by using `collections.Counter` on the elements (if not already a dict). Duplicate values are then removed so the length of the elements should be the number of unique elements. Using `collections.Counter` counts the occurrence of elements to maintain the higher probability for those values to be in the final sample.

Before this change the following would result in an infinite loop:
```python
from faker import Generator
from faker.providers import BaseProvider
provider = BaseProvider(Generator())
provider.random_sample_unique('aabcd', 5)
```
Because the sample will never be 5 values long because there are only 4 unique values (`'abcd'`) in the provided elements.

### DIFF
--- a/faker/providers/__init__.py
+++ b/faker/providers/__init__.py
@@ -1,5 +1,6 @@
 # coding=utf-8
 
+from collections import Counter
 import re
 import string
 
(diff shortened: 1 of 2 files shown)


153957:
I also made performance improvements by removing values from the elements which have already added to the sample, because they can not be chosen again.

Before:
```python
%timeit provider.random_sample_unique({'a':1000000, 'e':1}, 2)
# The slowest run took 9.92 times longer than the fastest. This could mean that an intermediate result is being cached.
# 2.72 s ± 1.83 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
After:
```python
%timeit provider.random_sample_unique({'a':1000000, 'e':1}, 2)
# 17.8 µs ± 479 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

fcurella:
Awesome! Thank you very much! 🚀

PR #645, opened 2017-11-05
Add he_IL locales

bjesus:
This adds Hebrew (Israeli) locale to Faker.

Supported providers are: address, person, ssn, phone_number and lorem.

Thank you for the great library!


fcurella:
Thank you! ✨

PR #646, opened 2017-11-06
Close #644. Fix inheritance of en_US bs company provider.

### LINKED ISSUE #644: fake.bs is missing for locale ja_JP
### Steps to reproduce

from faker import Faker
Faker('ja_JP').bs()

### Expected behavior

It should return something

### Actual behavior

It throws an exception.

I think that this ticket is less about this specific case, but more about whether the tests could find all the missing methods and then generate a report. This way I could restrict my usage of faker to those generators which exist in most locales. My use case is that I run all my tests in multiple locales to check that no issues creep into the code.

fcurella:
Fixes #644 

### What does this changes

The  methods in the `company/en_US` provider have been moved to the base `CompanyProvider`

### What was wrong

The fallback mechanism to `en_US` was broken because not all methods in `company/en_US` where defined in `CompanyProvider`

### How this fixes it

All methods that were in `company/en_US` are now available to all locales of `company`


pjsg:
Thank you for this!

Issue #643, opened 2017-11-03
IBAN/BIC  Provider needed


cdaller:
Has someone written an IBAN and BIC provider?
Any other people interested?

kungfu71186:
Not really interested in the data, but I can see it being beneficial. 

https://en.wikipedia.org/wiki/International_Bank_Account_Number
The IBAN consists of up to 34 alphanumeric characters, as follows:
country code using ISO 3166-1 alpha-2 – two letters,
check digits – two digits, and
Basic Bank Account Number (BBAN) – up to 30 alphanumeric characters that are country-specific.[1]

The preferred algorithm is:[8]
Check that the total IBAN length is correct as per the country. If not, the IBAN is invalid.
Replace the two check digits by 00 (e.g., GB00 for the UK).
Move the four initial characters to the end of the string.
Replace the letters in the string with digits, expanding the string as necessary, such that A or a = 10, B or b = 11, and Z or z = 35. Each alphabetic character is therefore replaced by 2 digits
Convert the string to an integer (i.e. ignore leading zeroes).
Calculate mod-97 of the new number, which results in the remainder.
Subtract the remainder from 98, and use the result for the two check digits. If the result is a single digit number, pad it with a leading 0 to make a two-digit number.

Basically IBAN is the same for all countries. BICs are different obviously and are defined per country. Should be fairly straightforward.

fcurella:
That's a good idea, but I don't have enough time to dedicate to the implementation. Feel free to submit a PR and I'll review it :)

cdaller:
PR is on the way :-)

mdxs:
Guess this issue is closed, as the PR has been merged.

Menda:
I just wanted to comment that would be good to have BIC as well.

PR #647, opened 2017-11-07
added iban provider

cdaller:
### What does this changes

see #643 
added a provider for fake iban/bban data.
the iban/bban codes are valid in term of checksum, but they do not hold valid bank identifiers or account numbers.
IBAN format was taken from wikipedia https://en.wikipedia.org/wiki/International_Bank_Account_Number

country support is far from complete, but can be extended easily

### DIFF
--- a/faker/providers/bank/__init__.py
+++ b/faker/providers/bank/__init__.py
@@ -1,0 +1,36 @@
+# coding=utf-8
+
+from .. import BaseProvider
+
+import string
+from string import ascii_uppercase
+import re
+
+localized = True
+
+class Provider(BaseProvider):
+    
+    ALPHA = {c: str(ord(c) % 55) for c in string.ascii_uppercase}
+
+     # see https://en.wikipedia.org/wiki/International_Bank_Account_Number
+    bban_format = '????#############'
+    country_code = 'GB'
+
+    def bank_country(self):
+        return self.country_code
+
+    def bban(self):
+        temp = re.sub(r'\?',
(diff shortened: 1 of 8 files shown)


faker/providers/bank/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+
+from .. import BaseProvider
+
+import string
+from string import ascii_uppercase
+import re
+
+localized = True
+
+class Provider(BaseProvider):
```
fcurella: add `default_locale = 'en_GB'` instead of creating the `en_US` dummy provider

faker/providers/bank/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    Provider for IBAN/BBAN: it generates valid (valid length, valid checksum)
+    IBAN/BBANs for the given country. But the ids of the banks are random and
+    not valid banks! Same for account numbers.
+    """
+    
+    ALPHA = {c: str(ord(c) % 55) for c in string.ascii_uppercase}
+
+     # see https://en.wikipedia.org/wiki/International_Bank_Account_Number
```
fcurella: PEP8: remove the extra leading space before this comment

faker/providers/bank/__init__.py
fcurella: PEP8: two blank lines above classes

cdaller:
all requested changed done

fcurella:
Thank you! ✨

Issue #332, opened 2016-02-29
AttributeError: 'module' object has no attribute 'Provider'   on Faker()


aldarund:
When i create a Faker class i got following error:

`````` ======================================================================
ERROR: Failure: AttributeError ('module' object has no attribute 'Provider')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/usr/local/lib/python2.7/dist-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/usr/local/lib/python2.7/dist-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/rentapplication/django-rentapplication/rentapplication/tests/test_api/test_sessions.py", line 10, in <module>
    from rentapplication.tests.helpers import ReferrerFactory, LandlordFactory
  File "/home/rentapplication/django-rentapplication/rentapplication/tests/helpers.py", line 35, in <module>
    f = Faker()
  File "/usr/local/lib/python2.7/dist-packages/faker/factory.py", line 38, in create
    prov_cls, lang_found = cls._get_provider_class(prov_name, locale)
  File "/usr/local/lib/python2.7/dist-packages/faker/factory.py", line 49, in _get_provider_class
    provider_class = cls._find_provider_class(provider, locale)
  File "/usr/local/lib/python2.7/dist-packages/faker/factory.py", line 87, in _find_provider_class
    return provider_module.Provider
AttributeError: 'module' object has no attribute 'Provider'```
``````

**Comment by EnTeQuAk** [comment]
This happened most recently in the 0.5.4 release and only in the wheel package. Somehow the locales get moved into the `providers` folder:

```
/Downloads/ff/faker/providers> ls
address/    color.py        currency/     el_GR/  fa_IR/   __init__.py  ko_KR/    miscelleneous.py  phone_number/    python/    user_agent/
address.py  company/        currency.py   en_CA/  fi_FI/   internet/    lorem/    misc.py           phone_number.py  python.py  user_agent.py
barcode/    company.py      date_time/    en_GB/  file/    internet.py  lorem.py  nl_NL/            pl_PL/           ru_RU/     zh_CN/
[...]

PR #649, opened 2017-11-10
fix work with aws lambda

dobrovolsky:
please see issue #332


fcurella:
Thank you so much for figuring this out! ✨

PR #650, opened 2017-11-11
Added animals by common en name

tytoons:
Needed a list of favorite animals, added list of common English animals.


fcurella:
Thanks for your contribution!

I'm not convinced this is a common enough scenario to be included in `Faker` by default. But you could package separately and I'll be happy to include it in the [Community Provider list](http://faker.readthedocs.io/en/master/communityproviders.html).

PR #659, opened 2017-11-24
fi_FI address and job lists updated based on Finnish open data 

mikkhola:
### What does this changes

- fi_FI job list updated based on a Finnish open data service.
- City list updated from the open data, contains all the cities of Finland.
- Original fruit based address prefixes replaced by the real prefixes from the city of Helsinki. Now prefixes contains spaces and special characters for matching better with the real addresses.
- Some basic unit tests added


fcurella:
Thank you!

PR #662, opened 2017-12-04
fix path issue with cx_freeze

AndreyGrafMsk27:
### What does this changes

Brief summary of the changes

### What was wrong

### How this fixes it

Fixes #...


fcurella:
Thank you for reminding me @AndreyGrafMsk27 :)

PR #663, opened 2017-12-13
fix incorrect value for that seed value

lmr2391:
### What does this changes
The name on the front page of the documentation for the section with the seeds

### What was wrong
The name for the given seed. Probably because the underlying data changed since that section was written/updated

### How this fixes it

Trivial


fcurella:
Thank you!

PR #668, opened 2018-01-04
Remove extras_require for unsupported Python versions

jdufresne:
Python 3.0 and 3.2 are not tested, nor supported. Drop the reference in `extras_require`.


fcurella:
thank you!

PR #670, opened 2018-01-04
Remove support for EOL Python 3.3

jdufresne:
Python 3.3 is EOL. It is no longer receiving bug fixes, including for security issues. It has been EOL since 2017-09-29.

https://devguide.python.org/#status-of-python-branches

Reduces test load.


fcurella:
Thank you!

PR #671, opened 2018-01-05
Close #667. Allow past dates within a second

### LINKED ISSUE #667: past_datetime() crashes when start_date is within one second of now
I ran into a problem while generating a bunch of fake dates where I was calling `fake.past_datetime()` with the start date using the result of a previous `fake.past_datetime()` call (simulating a bunch of files being created as part of a collection). The first call could randomly return a value which is within one second of the current time, causing the second call to trigger an exception from `random.randint()` because the `a` and `b` values were the same.

This was pretty easy to solve simply by using `fake.date_time_between(start_date=previous_created_date)` but it might be useful if `past_datetime()` was safe to call like this.

fcurella:
### What does this changes

We test for the difference between `start_date` and `end_date` in `date_time_between`. If it's `<= 1`,  we generate a random number between `0.0` and `1.0` by using `random.random()` instead of using `random.randint()`

### What was wrong

When the difference between `start_date` and `end_date` in `date_time_between` is <= 1 second, the call to `random.randomint` fails with a `ValueError`

### How this fixes it

`random.random()` won't raise `ValueError` when the difference is `<= 1`

Fixes #667

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -1342,7 +1342,10 @@ class Provider(BaseProvider):
         """
         start_date = self._parse_date_time(start_date, tzinfo=tzinfo)
         end_date = self._parse_date_time(end_date, tzinfo=tzinfo)
-        ts = self.generator.random.randint(start_date, end_date)
+        if end_date - start_date <= 1:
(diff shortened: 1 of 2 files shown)


acdha:
That looks good to me

PR #675, opened 2018-01-11
Fix/master/ro locale names

vicusbass:
This adds Romanian locale to Faker, person provider (gender based)

### DIFF
--- a/faker/providers/person/ro_RO/__init__.py
+++ b/faker/providers/person/ro_RO/__init__.py
@@ -1,0 +1,111 @@
+# coding=utf-8
+from __future__ import unicode_literals
+from .. import Provider as PersonProvider
+
+
+class Provider(PersonProvider):
+    formats_female = (
+        '{{first_name_female}} {{last_name}}',
+        '{{first_name_female}} {{last_name}}',
+        '{{first_name_female}} {{last_name}}',
+        '{{first_name_female}} {{first_name_female}} {{last_name}}'
+    )
+    formats_male = (
+        '{{first_name_male}} {{last_name}}',
+        '{{first_name_male}} {{last_name}}',
+        '{{first_name_male}} {{last_name}}',
+        '{{first_name_male}} {{first_name_male}} {{last_name}}'
+    )
+
+    first_names_female = (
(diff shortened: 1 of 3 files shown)


fcurella:
Thank you @vicusbass :)

Could you add the source of your data as comments above the list of names?

vicusbass:
Done, added comments.
Basically, scrapping Wikipedia and checking the list manually is one source. Myself, another source, I added some well-known Romanian last names to the list.

faker/providers/person/ro_RO/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+        'Pompiliu', 'Radu', 'Rafael', 'Rareș', 'Raul', 'Răducu', 'Răzvan', 'Relu', 'Remus', 'Robert', 'Romeo',
+        'Romulus',
+        'Sabin', 'Sandu', 'Sandu', 'Sava', 'Sebastian', 'Sergiu', 'Sever', 'Severin', 'Silvian', 'Silviu', 'Simi',
+        'Simion',
+        'Sinică', 'Sorin', 'Stan', 'Stancu', 'Stelian', 'Șerban', 'Ștefan', 'Teodor', 'Teofil', 'Teohari', 'Theodor',
+        'Tiberiu', 'Timotei', 'Titus', 'Todor', 'Toma', 'Traian', 'Tudor', 'Valentin', 'Valeriu', 'Valter', 'Vasile',
+        'Vasilică', 'Veniamin', 'Vicențiu', 'Victor', 'Vincențiu', 'Viorel', 'Visarion', 'Vlad', 'Vladimir', 'Vlaicu',
+        'Voicu', 'Zamfir', 'Zeno')
+
+    # sources: https://ro.wikipedia.org/wiki/Lista_celor_mai_uzuale_nume_de_familie#Rom%C3%A2nia and myself
```
fcurella: Could you please remove yourself from the list? One of the reasons we want sources is to delegate the editorial responsibility away from us. By modifying the list we are making an editorial choice and we can't claim that the list comes from there anymore.

vicusbass: Makes sense, removed the extra names.

fcurella:
Thank you! ✨

PR #676, opened 2018-01-11
en_GB phone and SSN localisation

TheSapper:
### What does this changes

Added all possible fake UK phone numbers from batches reserved for this purpose by Ofcom.
Added en_GB localisation for SSN (UK National Insurance Number).

### What was wrong

Genuine phone numbers may have been generated.
No en_GB SSN localisation.

### How this fixes it

The more specific formatting forces only numbers from reserved pools to be generated.
Localisation for UK NINO format added to SSN provider.


fcurella:
Nice work! Thank you! ✨

PR #677, opened 2018-01-11
user_agent

### LINKED ISSUE #673: Update user_agent chrome version
Right now provider **user_agent** can return chrome version between 13-15 which is too small (for example latest stable version is 63). I want to create PR to fix this.

lcd1232:
### What does this changes

- Increase chrome version
- Add optional parameters for function chrome()
- Increase mac os version

### What was wrong

Too old version of chrome version

### How this fixes it

Increase version to 63 and add optional parameters to change chrome version and build

Fixes #673


fcurella:
Thank you! ✨

PR #678, opened 2018-01-12
Improve email function

### LINKED ISSUE #661: NameError: name 'args' is not defined
I'm getting an error using the *args method.

```
fake.email(*args, **kwargs)
NameError: name 'args' is not defined

```
Also can I change the **kwargs with @gmail to get random@gmail.com?

lcd1232:
### What does this changes

Add optional argument domain which you can use your domain, not randomly generated.

### What was wrong

function **email** returns random domain each time, so user cannot use it for single domain.

### How this fixes it

Add optional argument domain

Fixes #661

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -75,9 +75,13 @@ class Provider(BaseProvider):
         return string
 
     @lowercase
-    def email(self):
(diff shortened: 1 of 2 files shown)


fcurella:
seems like a good idea :) Thank you ✨

PR #679, opened 2018-01-12
ip private ability

lcd1232:
### What does this changes

- Add optional parameter _private_ for **ipv4** and **ipv6** which gives ability to return ip address from private network

### What was wrong

You can't choose which ip do you need.

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -5,7 +5,7 @@ from text_unidecode import unidecode
 
 from .. import BaseProvider
 
-from ipaddress import ip_address, ip_network, IPV4LENGTH, IPV6LENGTH, _IPv4Constants, _IPv6Constants
+from ipaddress import ip_address, ip_network, IPV4LENGTH, IPV6LENGTH, IPv4Network, IPv6Network
 
 # from faker.generator import random
 # from faker.providers.lorem.la import Provider as Lorem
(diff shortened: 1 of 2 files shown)


faker/providers/internet/__init__.py
```diff
@@ -5,10 +5,11 @@
 
 from .. import BaseProvider
 
-from ipaddress import ip_address, ip_network, IPV4LENGTH, IPV6LENGTH
+from ipaddress import ip_address, ip_network, IPV4LENGTH, IPV6LENGTH, _IPv4Constants, _IPv6Constants
```
fcurella: I don't htink we should use undocumented feature of the stdlib. They might change in the future and break `Faker`

faker/providers/internet/__init__.py
```diff
@@ -169,20 +170,56 @@ def url(self):
         pattern = self.random_element(self.url_formats)
         return self.generator.parse(pattern)
 
-    def ipv4(self, network=False):
-        """Produce a random IPv4 address or network with a valid CIDR"""
-        address = str(ip_address(self.generator.random.randint(
-            0, (2 ** IPV4LENGTH) - 1)))
-        if network:
+    def ipv4(self, network=False, private=False, simple=True):
```
fcurella: I'm -1 on `simple`. Users should not be expected to worry about how `Faker` generates the value.

ZuluPro: Yes -1 on simple

lcd1232:
@fcurella done

sosiska:
ping @fcurella

faker/providers/internet/__init__.py
```diff
@@ -169,20 +199,50 @@ def url(self):
...
-            0, (2 ** IPV4LENGTH) - 1)))
-        if network:
+    def ipv4(self, network=False, private=False):
+        """Produce a random IPv4 address or network with a valid CIDR."""
+        if private is True:
+            networks = dict()
+            for i in range(len(_private_ipv4_networks)):
+                networks[i] = _private_ipv4_networks[i].num_addresses
+            k = self.generator.random_element(networks)
+            _net = _private_ipv4_networks[k]
```
fcurella: I'm not sure I'm following this logic.  I don't see the point of the `for` loop. Seems to me that it's picking up a random item of the `_private_ipv4_networks` list to assign to `_net` and could be done with just `_net = self.generator.random.choice(_private_ipv4_networks)`

fcurella: would you put this explanation in a comment?

faker/providers/internet/__init__.py
```diff
@@ -169,20 +199,50 @@ def url(self):
...
+            address = str(ip_address(self.generator.random.randint(
+                int(_net.network_address),
+                int(_net.network_address) + _net.num_addresses - 1
+            )))
+        else:
+            # It's unfair to trying to not find private ip but it's much faster
+            address = str(ip_address(self.generator.random.randint(
+                0, (2 ** IPV4LENGTH) - 1)))
+            while ip_address(force_text(address)).is_private is True:
```
fcurella: I would default the `private` argument to `None`, and have this check only `if private is False`

fcurella:
@lcd1232 do you mind reviewing https://github.com/joke2k/faker/pull/703 ? I think we're going to go with that implementation, since it gives us more flexibility (namely, users could choose address classes)

lcd1232:
@fcurella yes, I'm agree with you. Should I close this PR?

fcurella:
@lcd1232 yes, this one can be closed.

PR #682, opened 2018-01-15
Update __init__.py

Pomax:
Some words that can lead to *really* bad combinations have been removed in this PR so that people don't need to worry as they show a staging/test thing to their managers and suddenly being the center of a textual scandal. Before this PR, phrases like "die die die hang", "finger beautiful daughter" or "sexual man boy love" were all too common, and removing the possibility of people getting into genuine trouble for an out-of-the-box word list is probably not worth it.

I'm sure I missed some that others will have run into as being super problematic, and a better wordlist is available out there that has already been pruned, but until someone files a PR with a full list substitution I hope this PR can make it in with a new version pushed to pip.

Addresses #683


fcurella:
If this turns out to not be safe enough, i’d OK with removing the Loren provider for every language and leave the faux-Latin only.

PR #688, opened 2018-01-20
Fix indented list in changelog

153957:
### What does this changes

Remove indent for list in change log

### What was wrong

Looked like a quoted list on GitHub

### How this fixes it

Without indent it looks correct, like the other changes.


fcurella:
Thank you! ✨

Issue #692, opened 2018-01-31
Url provider schemes selection


ProvoK:
It could be very nice to have the possibility to generate url with control over the scheme chosen.

### Steps to reproduce
```python
from faker import Faker
fake = Faker()
print([fake.url() for i in range(10)])  # Statistically it will generate at least one 'https' url
# Output
['http://silva-choi.com/', 'http://lawson.net/', 'https://powell.com/', 'http://parker.com/', 'http://www.ortega.com/', 'https://www.davis.com/', 'http://maxwell.info/', 'http://lopez.com/', 'https://www.sanders.com/', 'http://www.baker-meadows.com/']
```
### Expected behavior
Something similar to this
```python
from faker import Faker
fake = Faker()
fake.url(http=False)  # Will generate only 'https' urls
fake.url(https=False)  # Will generate only 'http' urls
fake.url(https=True, http=True)  # Same as fake.url() (actual behaviour)
fake.url(https=False, http=False)  # Raise error
```
obviously this signature is purely for explaining the use case. Could be something like
```python
fake.url(exclude_scheme='http') # Will generate only 'https' urls
fake.url(exclude_scheme='https') # Will generate only 'http' urls
```
or with a enum
```python
fake.url(exclude_scheme=Scheme.HTTP) # Will generate only 'https' urls
fake.url(exclude_scheme=Scheme.HTTPS) # Will generate only 'http' urls
```

I could start working on the issue, if this behavior is acceptable 🙂

ProvoK:
I think this issue can be closed now @fcurella

PR #693, opened 2018-02-01
Implement schema selection for internet url provider

ProvoK:
### What does this changes

Adds the ability to generate url with control over the scheme.

### What was wrong

It is currently not possible. A random `http` or `https` url is generated.
Related to: https://github.com/joke2k/faker/issues/692

### How this fixes it

Adding optional parameter to internet provider `url` method.

The actual signature change is my proposal, i don't know if it's reasonable or good enough.
I'll do better test cases (and implementation?) if this change in behavior will be accepted.

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -46,10 +46,8 @@ class Provider(BaseProvider):
         '{{user_name}}@{{free_email_domain}}',
     )
     url_formats = (
-        'http://www.{{domain_name}}/',
-        'http://{{domain_name}}/',
-        'https://www.{{domain_name}}/',
-        'https://{{domain_name}}/',
+        '://www.{{domain_name}}/',
+        '://{{domain_name}}/',
     )
     uri_formats = (
         '{{url}}',
@@ -165,22 +163,12 @@ class Provider(BaseProvider):
     def tld(self):
         return self.random_element(self.tlds)
 
-    def _filter_url_formats(self, exclude_scheme):
-        if not isinstance(exclude_scheme, str):
-            raise ValueError('exclude_scheme param should be a string')
-        if exclude_scheme.lower() not in ['https', 'http']:
-            raise ValueError('exclude_scheme param should either `http` or `https`')
+    def url(self, schemes=None):
+        if not schemes:
+            schemes = ['http', 'https']
 
-        pattern = '{}://'.format(exclude_scheme)
-        return filter(lambda f: not f.startswith(pattern), self.url_formats)
+        pattern = '{}{}'.format(self.random_element(schemes), self.random_element(self.url_formats))
 
-    def url(self, exclude_scheme=None):
-        formats_allowed = self.url_formats
-
-        if exclude_scheme:
-            formats_allowed = self._filter_url_formats(exclude_scheme)
-
-        pattern = self.random_element(formats_allowed)
         return self.generator.parse(pattern)
 
     def ipv4(self, network=False):
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -165,8 +165,22 @@ class Provider(BaseProvider):
     def tld(self):
         return self.random_element(self.tlds)
 
-    def url(self):
-        pattern = self.random_element(self.url_formats)
+    def _filter_url_formats(self, exclude_scheme):
+        if not isinstance(exclude_scheme, str):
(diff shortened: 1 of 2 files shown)


ProvoK:
I don't really know why it's not working on certain python versions 😐 . I'll take a look asap.

faker/providers/internet/__init__.py
```diff
@@ -165,8 +165,22 @@ def domain_word(self,):
...
+    def _filter_url_formats(self, exclude_scheme):
+        if not isinstance(exclude_scheme, str):
+            raise ValueError('exclude_scheme param should be a string')
+        if exclude_scheme.lower() not in ['https', 'http']:
+            raise ValueError('exclude_scheme param should either `http` or `https`')
+
+        pattern = '{}://'.format(exclude_scheme)
+        return filter(lambda f: not f.startswith(pattern), self.url_formats)
+
+    def url(self, exclude_scheme=None):
```
fcurella: I would rather have this more flexible and change the signature to something like:

```python
def url(self, schemes=None):
    if schemes is None:
        schemes = ['http', 'https']
```
This way people can generate any kind of scheme.

ProvoK: Sure, I'm totally fine with that :)
I thought `exclude_scheme` were reasonable since actually only `http` and `https` schemes are generated.

I'm going to work on that asap, but I've a couple of questions.
What the expected behavior for the following scenarios?
```python
fake = Faker()

# One or all scheme/s does not "exists". 
fake.url(schemes=['foo', 'http'])   # foo://my-bar.org or error?
fake.url(schemes=['foo', bar'])

# Duplicates, raise an error?
fake.url(schemes=['foo', foo'])
fake.url(schemes=['http', 'http'])
```

swizzard: Why does `Faker` need to care about which schemes are real or not?

fcurella: @ProvoK:

* `schemes=['foo', bar']` will generate urls with `foo://` or `bar://`. As @swizzard say, `Faker` does't need to care if the scheme is real or not.
* `schemes=['foo', foo']` normalizes to a Set, so somehting like `schemes = Set(schemes)`

ProvoK: @swizzard 
Actually, I was questioning about it since i was not sure what would be the best choice. If the generation of the url/email/whatever is too much permissive, a random string could be the same and there would be not the need of faker. However this seems not the case. :)

@fcurella 
got it!

thanks guys :+1:

ProvoK: @fcurella 
I've done changes as requested. Tell me if there is something more i should/can do :smile:

faker/providers/internet/__init__.py
```diff
@@ -165,8 +163,12 @@ def domain_word(self,):
     def tld(self):
         return self.random_element(self.tlds)
 
-    def url(self):
-        pattern = self.random_element(self.url_formats)
+    def url(self, schemes=None):
+        if not schemes:
```
fcurella: use `if schemes is None:`. This way, users can pass an empty list and have schemeless URL ('malformed' urls like that are often used in HTML pages to load a resource regardless of http/https)

faker/providers/internet/__init__.py
```diff
@@ -165,8 +163,12 @@ def domain_word(self,):
...
+            schemes = ['http', 'https']
+
+        pattern = '{}{}'.format(self.random_element(schemes), self.random_element(self.url_formats))
```
fcurella: make sure `schemes` is not an empty list before calling `self.random_element(schemes)`

faker/providers/internet/__init__.py
ProvoK: In this case would it be something like `www.domain.com`, and not `://www.domain.com`, i suppose

fcurella: No, the expected output for `schemes=[]` would be a protocol-relative URL like `://www.domain.com`

ProvoK: Done :+1: I added a bit of documentation too

ProvoK: I don't understand how coveralls check had failed, it shows decreased coverage for parts I've not touched or whatever

fcurella:
@ProvoK don't worry about coverage. It's a little falke because the tests are randomized and every once in a while it just so happens that an edge case doesn't get hit.

PR #695, opened 2018-02-02
An en_NZ provider for addresses, phone numbers and email addresses

### LINKED ISSUE #694: Locale provider for en_NZ
Should be a simple job to clone the en_AU provider(s), and adjust. Might be nice to add plausible looking Māori placenames too.

doctorlard:
- Based loosely on the en_AU code.
- Also fakes Māori placenames, using common nouns and adjectives.
Fixes #694


faker/providers/address/en_NZ/__init__.py
```diff
@@ -0,0 +1,256 @@
...
+        # Christchurch
+        '80##',
+        '81##',
+        '82##',
+        '84##',
+        '85##',
+        '86##',
+        '88##',
+        '89##',
+        # The bottom of the Wurrrld
```
fcurella: No LOL please :) Does this administrative area have a name? If I'm reading wikipedia right, it should be Southland

doctorlard: ah - whoops :-P

faker/providers/person/en_NZ/__init__.py
```diff
@@ -0,0 +1,627 @@
...
+        '{{first_name_male}} {{last_name}}-{{last_name}}',
+        '{{first_name_female}} {{last_name}}',
+        '{{first_name_female}} {{last_name}}',
+        '{{first_name_female}} {{last_name}}',
+        '{{first_name_female}} {{last_name}}',
+        '{{first_name_female}} {{last_name}}-{{last_name}}',
+    )
+
+    first_names_male = (
+        # Copied from en_GB
```
fcurella: I would rather compile a list from https://www.dia.govt.nz/diawebsite.nsf/wpg_URL/Services-Births-Deaths-and-Marriages-Most-Popular-Male-and-Female-First-Names?OpenDocument. Make sure you take English and Māori names from the same birth year, and cite the source URL in a comment

doctorlard: First of all that would only be representative of people born in the last few years. Secondly, how do want to deal with the fact that New Zealanders don't have name suffixes? Setting `suffixes_male` and `suffixes_female` to empty tuples results in random.choice raising an IndexError.

doctorlard:
I'm trying to get the tests passing first. The tox and Py 2.7 build tests seem to be choking on "Māori" being in a comment line, despite the magic PEP-263 comment at the top.

doctorlard:
For now I have this:
```
 suffixes_female = ('',)
 suffixes_male = ('',)
```
Which at least passes.

faker/providers/person/en_NZ/__init__.py
fcurella: if you don't set `suffixes_*` at all, it should just work. the superclass' method (https://github.com/joke2k/faker/blob/master/faker/providers/person/__init__.py#L81) already checks if the attribute exists and returns an empty string if it doesn't.

fcurella: > that would only be representative of people born in the last few years

I think I'd be ok with that. If you can find a dataset for an earlier year that has both english and māori names, that'd would work too. I'd just rather use local data.

faker/providers/person/en_NZ/__init__.py
```diff
@@ -0,0 +1,627 @@
...
+        ('Huang', 0.2),
+        ('Wu', 0.2),
+        ('Yang', 0.1),
+        ('Xu', 0.1),
+        ('Wong', 0.07),
+        ('Zhou', 0.05),
+    ))
+
+    # No name suffixes in New Zealand
+    suffixes_female = ('',)
```
fcurella: These can be removed

doctorlard: Removing these results in "Jr." or "III" on the end of people's names, which only happens in New Zealand in re-runs of *Dallas*.

doctorlard: I take it back, you're right. That's odd, it wasn't behaving like that before. I blame __pycache__ :-)

doctorlard:
@fcurella - I've updated this commit with data from DIA, and surnames culled from cemetery plot and electoral roll data.

faker/providers/person/en_NZ/__init__.py
```diff
@@ -0,0 +1,1419 @@
...
+    ))
+
+    first_names = first_names_male.copy()
+    first_names.update(first_names_female)
+
+    # New Zealand surnames compiled (and cleaned up) from the following sources:
+    #
+    # NZ Cemetery plot data:
+    #    https://catalogue.data.govt.nz/dataset?q=cemetery+plots
+    # NZ Electoral Roll (not publicly available):
```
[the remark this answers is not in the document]
doctorlard: It turns out I didn't use that source, because it wasn't even available to me in end (it's tightly controlled by NZ legislation). I will remove the comment.

faker/providers/person/en_NZ/__init__.py
```diff
@@ -0,0 +1,1419 @@
...
+        ("Tiare", 671),
+        ("Tui", 1251),
+        ("Waimarie", 671),
+        ("Wikitoria", 583),
+    ))
+
+    first_names = first_names_male.copy()
+    first_names.update(first_names_female)
+
+    # New Zealand surnames compiled (and cleaned up) from the following sources:
```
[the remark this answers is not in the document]
doctorlard: Yes. And entries like "Unnamed infant" which we probably don't want to turn up in name lists.

fcurella:
LGTM! Thank you so much! ✨

PR #698, opened 2018-02-08
Increase lower bound on AD date generation

### LINKED ISSUE #696: `date_time_ad` nondeterministically errors
Occasionally, `date_time_ad` raises `OverflowError: date value out of range`.

### Steps to reproduce

1. Substitute the random offset generation in `date_time_ad` (at the time of writing, this is `ts = self.generator.random.randint(-62135600400, end_datetime)`) for the specific value `ts = -62135596801`.
2. Call `date_time_ad()`

### Expected behavior

`date_time_ad` returns a value of type `datetime` (a naive datetime).

### Actual behavior

`date_time_ad` raises `OverflowError: date value out of range`

I have reproduced this both on a Debian Jessie machine running Python 3.5.3 and on a macOS High Sierra 10.13.2 machine running Python 3.6.4.

The error appears to occur whenever the random `ts` generation generates a value strictly less than -62135596800.

Suggested fix would be to substitute the current lower bound of `-62135600400` for a new lower bound of `-62135596800`.

prophile:
### What does this changes

Increases the lower bound on AD date generation by one hour.

### What was wrong

The lower bound was one hour too far, which meant that it was possible to pick dates logically as far as one hour into BCE. This led to an OverflowError.

### How this fixes it

As far as the lowest possible value for `ts` now, this does not generate an `OverflowError` for underflow into BCE.

Fixes #696

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -1209,7 +1209,7 @@ class Provider(BaseProvider):
         :return datetime
         """
         end_datetime = self._parse_end_datetime(end_datetime)
-        ts = self.generator.random.randint(-62135600400, end_datetime)
+        ts = self.generator.random.randint(-62135596800, end_datetime)
         # NOTE: using datetime.fromtimestamp(ts) directly will raise
         #       a "ValueError: timestamp out of range for platform time_t"
         #       on some platforms due to system C functions;


fcurella:
Could you update the method's docstring to say `Get a datetime object for a date between January 1, 1970 @ 1AM and now`?

prophile:
Assuming you meant 1 rather than 1970?

I could, but that is not its behaviour: the docstring is now correct.

fcurella:
@prophile sorry, I mistyped. I meant `between January 1, 001 @ 1AM and now`

prophile:
Apologies: I figured. To be clear, "January 1, 001 @ 1AM" is what I was referring to when I said "that is not its behaviour", I wasn't just being obtuse :)

The new offset from January 1, 1970 is, correctly, January 1 001 @ midnight. Previously it would have been December 31, 001 BC @ 23:00.

fcurella:
@prophile my mistake, I misread the code.

Nice work!

PR #699, opened 2018-02-11
Address pt

ZuluPro:
### What does this changes

I included more set and generator to address/pt_PT.
Data are from:
- https://pt.wikipedia.org/wiki/Distritos_de_Portugal
- https://pt.wikipedia.org/wiki/Lista_de_freguesias_de_Portugal

### What was wrong

Nothing, just improvements

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,6 @@ nosetests.xml
 .ropeproject
 .DS_Store
 .venv
+
+# IDE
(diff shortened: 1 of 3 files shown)


ZuluPro:
Any news ?

fcurella:
since you have the sources, would you mind adding them as comments?

PR #701, opened 2018-02-12
issue #700: Added the ability to specify the min and max age for some…

### LINKED ISSUE #700: Add the ability to specify the min and max age for ssn numbers.
Some locale specific ssn numbers are dependent on the age of the person. In some use cases it is handy to be able to specify the age interval for a generated ssn number.

frangiz:
### What does this changes

It will be possible to specify the min and max age when generating a ssn number (for those locales that directly supports it).

### What was wrong

Could not generate ssn numbers for people with an age less than 16/18 or olrder than 90.

### How this fixes it

Added two optional parameters for the min and max age.

Fixes #700

### DIFF
--- a/faker/providers/ssn/et_EE/__init__.py
+++ b/faker/providers/ssn/et_EE/__init__.py
@@ -29,12 +29,10 @@ def checksum(digits):
 
 
 class Provider(SsnProvider):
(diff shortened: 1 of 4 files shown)


fcurella:
Thanks! ✨

PR #703, opened 2018-02-15
Added Private IPv4

ZuluPro:
### What does this changes

Added private IPv4 method. I designed it with IP classes (A, B, C)
https://en.wikipedia.org/wiki/Classful_network

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -201,19 +201,48 @@ class Provider(BaseProvider):
         pattern = self.random_element(self.url_formats)
         return self.generator.parse(pattern)
 
-    def ipv4(self, network=False):
-        """Produce a random IPv4 address or network with a valid CIDR"""
-        address = str(ip_address(self.generator.random.randint(
-            0, (2 ** IPV4LENGTH) - 1)))
+    def ipv4(self, network=False, address_class=None, private=None):
+        """
+        Produce a random IPv4 address or network with a valid CIDR.
+
+        :param network: Network address
+        :param address_class: IPv4 address class (a, b, or c)
+        :param private: Public or private
+        :returns: IPv4
+        """
+        if private is True:
+            return self.ipv4_private(address_class=address_class,
+                                     network=network)
+        elif private is False:
+            return self.ipv4_public(address_class=address_class,
+                                    network=network)
+        if address_class is None:
+            ip_range = (0, (2 ** IPV4LENGTH) - 1)
+        else:
+            net = IPV4_PUBLIC_NETS[address_class]
+            ip_range = (net._ip[0], net.ip[-1])
+        address = str(ip_address(self.generator.random.randint(*ip_range)))
         if network:
             address += '/' + str(self.generator.random.randint(0, IPV4LENGTH))
             address = str(ip_network(address, strict=False))
         return address
 
     def ipv4_network_class(self):
+        """
+        Returns a IPv4 network class 'a', 'b' or 'c'.
+
+        :returns: IPv4 network class
+        """
         return self.random_element('abc')
 
-    def ipv4_private(self, address_class=None, network=False):
+    def ipv4_private(self, network=False, address_class=None):
+        """
+        Returns a private IPv4.
+
+        :param network: Network address
+        :param address_class: IPv4 address class (a, b, or c)
+        :returns: Private IPv4
+        """
         address_class = address_class or self.ipv4_network_class()
         min_, max_, netmask = IPV4_NET_CLASSES[address_class]
         address = str(ip_address(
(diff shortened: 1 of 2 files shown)


faker/providers/internet/__init__.py
```diff
@@ -178,6 +178,22 @@ def ipv4(self, network=False):
             address = str(ip_network(address, strict=False))
         return address
 
+    def ipv4_private(self, klass=None, network=False):
```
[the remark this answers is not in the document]
fcurella: Could we merge this logic into the `ipv4` method and have non-private address classes? I'm thinking something like:

```python
def ipv4(self, network=False, private=None, address_class=None):
    if private is None and address_class is None:
        # pretty much the old logic
    if private is True:
        classes = {
            'a': (167772160, 184549375, 24),
            'b': (2886729728, 2887778303, 20),
            'c': (3232235520, 3232301055, 16),
        }
  if private is False:
        classes = {
            'b': (2886729728, 2887778303, 16),
            'c': (3232235520, 3232301055, 8),
        }
```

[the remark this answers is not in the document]
fcurella: should we have the same for IPV6 addresses?

fcurella:
alternative implementation of https://github.com/joke2k/faker/pull/703

faker/providers/internet/__init__.py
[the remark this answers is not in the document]
ZuluPro: > I would use address_class instead of klass
Ok

> Could we merge this logic into the ipv4 method and have non-private address classes?
I don't think we should merge Unaware/Public/Private. 

> should we have the same for IPV6 addresses?
Yes of course, but another PR ?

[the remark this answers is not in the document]
fcurella: > I don't think we should merge Unaware/Public/Private.

Could you elaborate why that would be a bad idea? I'd like to keep the number of methods to a minimum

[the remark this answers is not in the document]
ZuluPro: Yes, of course,
1. To get private or public is a little complex and separated code is better.
2. Like the `person` provider with male & female, we should provide simple way to get private or public IP.

ZuluPro:
I'm doing `ipv4_public`

ZuluPro:
@fcurella 
What do you think about the code ?
Do you know how to fix this Python 2 bug ?

faker/providers/internet/__init__.py
```diff
@@ -178,6 +210,45 @@ def ipv4(self, network=False):
...
+            address = str(ip_network(address, strict=False))
+        return address
+
+    def ipv4_public(self, network=False, address_class=None):
+        address_class = address_class or self.ipv4_network_class()
+        public_net = IPV4_PUBLIC_NETS[address_class]
+        private_blocks = IPV4_PRIVATE_NET_BLOCKS[address_class]
+        # Make valid IP ranges
+        net_ranges = []
+        if public_net[0] != private_blocks[0][0]:
```
fcurella: Could you explain why do we need this check? This code is quite hard to follow, comments would help

faker/providers/internet/__init__.py
```diff
@@ -178,6 +210,45 @@ def ipv4(self, network=False):
             address = str(ip_network(address, strict=False))
         return address
 
+    def ipv4_network_class(self):
+        return self.random_element('abc')
+
+    def ipv4_private(self, address_class=None, network=False):
```
fcurella: eventually I want to remove `ipv4_private` and `ipv4_public` from the api in favor of a `private=None` argument for `ipv4`

fcurella:
> Do you know how to fix this Python 2 bug

I think `py2-ipaddress` is our best option in terms of maintainability.

fcurella:
On a second thought, you could use `force_text` from `faker.utils.text`

faker/providers/internet/__init__.py
fcurella: n/m this comment, I've read a little bit and know I understand that private networks are in multiple classes

fcurella: I think the best way to go about this is to have:
```python
def ipv4(self, network=False, address_class=None, private=None):
    if address_class is not None or private is not None:
        if private is True:
            return self.ipv4_private(address_class=address_class, network=network)
        else:
            return self.ipv4_public(address_class=address_class, network=network)
    # old logic here...
```

faker/providers/internet/__init__.py
ZuluPro: I check if the 1st available range starts after the 1st private block.
Example:
Class C: 192.0.0.0/24
The first address 192.0.0.0 is in 192.0.0.0/8 which is private.
So the 1st `net_ranges` will be from 192.0.0.8 to 192.X.X.X

I'll add comments

ZuluPro:
Ok,
- Comments added
- Docstring added
- Test added
- `ipv4` improved
- Python 2 fixed with `text.force_text`

faker/providers/internet/__init__.py
```diff
@@ -169,15 +201,90 @@ def url(self):
...
+        :param private: Public or private
+        :returns: IPv4
+        """
+        if private is True:
+            return self.ipv4_private(address_class=address_class,
+                                     network=network)
+        elif private is False:
+            return self.ipv4_public(address_class=address_class,
+                                    network=network)
         address = str(ip_address(self.generator.random.randint(
```
fcurella: can we add the option to return any IP (without checking if public or private) from a specific address_class?

ZuluPro: I prefer to not provider address class for `ipv4()`, because I would like this method to provide any IP.
From private/public, from reserved blocks, etc

fcurella: the `address_class` will optional, and default to the old algorithm of returning any IP. We would filter by address class only if `address_class` is specified

ZuluPro:
@fcurella  Changed !

fcurella:
Thank you so much for working on this! ✨

PR #705, opened 2018-02-15
NEW: Added dutch nl_NL bank provider

PatSousa:
### What does this changes

Added Dutch iban provider according to specs:
https://en.wikipedia.org/wiki/International_Bank_Account_Number

### What was wrong
This support was missing for this language

### How this fixes it
Added configuration according to standard


fcurella:
Thank you! ✨

PR #706, opened 2018-02-20
Added street names to the Ukrainian address provider

cadmi:
### What does this changes

Added street titles to the Ukrainian address provider

### What was wrong

This support was missing for this language

### How this fixes it

Added configuration


fcurella:
Looks good! Thank you! ✨

PR #707, opened 2018-02-24
Added hexify

ZuluPro:
### What does this changes

In the same spirit than lexify, I added hexify, to format strings with hexadecimal numbers.

I could be useful in many case, especially to develop providers.

### DIFF
--- a/faker/providers/__init__.py
+++ b/faker/providers/__init__.py
@@ -213,3 +213,13 @@ class BaseProvider(object):
         :returns: string with all numerical and letter placeholders filled in
         """
         return self.lexify(self.numerify(text), letters=letters)
+
+    def hexify(self, text='????'):
+        """
+        Replaces all question mark ('?') occurrences with a random
+        hexadecimal character.
+
+        :param text: string to be parsed
+        :returns: string with all letter placeholders filled in
+        """
+        return self.lexify(text, letters=string.hexdigits[:-6])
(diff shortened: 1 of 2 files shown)


faker/providers/__init__.py
```diff
@@ -213,3 +213,17 @@ def bothify(self, text='## ??', letters=string.ascii_letters):
         :returns: string with all numerical and letter placeholders filled in
         """
         return self.lexify(self.numerify(text), letters=letters)
+
+    def hexify(self, text='????', upper=False):
```
fcurella: I would try to come up with a different character than `?` or `#`, maybe `x`?

ZuluPro: No, `x` is too ASCII, more `$` ?

fcurella: how about `×` (unicode 'multiplication sign')?

ZuluPro: How do you do it ? 😄

fcurella: good point, i think it's too inconvenient. I'm worried `$` is too close to the idea of 'currency'. Maybe we could use `*`, `^` or `&`

ZuluPro: What about `~` ?
Easy to do for everyone, without being used all the time.

ZuluPro: In fact, forget my last comment, let's do `^` !

ZuluPro:
@fcurella Changed !

fcurella:
Thank you!

PR #708, opened 2018-02-25
Added address provider for Austria (DeAT).

bessl:
### What does this changes

Add address namespaces for Austria.

### DIFF
--- a/faker/providers/address/de_AT/__init__.py
+++ b/faker/providers/address/de_AT/__init__.py
@@ -1,0 +1,152 @@
+# coding=utf-8
+
+from __future__ import unicode_literals
+from .. import Provider as AddressProvider
+
+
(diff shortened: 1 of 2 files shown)


faker/providers/address/de_AT/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+    def street_suffix_long(self):
+        return self.random_element(self.street_suffixes_long)
+
+    def city_name(self):
+        return self.random_element(self.cities)
+
+    def state(self):
+        return self.random_element(self.states)
+
+    def country(self):
```
fcurella: we don't need to override this method if it's exactly the same as the superclass

fcurella:
Thank you! ✨

PR #709, opened 2018-02-25
Duplicated nl_NL for nl_BE

quintendewilde:
#709 For first and last names!

### What does this changes

I've added a duplicate of the nl_NL version, for similarities in language and regions it's pretty good start. Until I populate them with a better or more correct belgian dataset of names.

### What was wrong
When using nl_Be for profiles, it gave me us_Us names.


fcurella:
I'll wait until you populate this provider with a dataset from Belgium

PR #713, opened 2018-03-03
Capital 'Ø' for last name 'Ødegård', and alphabetically at the end 

cloveras:
### What does this changes

The last name 'ødegård' was spelled with a lowercaser 'ø' and not sorted correctly alphabetically. Now spelled 'Ødegård', and placed alphabetically at the end of the list, as all three Norwegian special characters are (in Norway, at least).

### What was wrong

Lowercase 'ø' as first letter.

### How this fixes it

See above.

Fixes #...


fcurella:
Thank you! ✨

PR #716, opened 2018-03-07
Added a provider for a Norwegian ssn number.

### LINKED ISSUE #714: Norwegian SSN: Incorrect format
The format for Norwegian SSN is 11 digits: 
- The first six digits represent the date of birth in the order date, month, year
- The next three digits are an individual number, the third digit of which indicates gender – even numbers for women and odd numbers for men
- The last two digits are control digits

More details: http://www.skatteetaten.no/en/person/National-Registry/Birth-and-name-selection/Children-born-in-Norway/National-ID-number/

### Steps to reproduce

from faker import Faker
fake = Faker('no_NO')
print(fake.ssn())

The SSN format for 'no_NO' is currently '465-18-3530': https://faker.readthedocs.io/en/master/locales/no_NO.html#faker-providers-ssn 

### Expected behavior

SSN should be on the format specified here:
http://www.skatteetaten.no/en/person/National-Registry/Birth-and-name-selection/Children-born-in-Norway/National-ID-number/

fodselsnummer.py here may be useful: https://github.com/magnuswatn/fodselsnummer

### Actual behavior

SSN returned is on the format '465-18-3530'.

frangiz:
### What does this changes

Added a Norwegian ssn provider.

### What was wrong

The was no provider for a Norwegian ssn number.

### How this fixes it

Since there was no provider for a no_NO number, the default provider was used and resulted in incorrect format and number.

Fixes #714


frangiz:
Should maybe add a dateofbirth argument and use that first and if that is not used, then use min_age/max_age.

fcurella:
@frangiz as per discussion in #714, the signature should match the [`hu_HU` provider](https://github.com/joke2k/faker/blob/master/faker/providers/ssn/hu_HU/__init__.py#L16):

```
def ssn(self, dob=None, gender=None):
```

Where `dob` can be only by a string in `YYMMDD` format and `gender` can only by `'M'` or `'F'`

frangiz:
Has to be on the form YYYYMMDD since the ssn is dependent on the century of the dob.

fcurella:
@frangiz we can make `dob` be in the `YYYYMMDD` format. Let's also add a check to make sure users are passing the arguments in the correct format.

frangiz:
Will try it.

cloveras:
It would be great to have Norwegian SSNs, and I would like to help, but not sure where to begin. How can the 0.02% drop in coverage be fixed?

fcurella:
@cloveras don't worry about the coverage drop

frangiz:
@cloveras , I think that the Norwegian SSN is finished. All depends if @fcurella wants to accept the PR or not.

fcurella:
LGTM! Thank you! ✨

cloveras:
Sorry if I'm missing something here, but I think this may not provide correct Norwegian SSNs.

Using December 31 1999 as the date of birth (YYYYMMDD):
ssn('19991231','M')
 
This results in SSNs like these:
99123145123
99123108724
99123127532

These start with the date of birth in YYMMDD format, but the correct format is DDMMYY.

Possibly a quick fix, @frangiz?

frangiz:
@cloveras, ohh that is correct. Can create a fix PR this weekend.

PR #720, opened 2018-03-10
Add latitude and longitude to Austria Provider

bessl:
### What does this changes

Geo coordinates in the area from Austria.
Latitude and Longitude for the de_AT provider.


fcurella:
Thank you! ✨

PR #726, opened 2018-03-13
NEW: Added Norwegian no_NO bank provider

cloveras:
11 digits. 

### What does this changes

Norwegian bank accounts have 11 digits: https://www.ecbs.org/iban/norway-bank-account-number.html

1-4: Bank identifier
5-10: Account identifier
11: Control digit

### What was wrong

Even with no_NO locale, calls to fake.bban() returned en_GB account numbers.

### How this fixes it

Returns 11 digits. No check for bank identifier (first 4 digits) or control digit (last digit).

### DIFF
--- a/faker/providers/bank/no_NO/__init__.py
+++ b/faker/providers/bank/no_NO/__init__.py
@@ -2,4 +2,4 @@ from .. import Provider as BankProvider
 
 class Provider(BankProvider):
     bban_format = '###########'
-country_code = 'NO'
+    country_code = 'NO'
--- a/faker/providers/bank/no_NO/__init__.py
+++ b/faker/providers/bank/no_NO/__init__.py
@@ -1,0 +1,5 @@
+from .. import Provider as BankProvider
+
+class Provider(BankProvider):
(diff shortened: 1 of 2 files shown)


faker/providers/bank/no_NO/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .. import Provider as BankProvider
+
+class Provider(BankProvider):
+    bban_format = '###########'
+country_code = 'NO'
```
fcurella: this is missing indentation

cloveras: Ouch! Sorry. Fixed now.

cloveras:
Will try!

tommythorsen:
Hi @fcurella! I added a unit test, which seems to have made the checks happier. Do you think this PR is mergeable now?

tommythorsen:
@cloveras Yeah, that was a silly copy-paste error. I fixed with my latest commit.

fcurella:
Looks good! thank you! ✨

PR #728, opened 2018-03-14
Use isounidecode over other ports of Text::Unidecode

### LINKED ISSUE #727: text-unidecode is released under the Artistic license

moggers87:
### What does this change

Swaps out `text-unidecode` for `isounidecode`

### What was wrong

`text-unidecode` is released under the Artistic License v1.0, which has some clauses which make it more restrictive than Faker's current license.

`text-unidecode` itself replaced `unidecode` as that package was released under the terms of the GPL, which again is more restrictive than Faker's current license.

### How this fixes it

This library is released under a BSD license
(https://github.com/redvasily/isounidecode/blob/master/LICENSE.txt)
which has none of the problems associated with the GPL or Artistic License that unidecode and
text-unidecode are released under.

Tests pass locally on Python 2.7 and 3.6

Fixes #727

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -1,7 +1,7 @@
 # coding=utf-8
 from __future__ import unicode_literals
 
-from text_unidecode import unidecode
+from isounidecode import unidecode
 
 from .. import BaseProvider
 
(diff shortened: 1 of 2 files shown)


moggers87:
So running tests via `tox -e <env>` works, but running them via `python setup.py test`. This appears to be an upstream issue with `easy_install`: https://github.com/redvasily/isounidecode/issues/2

fcurella:
Hi @moggers87 ,

Thank you very much. I think we should switch, but I'd rather wait for redvasily/isounidecode#2 to be merged in and released

moggers87:
I've pinged the maintainer of that package, but I've had no response so far.

PR #730, opened 2018-03-21
Skillprovider

Lane012:
This is a skill provider feature. Skill provider generates a skill by type (hard 'Database Design' or soft 'Effective Communication') or just pulls a random skill if type doesn't matter. Would like to use this for a resume provider if approved.

### DIFF
--- a/faker/providers/skill/__init__.py
+++ b/faker/providers/skill/__init__.py
@@ -1,0 +1,38 @@
+from .. import BaseProvider
+
+
+class Provider(BaseProvider):
+    """A skill provider that groups skills: hard skills and soft skills."""
+
+    hard_skills = ('Analysis','Benchmarking','Big Data', 'Calculating', 'Coding', 'Configuration', 'Data Analytics',
+                   'Database Design', 'Data Mining', 'Debugging','Documentation', 'Modeling', 'Needs Analysis',
+                   'Statistical Analysis', 'Hardware', 'Implementation','Infrastructure', 'Maintenance', 'Networking',
+                   'Network Security', 'Network Architecture', 'Programming', 'Microsoft Word', 'Linux', 'Servers', 'Security'
+                   'Systems Analysis', 'Testing', 'Engineering', 'Project Planning', 'Task Management', 'Quality Control',
+                   'Task Management', 'Task Delegation', 'Blogging', 'SEO', 'Web Analytics', 'Research', 'Requirements Gathering',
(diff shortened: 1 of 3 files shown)


Lane012:
Not exactly sure how to correct failing checks. This is my first ever pull request so I apologise. If this provider seems useful I would like any insight to get the checks passing

gamesbook:
Would `hard_skills` not be better termed "IT Skills"?  Then this set could be expanded in future with, for example, "Laboratory Skills" or "Engineering Skills".

Lane012:
Hard_skills are not just IT skills. Tge types are to contrast "hard" and "soft". We could definitely do specific sets of skills like "Auto skills" ir "Labratory skills" ect... this is just a prototype I am waiting for the host to approve of. I ultimately would like to implement a resume provider, I see good usage with it.

gamesbook:
I agree that hard skills are not just IT skills - but the ones you have listed are; hence my suggestion to rename that group.  

I see room for a two-level skill set - "lower" level being IT, Lab, Engineering etc, - and "higher" level being hard skills.

Lane012:
The ones I have listed? I'm confused, if you are referring to the entire hard_skills set then no they are not all IT skills exclusively: "blogging", "statistical anaylisis", "Task management", ect....so it would be remisswd to rename that set as IT skills, now of course i could pull out the ones that are specifically IT in nature as their own set, but once again this is a prototype that will later be extended and refactored; just need to get host approval of the POC.

gamesbook:
Blogging is an IT Skill; "Stats Analysis" is often done by people with IT skills (although it can be classed as a math or science skill) - but its your patch; you can call these whatever you want.

fcurella:
Hi @Lane012 , thanks for your work!

I think a skill or resume provider is not a case common enough to be included in Faker, but you could make it an external provider and I'll be happy to include it in https://github.com/joke2k/faker/blob/master/docs/communityproviders.rst

Lane012:
Blogging is definitely not an IT skill. Also Stat analysis is used in any field that requires it accounting, mathematics, any study done that needs it, ect. 

Semantic debate aside I am going to develop this SkillProvider as a community provider and also do a ResumeProvider. You are more than welcome to help me develop it if you wish.

Lane012:
@fcurella I will definitely build it. Do I just notify you when it's finished?

fcurella:
@Lane012 Just send me a PR adding your provider to that doc and I'll merge it.

PR #731, opened 2018-03-23
Add car license plate format for de_DE

gsilvan:
`fake.license_plate()`

returns now a valid license plate format for **de_DE**

### DIFF
--- a/faker/providers/automotive/de_DE/__init__.py
+++ b/faker/providers/automotive/de_DE/__init__.py
@@ -3,14 +3,44 @@
 
 from __future__ import unicode_literals
 from .. import Provider as AutomotiveProvider
+import string
 
 
 class Provider(AutomotiveProvider):
-    license_formats = (
-        '???-??-%@@@',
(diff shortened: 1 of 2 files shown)


faker/providers/automotive/de_DE/__init__.py
```diff
@@ -0,0 +1,16 @@
+# coding=utf-8
+
+
+from __future__ import unicode_literals
+from .. import Provider as AutomotiveProvider
+
+
+class Provider(AutomotiveProvider):
+    license_formats = (
```
fcurella: according to https://en.wikipedia.org/wiki/Vehicle_registration_plates_of_Germany, this would generate invalid license plates. Can you write an implementation that's closer to the specification?

gsilvan: @fcurella yes you're right. I changed the implementation. It generates now licence plates in the valid format with allowed prefixes.

fcurella:
Thank you! ✨

PR #732, opened 2018-03-28
don't exclude edgecases from Finnish SSN by default

sanga:
### What does this changes

Do not exclude the "edgecases" for the Finnish SSN provider by default. Where "edgecases" means people born after 2000 (who's SSN have a slightly different format in Finland)

### What was wrong

Presumably this library is used mostly for testing. Currently, by default the Finnish SSN provider was missing the was not generating precisely the class of SSNs that are most likely to cause problems for SSN parsers.

### How this fixes it

Includes SSN of people born after 2000

Fixes #...

### DIFF
--- a/faker/providers/ssn/fi_FI/__init__.py
+++ b/faker/providers/ssn/fi_FI/__init__.py
@@ -7,11 +7,10 @@ import datetime
 
 class Provider(SsnProvider):
 
-    def ssn(self, min_age=18, max_age=90):
+    def ssn(self, min_age=0, max_age=105):
         """
         Returns 11 character Finnish personal identity code (Henkilötunnus,
-        HETU, Swedish: Personbeteckning). Age of person is between 18 and 90
-        years, based on local computer date. This function assigns random
+        HETU, Swedish: Personbeteckning). This function assigns random
         sex to person.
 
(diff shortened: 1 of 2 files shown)


faker/providers/ssn/fi_FI/__init__.py
```diff
@@ -7,7 +7,7 @@
 
 class Provider(SsnProvider):
 
-    def ssn(self, min_age=18, max_age=90):
+    def ssn(self, min_age=0, max_age=105):
         """
         Returns 11 character Finnish personal identity code (Henkilötunnus,
         HETU, Swedish: Personbeteckning). Age of person is between 18 and 90
```
fcurella: this docstring will need to be updated to say `Age of person is between 18 and 90 years by default`

sanga: :+1: good catch. Cheers. Fix incoming

sanga:
I just removed that text from the docstring. To me it doesn't make the same info repeated in 2 places and the func signature should be good enough as documentation.

I also noticed a bug in the code, so did a lite refactor, fixed then bug and added some tests.

fcurella:
Thank you! ✨

PR #733, opened 2018-04-05
Update requirements

confirmationbias616:
### What does this changes

Adds text_unidecode, six, and dateutil to the requirements.txt file

### What was wrong

couldn't run faker in development mode

### How this fixes it

self-explanatory


fcurella:
Thanks @confirmationbias616 ,

Could you use the same version numbers as in the `setup.py` file?

PR #738, opened 2018-04-06
Added Unix device & partition

ZuluPro:
### What does this changes

I added:
- File.unix_device : To create Unix device path. ie: /dev/sda, /dev/xvde, etc
- File.unix_partition: To create Unix partition path. ie: /dev/vdc1, /dev/sdb1, etc

With tests, of course ;)


fcurella:
Thank you! ✨

PR #739, opened 2018-04-06
Added random lower and uppercase letters

ZuluPro:
### What does this changes

Added method to get lowercase and uppercase ASCII letters.

### What was wrong

In my previous PR (#738), I wanted to have a random ASCII letter,
But have to import string or make a small trick.
Here's all is made is the base provider.


fcurella:
good idea! Thank you!

PR #740, opened 2018-04-11
Minor Refactoring for cli.py

confirmationbias616:
### What does this changes
Use underscore for "throwaway" variable in for loop instead of variable `i`, which never gets used.

### What was wrong
Readability / simplicity

Update (Apr 11, 2018): reversed previous change made in a commit, which tried to simplify the code but ended up breaking it.


fcurella:
thank you!

PR #741, opened 2018-04-12
Clarify prolog for cli help option

confirmationbias616:
This pull request revises the argparse prolog which gets displayed when the user calls the --help argument in CLI. 

The existing prolog was somewhat misleading for the user. It didn't mention that, in the absence of specifying a locale as argument, the user's OS' locale will be used by default. It said that Faker's standard default (en_US) would be used instead.

This revised prolog now shows the user what their default locale will be. It also clarifies that en_US is the ultimate fallback default. Also, the prolog now uses string formatting to print the DEFAULT_LOCALE, instead of hardcoding 'en_US' in the text - in case that ever changes!

I realize these changes could seem pointless or trivial to en_US users but it's an obvious upgrade when you're a user from a different locale.


fcurella:
Thank you so much! This is great ✨

PR #745, opened 2018-04-22
Add logging functionality

confirmationbias616:
This pull request adds logging functionality to the faker factory and CLI.

In the past, users of faker could have been misled into thinking they were using a specified locale when in fact that was not the case. The locale they were calling could have been unavailable with the provider they were using.

Through logging, this pull request attempts to add transparency so that the user is fully aware of any locale default switches that are happening in the background.

**Implementation**
Running as a module:
- INFO log events are created when provider has been localized with the intended locale
- WARNING log events are created when intended locale is not available for a provider so the default locale is used instead

In CLI :
 - when invoked, optional argument -v / --verbose will output the above-listed log info to the command line


faker/factory.py
```diff
@@ -6,11 +6,24 @@
...
+import sys
+
 from faker import Generator
 from faker.config import DEFAULT_LOCALE, PROVIDERS, AVAILABLE_LOCALES
 from faker.utils.loading import list_module
 
 
+logger = logging.getLogger(__name__)
+
+# identify if python is being run in interactive mode. If so, dsiable logging.
```
fcurella: small typo: `dsiable`

confirmationbias616:
Thanks for your suggestions - I've implemented all of them. I even went further and made a few more modifications:

1. Streamlined the quoting style throughout factory.py (probably should be done throughout entire codebase as well).
2. Added INFO logging event for when the provider does not feature localization.

fcurella:
Could you add some test for the verbosity level?

Check https://github.com/joke2k/faker/blob/master/tests/test_factory.py#L104 for an example of test for the CLI

confirmationbias616:
Sorry for all the commits - I'm still very new at this whole GitHub Flow thing. 

And also, thanks for pushing me to add some tests for my original commit. I was intimidated at first because I had never dealt with test automation or unit tests before. This has introduced me to a whole new facet of software engineering! 😁

fcurella:
@confirmationbias616 do you mind explaining why did you had seeding to the CLI?

confirmationbias616:
I did it for testing verbosity. 

In the unit test, I use the CLI to call the same exact same faker function twice - once without verbosity and once with verbosity. Then, the test will pass only if verbosity actually produced something more than just the faker result. I realize this might not be the best way to test it but I thought of doing it this way to make sure that any changes to the provider and/or logging functionality would have as little of a chance as possible to fail this test. 

Outside of just being useful for testing verbosity, I think having this extra CLI feature available to the user is kind of nice, especially when paired with repeat, wouldn’t you agree?

Hope this all makes sense!

fcurella:
@confirmationbias616 No, I'd rather merge it at once.

Thank you very much for your work! ✨

PR #746, opened 2018-04-24
Improve Korean person provider.

unace:
Improve Korean person provider.


unace:
@fcurella , I just realized author name is wrong, I will fix it and create a new pull-request again. sorry about that.

PR #752, opened 2018-05-11
Update README.rst

Okroshiashvili:
Adds Georgian locale provider link  as list does not included Georgian provider link


fcurella:
Thank you!

PR #756, opened 2018-05-28
Update hu_HU jobs to be unicode literals

proteusvacuum:
### What does this changes

Makes hu_HU locale jobs unicode

### What was wrong

When fetching jobs with the hu_HU locale, I was getting UnicodeDecodeError since these were not encoded correctly.

### How this fixes it

Correctly encodes these strings as unicode


fcurella:
Thank you!

PR #764, opened 2018-06-08
Fix Canadian SIN generation and add tests

### LINKED ISSUE #763: Canadian SSN Provider returns a string nothing like a Canadian SIN under python3
Canadian SSN Provider returns a string nothing like a Canadian SIN under python3. There are also no tests for this provider -- if there were, the change would have been detected.

### Steps to reproduce

```
Python 3.6.5 (default, Apr 25 2018, 14:23:58)
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)] on darwin
>>> from faker import Faker
>>> print(Faker('en-Ca').ssn())
37.09 8.002.0 83.08
```
### Expected behavior

An expected result follows the pattern `### ### ###`.  

### Actual behavior

Something else entirely. Here, it's `37.09 8.002.0 83.08`.

### Explanation

This is due to the integer division operator change between python2 and python3.

crd:
### What does this change

* Fixes Canada SSN Provider
* Implements tests
* Clarified comments
* Corrects a few subtle defects (previously-generated SINs had trailing spaces, two off-by-one errors)

### What was wrong

Between python2 and python3 the integer division operator changed from `/` to `//`. There were no tests implemented for this provider to detect the introduction of the defect.

### How this fixes it

Implemented and tested the appropriate fix.

Fixes #763

### DIFF
--- a/faker/providers/ssn/en_CA/__init__.py
+++ b/faker/providers/ssn/en_CA/__init__.py
@@ -2,36 +2,81 @@
 from __future__ import unicode_literals
 from .. import Provider as SsnProvider
 
+def checksum(sin):
+    """
+    Determine validity of a Canadian Social Insurance Number.
+    Validation is performed using a modified Luhn Algorithm.  To check
+    the Every second digit of the SIN is doubled and the result is
+    summed.  If the result is a multiple of ten, the Social Insurance
+    Number is considered valid.
+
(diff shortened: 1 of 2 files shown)


fcurella:
Nice work! Thank you so much ✨

PR #768, opened 2018-06-10
Implement support for start_datetime in unix_time()

### LINKED ISSUE #765: Allow the user to specify a start_datetime in addition to end_datetime in unix_time()
In addition to the `end_datetime` parameter added via PR #606 (nice work @kungfu71186) I see value in being able to generate random datetimes within a specific window -- for example, when testing behavior in response to ten-digit Unix epoch datetimes they would want to specify a start_date of `Sunday, September 9, 2001 1:46:40 AM GMT` corresponding to `1000000000`

crd:
### What does this change

Implements support for start_datetime in unix_time() and creates tests for this function.

### What was wrong

unix_time() supported the ability to explicitly declare an end_datetime parameter (to aid in deterministic testing) but didn't support start_datetime. This is necessary if there's a need to generate unix epoch values within certain ranges.

### How this fixes it

* Implement tests for all permutations of unix_time() parameters
* Defaults to previous behavior if start_datetime is omitted
* Make TestDateTime unit tests deterministic

Closes #765

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -1172,13 +1172,15 @@ class Provider(BaseProvider):
 
     regex = re.compile(timedelta_pattern)
 
-    def unix_time(self, end_datetime=None):
+    def unix_time(self, start_datetime=None, end_datetime=None):
         """
-        Get a timestamp between January 1, 1970 and now
+        Get a timestamp between January 1, 1970 and now, unless passed
+        explicit start_datetime and end_datetimes
         :example 1061306726
         """
+        start_datetime = self._parse_start_datetime(start_datetime)
         end_datetime = self._parse_end_datetime(end_datetime)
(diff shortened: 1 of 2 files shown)


faker/providers/date_time/__init__.py
```diff
@@ -1172,13 +1172,15 @@ class Provider(BaseProvider):
 
     regex = re.compile(timedelta_pattern)
 
-    def unix_time(self, end_datetime=None):
+    def unix_time(self, start_datetime=None, end_datetime=None):
```
fcurella: I'm a little worried about backward compatibility, in case people have started to use `unix_time` with positional argument.

I would change the signature to be `unix_time(self, end_datetime=None, start_datetime=None)`.

crd:
Agree with your feedback. Will close this PR and resubmit later, I want to get authorization to work on this as part of my day job.

crd:
Resubmitted via #770

PR #769, opened 2018-06-14
add China mainland job provider

ramwin:
### What does this changes

Add a job provider for zh_cn language

### What was wrong

There is not job provider in zh_cn language

### How this fixes it

I just translate the zh_TW job provider to zh_cn language


fcurella:
Is there any official source of data for `zh_CN` that we can use, instead of translating from `zh_TW`?

faker/providers/job/zh_CN/__init__.py
```diff
@@ -0,0 +1,483 @@
+# coding=utf-8
+from __future__ import unicode_literals
+from .. import Provider as BaseProvider
+
+
+# Taiwan jobs from
+# http://www.104.com.tw/public/function01/utf8/jsonJobCat.js
+# on Fri May 20 23:13:46 CST 2016
```
fcurella: Is there any official source of data for zh_CN that we can use, instead of translating from zh_TW?

ramwin:
@fcurella I changed the source to [51jobcdn's source](https://js.51jobcdn.com/in/js/2016/layer/funtype_array_c.js?20180611). [51job](https://www.51job.com/) is a large recruitment website in China. I think it is more reliable.

fcurella:
That will work. Thank you! :)

PR #772, opened 2018-06-15
Implement US ITIN and EIN tax identification numbers

### LINKED ISSUE #771: Support United States ITIN and EIN tax identification numbers
While United States SSNs are used by naturalized aliens and citizens, but businesses are identified by EINs and IRS issues ITINs to foreign nationals and others who have federal tax reporting or filing requirements and do not qualify for SSNs. 

Each type of tax identification number follows specific rules.

crd:
* Correctly generate ITIN and EIN tax identification numbers
* Implement tests
* Apply PEP8

Closes #771

### DIFF
--- a/faker/providers/ssn/en_US/__init__.py
+++ b/faker/providers/ssn/en_US/__init__.py
@@ -6,9 +6,9 @@ from .. import Provider as BaseProvider
 
 
 class Provider(BaseProvider):
-    SSN_TYPE = 'ssn'
-    ITIN_TYPE = 'itin'
-    EIN_TYPE = 'ein'
+    SSN_TYPE = 'SSN'
+    ITIN_TYPE = 'ITIN'
+    EIN_TYPE = 'EIN'
 
     def itin(self):
         """Generate a random United States Individual Taxpayer Identification Number (ITIN).
@@ -145,27 +145,25 @@ class Provider(BaseProvider):
         If no type is specified, a US SSN is returned.
         """
 
-        if taxpayer_identification_number_type.lower() == self.ITIN_TYPE:
+        if taxpayer_identification_number_type == self.ITIN_TYPE:
             return self.itin()
-        elif taxpayer_identification_number_type.lower() == self.EIN_TYPE:
+        elif taxpayer_identification_number_type == self.EIN_TYPE:
             return self.ein()
-        elif taxpayer_identification_number_type.lower() == self.SSN_TYPE:
+        elif taxpayer_identification_number_type == self.SSN_TYPE:
 
-            if taxpayer_identification_number_type.lower() == 'ssn':
+            # Certain numbers are invalid for United States Social Security
+            # Numbers. The area (first 3 digits) cannot be 666 or 900-999.
+            # The group number (middle digits) cannot be 00. The serial
+            # (last 4 digits) cannot be 0000.
+        
+            area = self.random_int(min=1, max=899)
+            if area == 666:
+                area += 1
+            group = self.random_int(1, 99)
+            serial = self.random_int(1, 9999)
 
-                # Certain numbers are invalid for United States Social Security
-                # Numbers. The area (first 3 digits) cannot be 666 or 900-999.
-                # The group number (middle digits) cannot be 00. The serial
-                # (last 4 digits) cannot be 0000.
-            
-                area = self.random_int(min=1, max=899)
-                if area == 666:
-                    area += 1
-                group = self.random_int(1, 99)
-                serial = self.random_int(1, 9999)
-
-                ssn = "{0:03d}-{1:02d}-{2:04d}".format(area, group, serial)
-                return ssn
+            ssn = "{0:03d}-{1:02d}-{2:04d}".format(area, group, serial)
+            return ssn
     
         else:
-            raise ValueError("taxpayer_identification_number_type must be one of 'ssn', 'ein', or 'itin'.")
+            raise ValueError("taxpayer_identification_number_type must be one of 'SSN', 'EIN', or 'ITIN'.")
--- a/faker/providers/ssn/en_US/__init__.py
+++ b/faker/providers/ssn/en_US/__init__.py
@@ -1,19 +1,153 @@
 # coding=utf-8
 from __future__ import unicode_literals
+
+import random
 from .. import Provider as BaseProvider
 
 
 class Provider(BaseProvider):
 
-    def ssn(self):
-        # Certain numbers are invalid for U.S. SSNs. The area (first 3 digits)
-        # cannot be 666 or 900-999. The group number (middle digits) cannot be
-        # 00. The serial (last 4 digits) cannot be 0000
-        area = self.random_int(min=1, max=899)
-        if area == 666:
-            area += 1
-        group = self.random_int(1, 99)
-        serial = self.random_int(1, 9999)
+    def ssn(self, taxpayer_identification_number='ssn'):
 
-        ssn = "{0:03d}-{1:02d}-{2:04d}".format(area, group, serial)
+        # SSN
+        # Certain numbers are invalid for United States Social Security
+        # Numbers. The area (first 3 digits) cannot be 666 or 900-999.
+        # The group number (middle digits) cannot be 00. The serial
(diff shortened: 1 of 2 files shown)


crd:
You sure? We seem to have overloaded `.ssn()` to represent any tax identifier. Consider `Faker('en-CA').ssn()` for example, which is really a Canadian Social Insurance Number. A US ITIN is equivalent to an SSN but are assigned to individuals who are not eligible for SSNs. Same for EINs, they are SSNs except assigned to legal entities rather than human beings.

If you're positive I don't mind refactoring but I wanted to double-check.

fcurella:
we could have the separated methods + the `identification_number_type` argument in `.ssn()`:
```python
class Provider(BaseProvider):
    SSN_TYPE = 'ssn'
    ITIN_TYPE = 'itin'
    EIN_TYPE = 'ein'

    def itin(self):
        # .....

    def ein(self):
        # .....

    def ssn(self, identification_number_type=SSN_TYPE):
        if identification_number_type == self.ITIN_TYPE:
            return self.itin()
        elif identification_number_type == self.EIN_TYPE:
            return self.ein()

       # original code...
```

crd:
Sounds good. I'll make the revisions.

crd:
@fcurella this is ready for another look -- let me know if you see anything you'd like me to tweak

faker/providers/ssn/en_US/__init__.py
```diff
@@ -1,19 +1,172 @@
...
-        if area == 666:
-            area += 1
-        group = self.random_int(1, 99)
-        serial = self.random_int(1, 9999)
+        # The group number must be between 70 and 99 inclusively but not 89 or 93
+        group = random.choice([x for x in range(70,100) if x not in [89, 93]])
 
         ssn = "{0:03d}-{1:02d}-{2:04d}".format(area, group, serial)
-        return ssn
+        return(ssn)
```
fcurella: why are these parenthesis here?

faker/providers/ssn/en_US/__init__.py
```diff
@@ -1,19 +1,172 @@
...
-        # Certain numbers are invalid for U.S. SSNs. The area (first 3 digits)
-        # cannot be 666 or 900-999. The group number (middle digits) cannot be
-        # 00. The serial (last 4 digits) cannot be 0000
-        area = self.random_int(min=1, max=899)
-        if area == 666:
-            area += 1
-        group = self.random_int(1, 99)
-        serial = self.random_int(1, 9999)
+        # The group number must be between 70 and 99 inclusively but not 89 or 93
+        group = random.choice([x for x in range(70,100) if x not in [89, 93]])
```
fcurella: What's the reason for changing the SSN generation rules? Were the sets of `ssn` and `itin` numbers overlapping?

faker/providers/ssn/en_US/__init__.py
```diff
@@ -1,19 +1,172 @@
...
+            '93',
+            '94',
+            '95',
+            '98',
+            '99']
+
+        ein_prefix = random.choice(ein_prefix_choices)
+        sequence = self.random_int(min=0, max=9999999)
+
+        ssn = "{0:s}-{1:07d}".format(ein_prefix, sequence)
```
fcurella: this is not an SSN, but an EIN. So this variable should be called `ein`

faker/providers/ssn/en_US/__init__.py
```diff
@@ -1,19 +1,172 @@
...
+        return(ssn)
```
fcurella: no parenthesis please :)

faker/providers/ssn/en_US/__init__.py
```diff
@@ -1,19 +1,172 @@
...
+                # (last 4 digits) cannot be 0000.
+            
+                area = self.random_int(min=1, max=899)
+                if area == 666:
+                    area += 1
+                group = self.random_int(1, 99)
+                serial = self.random_int(1, 9999)
+
+                ssn = "{0:03d}-{1:02d}-{2:04d}".format(area, group, serial)
+                return(ssn)
```
fcurella: parenthesis here, too

faker/providers/ssn/en_US/__init__.py
```diff
@@ -1,19 +1,172 @@
...
+    
+        else:
+            raise ValueError
```
fcurella: can you add a descriptive message to the `ValueError`? something like `raise ValueError("`taxpayer_identification_number_type` must be 'ssn', 'ein' or '' itin'.")`

tests/providers/test_ssn.py
```diff
@@ -143,7 +339,10 @@ def test_pt_BR_ssn(self):
 
     def test_pt_BR_cpf(self):
         for _ in range(100):
-            self.assertTrue(re.search(r'\d{3}\.\d{3}\.\d{3}\-\d{2}', self.factory.cpf()))
```
fcurella: I think your editor or some of your tools reformatted this. Can you revert these whitespace changes? (the ones above re: `ssn` are okay)

faker/providers/ssn/en_US/__init__.py
crd: Right, SSNs and ITINs have different generation/validation rules -- the previous approach didn't generate either consistently: It generated a SSN/ITIN like value that would fail validation checks.

faker/providers/ssn/en_US/__init__.py
crd: Good catch.

faker/providers/ssn/en_US/__init__.py
crd: You bet.

tests/providers/test_ssn.py
crd: Absolutely

faker/providers/ssn/en_US/__init__.py
```diff
@@ -1,19 +1,171 @@
...
-        # 00. The serial (last 4 digits) cannot be 0000
-        area = self.random_int(min=1, max=899)
-        if area == 666:
-            area += 1
-        group = self.random_int(1, 99)
-        serial = self.random_int(1, 9999)
+        # The group number must be between 70 and 99 inclusively but not 89 or 93
+        group = random.choice([x for x in range(70,100) if x not in [89, 93]])
 
         ssn = "{0:03d}-{1:02d}-{2:04d}".format(area, group, serial)
```
fcurella: this variable should now be called `itin`

crd: Of course it should

tests/providers/test_ssn.py
crd: I didn't forget about fixing these whitespace changes -- they're coming shortly

faker/providers/ssn/en_US/__init__.py
```diff
@@ -1,19 +1,171 @@
...
+        ein = "{0:s}-{1:07d}".format(ein_prefix, sequence)
+        return ein
+
+    def ssn(self, taxpayer_identification_number_type=SSN_TYPE):
+        """ Generate a random United States Taxpayer Identification Number of the specified type.
+
+        If no type is specified, a US SSN is returned.
+        """
+
+        if taxpayer_identification_number_type.lower() == self.ITIN_TYPE:
```
fcurella: let's skip the `lower()`s and make the argument case-sensitive, so that the exception message is accurate.

crd: I'm not convinced. Since SSN, ITIN, and EIN are initialisms/acronyms, it would be more correct to require these to be capitalized and not all lowercase (`taxpayer_identification_number_type in ['SSN', 'EIN', 'ITIN']`) but it seems unnecessarily strict to make the arguments case sensitive if there's no ambiguity in what the user meant.

Are you flexible on this one?

fcurella: we can make them uppercase, but I still want them to be case-sensitive, so the error message is 100% accurate. If it says "taxpayer_identification_number_type must be one of 'SSN', 'EIN', or 'ITIN'", then we must only accept those exact values. I'm ok with being overly strict if it means removing ambiguity.

PR #775, opened 2018-06-19
Add a date of birth provider

### LINKED ISSUE #774: Give the user the ability to generate adult vs. minor dates of birth
Presently, dates of birth in the Profile provider are chosen randomly between unix epoch timestamp 0 and the present date and time. This isn't particularly useful from a testing perspective, as `0 <= age <= 48 years` isn't the most useful range.

### Expected behavior

Users should be able to explicitly generate fake dates of birth for adults or minors.

crd:
### What does this change

Implements a `date_of_birth()` function as part of the `date_time` provider. Minimum and maximum ages can be specified as optional parameters.

This PR also updates the profile provider.

### What was wrong

The profile provider previously generated dates of birth in the range of unix epoch timestamp 0 to the present date. This was inflexible and put an arbitrarily low cap on the age of the generated person of ~48 years.

Closes #774

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -1203,15 +1203,24 @@ class Provider(BaseProvider):
         return datetime(1970, 1, 1, tzinfo=tzinfo) + \
             timedelta(seconds=self.unix_time(end_datetime=end_datetime))
 
-    def date_time_ad(self, tzinfo=None, end_datetime=None):
+    def date_time_ad(self, tzinfo=None, end_datetime=None, start_datetime=None):
         """
         Get a datetime object for a date between January 1, 001 and now
         :param tzinfo: timezone, instance of datetime.tzinfo subclass
         :example DateTime('1265-03-22 21:15:52')
         :return datetime
         """
+
+        # 1970-01-01 00:00:00 UTC minus 62135596800 seconds is
+        # 0001-01-01 00:00:00 UTC.  Since _parse_end_datetime() is used
+        # elsewhere where a default value of 0 is expected, we can't
+        # simply change that class method to use this magic number as a
+        # default value when None is provided.
+
+        start_time = -62135596800 if start_datetime is None else self._parse_start_datetime(start_datetime)
         end_datetime = self._parse_end_datetime(end_datetime)
-        ts = self.generator.random.randint(-62135596800, end_datetime)
+
+        ts = self.generator.random.randint(start_time, end_datetime)
         # NOTE: using datetime.fromtimestamp(ts) directly will raise
         #       a "ValueError: timestamp out of range for platform time_t"
         #       on some platforms due to system C functions;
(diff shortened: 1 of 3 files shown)


faker/providers/date_of_birth/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+
+from __future__ import unicode_literals
+
+from .. import BaseProvider
+
+
+localized = True
+
+
+class Provider(BaseProvider):
```
fcurella: I would move this provider to `person/__init__.py` instead of being its own.

faker/providers/date_of_birth/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+    # to human lifespan of about 115 years old. This seems like a
+    # reasonable max value for our purposes.
+    # 
+    # https://www.nature.com/news/human-age-limit-claim-sparks-debate-1.20750
+
+    age_range = {}
+    age_range['minor'] = {'minimum': 0, 'maximum': 18}
+    age_range['adult'] = {'minimum': 18, 'maximum': 115}
+
+    def date_of_birth(self, stage_of_life='adult'):
```
fcurella: Instead of `stage_of_life`, I'd rather use numbers, like a `min_age` and a `max_age` parameters.

crd:
Any thoughts as to why the specific unit test passes, but the full `coverage run` execution fails?

```
$ python -m unittest tests/providers/test_person.py
...............
----------------------------------------------------------------------
Ran 15 tests in 0.347s

OK
```

Versus:

```
$ coverage run --source=faker setup.py test
...
======================================================================
ERROR: test_documentor (tests.test_factory.FactoryTestCase)
----------------------------------------------------------------------
...
File "/Users/hwv729/workspaces/contrib/faker/faker/providers/person/__init__.py", line 134, in date_of_birth
    return self.generator.date_between(start_date=start_date, end_date=end_date)
AttributeError: 'Generator' object has no attribute 'date_between'

----------------------------------------------------------------------
Ran 246 tests in 15.029s

FAILED (errors=1)
Test failed: <unittest.runner.TextTestResult run=246 errors=1 failures=0>
error: Test failed: <unittest.runner.TextTestResult run=246 errors=1 failures=0>
```

fcurella:
@crd I'm suprised it works on the unit test. The `faker.generator.Generator` class does not have that method.

I think I workaround would be move the `date_of_birth` method from the `person` provider to the `date_time` provider. It might be a little weird, but it wouldn't be totally wrong for it to be there. We're generating a `date` object after all.

crd:
Please don't merge this quite yet -- I'm wrestling with Windows limitations. I'll let you know when I think it's in a good state.

crd:
This is ready for your review, @fcurella -- it took a surprising amount of fiddling to get this to work due to poor Windows support of dates prior to 1970-01-01.

faker/providers/date_time/__init__.py
```diff
@@ -1203,15 +1203,20 @@ def date_time(self, tzinfo=None, end_datetime=None):
...
+    def date_time_ad(self, tzinfo=None, end_datetime=None, start_datetime=None):
         """
         Get a datetime object for a date between January 1, 001 and now
         :param tzinfo: timezone, instance of datetime.tzinfo subclass
         :example DateTime('1265-03-22 21:15:52')
         :return datetime
         """
+
+        # -62135596800 is roughly the time delta in seconds between 1970-01-01 and 0001-01-01
+
```
fcurella: could add explicitly say why we're using a number instead of `None`?

tests/providers/test_date_time.py
```diff
@@ -471,3 +471,42 @@ def test_ar_eg(self):
...
+
+    def test_bad_age_range(self):
+        with self.assertRaises(ValueError):
+            self.factory.date_of_birth(minimum_age=5, maximum_age=0)
+
+    def test_acceptable_age_range(self):
+        dob = self.factory.date_of_birth(minimum_age=0, maximum_age=5)
+        assert isinstance(dob, date)
+
+    def test_identical_age_range(self):
```
fcurella: can you add a test with `minimum_age=100`? just to see that it works when we go earlier than 1970

crd:
Added a new test to generate a date of birth for an age between 100 and 110 and a note better explaining the `-62135596800` magic number. Should be in a good spot for acceptance now. Thanks for the feedback, @fcurella.

tests/providers/test_date_time.py
```diff
@@ -471,3 +471,89 @@ def test_ar_eg(self):
...
+            days_since_five_years_ago = now - now.replace(year=now.year-5)
+            days_since_six_years_ago = now - now.replace(year=now.year-6)
+
+            dob = self.factory.date_of_birth(minimum_age=5, maximum_age=5)
+            days_since_dob = now - dob
+
+            assert isinstance(dob, date)
+            assert days_since_six_years_ago >= days_since_dob >= days_since_five_years_ago
+
+def test_distant_age_range(self):
```
fcurella: fix the indentation here

crd:
Dancing around poor Windows support for dates < 1970-01-01 has been an adventure. I came across a defect in the date_time provider that I'll document and address separately, but I think this PR is in a good spot. Thanks for the back-and-forth.

fcurella:
Thank you for al your hard work on making this work on Windows! ✨

PR #786, opened 2018-07-19
Change blood group to "O"

mohi7solanki:
Change blood group from "0" (zero) to "O"


fcurella:
Thank you! Nice catch! ✨

fcurella:
For future reference, some countries do call the O group 'zero', so that might need to be localized in the specific providers. By the default should be `O` (ie: 'oh')

PR #794, opened 2018-08-06
date_time ignores tzinfo

Fraterius:
### What does this changes

TZinfo was overwritten by utc which caused faker to return correct datetime object only for utc TZinfo whenever datetime_to_timestamp function was used. Now it looks ok for instance:

`tz = pytz.timezone('Europe/Warsaw')`
`fake = faker.Faker()`
`fake.future_datetime(end_date="+1h", tzinfo=tz)` 

### What was wrong

Mistake in code.

### How this fixes it

Simple bugfix.

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -1557,7 +1557,12 @@ class Provider(BaseProvider):
             ts = start_date + self.generator.random.random()
         else:
             ts = self.generator.random.randint(start_date, end_date)
-        return datetime(1970, 1, 1, tzinfo=tzinfo) + timedelta(seconds=ts)
+        if tzinfo is None:
+            return datetime(1970, 1, 1, tzinfo=tzinfo) + timedelta(seconds=ts)
+        else:
+            return (
+                datetime(1970, 1, 1, tzinfo=tzutc()) + timedelta(seconds=ts)
+            ).astimezone(tzinfo)
 
     def date_between(self, start_date='-30y', end_date='today'):
         """


Fraterius:
Ok I think that:

`dt = dt.astimezone(tzutc())`

should be removed from the code, not sure what is the purpose of this line but with that line faker returns wrong values, like future_datetime isn't future at all. If end_date is +30d or something big enough it can be overlooked.

I'm leaving pull request for further discuss on the subject.

Fraterius:
It also can be fixed by replacing:

`return datetime(1970, 1, 1, tzinfo=tzinfo) + timedelta(seconds=ts)`

with 

`return (datetime(1970, 1, 1, tzinfo=tzutc()) + timedelta(seconds=ts)).astimezone(tzinfo)`

in faker/faker/providers/date_time/__init__.py on [line](https://github.com/joke2k/faker/blob/e053a319779fd20573978087d13700d0133adf6c/faker/providers/date_time/__init__.py#L1560).

But I'm not sure whether the same issue occurs somewhere else in the code. Not sure why someone wanted to change the timezone in [datetime_to_timestamp](https://github.com/joke2k/faker/blob/e053a319779fd20573978087d13700d0133adf6c/faker/providers/date_time/__init__.py#L22) in the first place, but I'm sure there must be some very good reason for that (I'm afraid it might impact the code in some other place).

Fraterius:
I can also add that using [date_time_between_dates](https://github.com/joke2k/faker/blob/e053a319779fd20573978087d13700d0133adf6c/faker/providers/date_time/__init__.py#L1633) does work as expected with tzinfo.

fcurella:
@Fraterius thank you for looking into this.

The reason why we convert to `utc` in `datetime_to_timestamp` is that `calendar.timegm` assumes the input to be in UTC:

https://docs.python.org/2/library/calendar.html#calendar.timegm
https://docs.python.org/2/library/time.html#time.gmtime

The reason we use `calendar.timegm` and not `time.mktime` is that `time.mktime()` might use different epoch, depending on the platform:

> The epoch is the point where the time starts, and is platform dependent. For Unix, the epoch is January 1, 1970, 00:00:00 (UTC). To find out what the epoch is on a given platform, look at time.gmtime(0).

https://docs.python.org/3/library/time.html

While for `calendar`, it is hardcoded to the POSIX Epoch, 1970-01-01UTC.

Fraterius:
@fcurella thanks a lot for deep clarification on the subject. Not sure why it also works (and actually as expected) when I'm passing datetime object with different tzinfo than UTC, as You said it expects UTC tz. Probably there should be UTC rather for code consistency than a method requirement.

This way or another tzinfo passed in is dropped in date_time_between method and it needs to be fixed. Not sure if the issue occurs anywhere else in the code. From what I've read the source this is the only place which is used by many other methods. I can write a commit to fix this like I wrote above, or I can close this pull request and create an issue on the subject.

Just let me know and maybe give a little bit of assurance that I'm correct. Basically change of the line [1560](https://github.com/joke2k/faker/blob/e053a319779fd20573978087d13700d0133adf6c/faker/providers/date_time/__init__.py#L1560) with below code should do the trick:

`return (datetime(1970, 1, 1, tzinfo=tzutc()) + timedelta(seconds=ts)).astimezone(tzinfo)`

fcurella:
@Fraterius your solution makes a lot of sense. Looks good to me 👍

fcurella:
Thank you!

PR #797, opened 2018-08-10
Set tzinfo correctly when generating date-of-birth (ticket #796)

### LINKED ISSUE #796: test_identical_age_range: set tzinfo when generating date-of-birth
The test_identical_age_range unit-test can fail in 0.8.17, depending on the timezone and time-of-day.  This problem was introduced in https://github.com/joke2k/faker/commit/6c83a263a589c031ef68244fa81d21c03c620100 ; pull request is coming soon.

### Steps to reproduce

1. Wait until a time of day such that the date in UTC is one day ahead of your local date.  This naturally depends on your timezone being behind UTC.  I'm in US/Central timezone (a.k.a. GMT-5:00/GMT-6:00 depending on time of year) so waiting until after 7:00 PM is about right.
2. Run the unit tests (python setup.py test)

### Expected behavior

Unit tests should pass.

### Actual behavior

For me, test_identical_age_range reliably fails at least once out of the 100 repetitions, because it accidentally generates an exactly 6-year-old DoB.

Kelledin:
Without this patch applied, the test_identical_age_range unit-test can
fail at certain times of day on systems where local time is behind UTC.
This is because the test generates "now" based off UTC, but apparently
lets "dob" follow the local timezone.

The actual library funtionality appears to be sane; only the unit-tests
are affected.

### What does this changes

With this change, we set tzinfo=utc when generating date-of-birth.

### What was wrong

For timezones behind UTC, the date in UTC can be one day ahead of the localized date depending on time-of-day.  Failing to set a consistent timezone in this case allows the date-of-birth to be exactly six years old.

### How this fixes it

Timezone is set consistently when reading current time and when generating date-of-birth.

Fixes #796

### DIFF
--- a/tests/providers/test_date_time.py
+++ b/tests/providers/test_date_time.py
@@ -545,7 +545,7 @@ class DatesOfBirth(unittest.TestCase):
             days_since_five_years_ago = now - now.replace(year=now.year-5)
             days_since_six_years_ago = now - now.replace(year=now.year-6)
 
-            dob = self.factory.date_of_birth(minimum_age=5, maximum_age=5)
+            dob = self.factory.date_of_birth(tzinfo=utc, minimum_age=5, maximum_age=5)
             days_since_dob = now - dob
 
             assert isinstance(dob, date)


fcurella:
Thank you for fixing this! ✨

PR #793, opened 2018-08-03
Comma in Cardoso

Nichlas:
Correcting name that ruins my CSVs

### What does this changes

Removed a comma.

### What was wrong

The comma shouldn't have been there.

### How this fixes it

No comma.

PR #799, opened 2018-08-14
Missing misplaced commas

153957:
### What does this changes

Fix mistakes in the content of some providers.

### What was wrong

Inspired by https://github.com/joke2k/faker/pull/793 I did some quick searches in the code for similar mistakes.

Some elements which should be separate were joined together, sometimes a comma was mistakenly part of an element, and sometimes a trailing space was part of an element.

### How this fixes it

I added commas between elements, and removed characters which should not be part of elements.


fcurella:
Nice work! Thank you so much! ✨

PR #805, opened 2018-08-28
Refactor IPv4 address generation to use ipaddress module #803

### LINKED ISSUE #803: Public IP blocks used with ipv4_public encompass reserved address namespaces
The `ipv4_public` faker generates IP addresses from public blocks, excluding private blocks configured in `IPV4_PUBLIC_NETS` and `IPV4_PRIVATE_NET_BLOCKS`, however it often results in addresses coming from reserved spaces.

One example is 192.0.0.145, or in general 192.0.0.0/24 addresses from reserved space, per https://en.wikipedia.org/wiki/Reserved_IP_addresses

This clashes with many popular libraries checking whether IP is public and routable such as `django-ipware`.

### Steps to reproduce

```
>>> startswith = False
>>> while not startswith:
...   ip = fake.ipv4_public()
...   if ip.startswith('192.0.0'):
...     startswith = True
...     print(ip)
...
192.0.0.145
```

### Expected behavior

No address starting with `192.0.0` generated

### Actual behavior

`192.0.0.145` comes up often.

maticomp:
### What does this changes

I refactored generation of IPv4 addresses with huge help and input @kungfu71186 to use `ipaddress` module instead of doing internal math on ranges. Network constants have been provided based on `ipaddress._IPv4Constants` with private networks updated to reflect current IANA lists. The IP generation now follows the general pattern both for public and private addresses:

1) Choose the desired network class (a, b or c)
2a) For private networks - choose all private networks overlapping the class
2b) For public networks - exclude all private networks from the class
3) Exclude special networks that should not yield IPs from faker such as 0.0.0.0/8
4) Choose a random subnet from the lot, and then a random IP from the subnet

Tests have been updated to reflect the changes.

**Caveat for discussion:** private IP generation chooses from all private networks from IANA list, not only from `10.0.0.0`, `172.16.0.0` and `192.168.0.0` ranges. This may not be a desired effect and it's possible additional constant should be added for private IP generation to yield only IPs from general use private subnets.

### What was wrong

`faker` was using internal constants for private networks that was not up to date with IANA documents, yielding non-public addresses from public method. After digging in it appears that `ipaddress` constants are not up to date either.

### How this fixes it

The core issue is fixed by replacing the private network constants with an up to date list. Refactor for the generation method is a side effect.

Fixes #803

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -20,10 +20,8 @@ class _IPv4Constants:
     IPv4 network constants used to group networks into different categories.
     Structure derived from `ipaddress._IPv4Constants`.
 
-    Private network list is updated to comply with current IANA list.
-
-    Excluded networks are added to allow filtering out certain networks
-    from faker output.
+    Excluded network list is updated to comply with current IANA list of
+    private and reserved networks.
     """
     _network_classes = {
         'a': ip_network('0.0.0.0/1'),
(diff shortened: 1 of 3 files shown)


maticomp:
Ouch, I didn't see that coming. @kungfu71186 hold on please, I'll see why the backported `ipaddress` fails here.

maticomp:
Ouch, I didn't see that coming 2.0 — Python's 3.4 `ipaddress` lacks constants, they had a much simpler implementation. The backport is fine, but it uses the same module name, so it won't shadow the system one. *scratches head*

kungfu71186:
Lol, they didn't have constants in the 3.4 library either. I would suggest we forget what they have there. It's not really correct anyway. I would just say just hard code all subnets in there. I would even go on to suggest that private only include the 3 normal private networks and then just create another one called _other_private_networks or something and the remaining in there. Until these can really be defined properly, I don't think it's going to matter too much at this point anyway.


```
_linklocal_network = IPv4Network('169.254.0.0/16')

_loopback_network = IPv4Network('127.0.0.0/8')

_multicast_network = IPv4Network('224.0.0.0/4')

_private_networks = [
    IPv4Network('10.0.0.0/8'),
    IPv4Network('172.16.0.0/12'),
    IPv4Network('192.168.0.0/16')
]
_other_private_networks = [
    IPv4Network('100.64.0.0/10'),
    IPv4Network('192.0.0.0/24'),
    IPv4Network('192.0.0.0/29'),  
    so on...
]
```
What you think?

maticomp:
@kungfu71186 I think it's a sensible choice, and it will resolve the problem with 3.4's not having constants at the same time. I'll be able to update this PR tomorrow.

Regarding your last remark, I think we should generate private from just the three common subnets.

Generating from other private subnets is an edge case `faker` didn't support before. Now it's possible to generate private addresses using `ipv4_private` and `ipv4` with parameter `private=True`. Adding even more flexibility would either cloud the method signature of `ipv4` further, or make that functionality available through `ipv4_private` and not through `ipv4`. 

I'd consider making another issue for thinking over the these methods though.

kungfu71186:
@michaelcho I think we are saying the same thing. I'm all for just adding in the private subnets now as the main 3. My comment was about if you were to implement the other private networks.

I was thinking of something like this:
`def ipv4_private(self, network=False, address_class=None, include_other=false):`

But default it would be false. If true, then include those other ones. I think the expected behavior is what you are suggesting and that you would get an ip from one of those 3 main private classes if you wanted a private ip.

maticomp:
@kungfu71186 

> `def ipv4_private(self, network=False, address_class=None, include_other=false):`

My issue with that is that right now `ipv4` can proxy _all_ kind of generated IPs. If we extend just `ipv4_private` with new method parameter, it will no longer be the case. If we extend `ipv4` method too - it will become bulky.

But that aside, I wonder if there is a valid use case for generating IPs from the other ranges you mention. The term _private IP_ is not really well defined. Some of them are for documentation purposes, some for operator grade NAT, some are restricted for other purposes, their purpose does not overlap. Most of those other ranges should not really be used and I'm not sure there is a valid use case for generating them randomly from one big bucket (`include_others`). You'd sometimes randomly get a documentation IP, sometimes link local, and sometimes... a truly reserved one.

Meanwhile, after the last change we have all builds passing, at last 😄

fcurella:
Thank you all for your hard work on this! Seriously ✨s!

@maticomp What's the status on it? Is it ready to be merged?

maticomp:
@fcurella I believe it is. It's complete. If we'll need more/different features to improve this even further, I can make another PR with that - but this one is already self-contained :) Cheers!

fcurella:
Thank you all for your hard work! ✨

PR #808, opened 2018-09-08
Fix/master/make words return unique words

### LINKED ISSUE #791: Ensure words() returns a unique list of words without duplicates
When calling `words(10)` there is no guarantee that the returned list of elements is unique.   While I could do `set(words(10))`, this may only return 9 items when I am in need of 10.

If creating a list of words with potential duplicates is a useful feature to some, maybe adding an optional parameter called `unique` to the `words` generator could prove useful.

micahstrube:
### What does this change

Close #791. Add `unique` option to `words()` to get a list of unique words. Add new tests for `words()`.

### What was wrong

`words()` did not have an option for retrieving a list of unique words. See #791.

### How this fixes it

Adds `unique` as a boolean argument to `words()`. Default behavior for `words()` is non-unique. Additional unit tests are added for testing the behavior of `words()`.

### DIFF
--- a/faker/providers/lorem/__init__.py
+++ b/faker/providers/lorem/__init__.py
@@ -34,6 +34,7 @@ class Provider(BaseProvider):
         :param nb: how many words to return
         :param ext_word_list: a list of words you would like to have instead of
             'Lorem ipsum'
+        :param unique: If True, the returned word list will contain unique words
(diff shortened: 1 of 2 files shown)


fcurella:
Thank you! ✨

PR #809, opened 2018-09-09
Close #800. Allow territories to be excluded from `state_abbr()` for `en_US`.

### LINKED ISSUE #800: State Abbreviations for en_US have too many
The output of the state provider in address only outputs the 50 states, but the state_abbr has 59 potential outputs

### Steps to reproduce

Generate a value using the state_abbr provider

### Expected behavior

The value should be one of the 50 US states abbreviations

### Actual behavior

The value is one of the 50 US states, Washington DC (DC), American Samoa (AS), Micronesia (FM), Guam (GU), Marshall Islands (MH), Northern Marianas (MP), Palau (PW), Puerto Rico (PR), and Virgin Isles (VI). It appears the list was pulled from sources such as https://www.factmonster.com/us/postal-information/state-abbreviations-and-state-postal-codes that list every possible postal state abbreviation.

micahstrube:
### What does this changes

Adds `include_territories` parameter to `state_abbr()`, which allows including/excluding US territories from `state_abbr()` return. Default is True, to include territories.

### What was wrong

`state_abbr()` did not allow excluding US territories from returned state abbreviations. See #800.

### How this fixes it

Removed territories from the `Provider.address.en_US.states_abbr` list, and created new list, `Provider.address.en_US.territories_abbr` for only territories.

Updated the method `Provider.address.en_US.state_abbr()` to have new boolean parameter `include_territories=True`, allowing exclusion of territories if needed.

Updated unit tests to reflect the change in behavior for `state_abbr()`.

### DIFF
--- a/faker/providers/address/en_US/__init__.py
+++ b/faker/providers/address/en_US/__init__.py
@@ -283,6 +283,8 @@ class Provider(AddressProvider):
         'AS', 'FM', 'GU', 'MH', 'MP', 'PW', 'PR', 'VI',
     )
 
+    states_and_territories_abbr = states_abbr + territories_abbr
+
     military_state_abbr = ('AE', 'AA', 'AP')
 
     military_ship_prefix = ('USS', 'USNS', 'USNV', 'USCGC')
@@ -340,10 +342,9 @@ class Provider(AddressProvider):
         :param include_territories: If True, territories will be included.
             If False, only states will be returned.
         """
-        abbr_options = self.states_abbr
         if include_territories:
-            abbr_options += self.territories_abbr
-        return self.random_element(abbr_options)
+            self.random_element(self.states_and_territories_abbr)
+        return self.random_element(self.states_abbr)
 
     def postcode(self):
         return "%05d" % self.generator.random.randint(501, 99950)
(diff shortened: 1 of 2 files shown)


faker/providers/address/en_US/__init__.py
```diff
@@ -329,8 +333,17 @@ def secondary_address(self):
...
+    def state_abbr(self, include_territories=True):
+        """
+        :returns: A random state or territory abbreviation.
+
+        :param include_territories: If True, territories will be included.
+            If False, only states will be returned.
+        """
+        abbr_options = self.states_abbr
+        if include_territories:
+            abbr_options += self.territories_abbr
```
fcurella: This method could potentially be called thousands of times, and I'm worried about the performance of creating the list at every call. I would rather define another class constant like `states_with_territories_abbr = states_abbr + territories_abbr` and have
```python
if include_territories:
    return self.random_element(states_with_territories_abbr)
```

micahstrube: That makes sense. Thanks for reviewing my pull request! :) I will get a new commit in shortly to address that.

micahstrube:
Hey @fcurella , I made the changes you recommended on the branch. Thanks again for reviewing my PR.

fcurella:
Thank you, @micahstrube! ✨

PR #816, opened 2018-09-16
Use six compatibility shims; avoid reimplementing them

jdufresne:
The six library is already a dependency. The text_type, binary_type, and
StringIO shims exist in the six library. No need to reimplement them.


fcurella:
Thank you so much for taking the time to clean this up! ✨

PR #821, opened 2018-09-17
Simplify slugify; use six.text_type instead of force_text

jdufresne:
Django's `force_text()` is overkill for the purposes of Faker's `slugifly`. Simply need to coerce the input to Unicode text. Django's `force_text()` is designed to handle many more cases that don't apply to developer controlled data.


fcurella:
Makes sense. Thank you! ✨

PR #824, opened 2018-09-20
Unify test environments between Travis CI & tox.ini

jdufresne:
Now, the running tests locally uses a Python environment closer to
Travis. Allows more easily verifying tests across all supported
environments with a single command.


fcurella:
Nice Work! Thank you! ✨

PR #825, opened 2018-10-01
New image URLs

clarmso:
### What does this changes

Add more image URLs for faker.provider.internet.image_url()

### What was wrong

This PR is just an enhancement.  No functionality of image_url() has been changed.

### How this fixes it

Add a few more image URL templates that I have used in the past.  I have verified, as of today, that the URLs work.

I have also added a unit test.

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -115,13 +115,8 @@ class Provider(BaseProvider):
         '?txtsize=55&txt={width}x{height}&w={width}&h={height}',
         'https://www.lorempixel.com/{width}/{height}',
         'https://dummyimage.com/{width}x{height}',
-        'https://baconmockup.com/{width}/{height}',
-        'https://loremflickr.com/{width}/{height}',
         'https://placekitten.com/{width}/{height}',
-        'https://www.placecage.com/{width}/{height}',
         'https://placeimg.com/{width}/{height}/any',
-        'https://placebear.com/{width}/{height}',
-        'https://picsum.photos/{width}/{height}',
     )
 
     replacements = tuple()
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -116,14 +116,11 @@ class Provider(BaseProvider):
         'https://www.lorempixel.com/{width}/{height}',
         'https://dummyimage.com/{width}x{height}',
         'https://baconmockup.com/{width}/{height}',
-        'https://www.fillmurray.com/{width}/{height}',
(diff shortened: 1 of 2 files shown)


faker/providers/internet/__init__.py
```diff
@@ -115,6 +115,13 @@ class Provider(BaseProvider):
         '?txtsize=55&txt={width}x{height}&w={width}&h={height}',
         'https://www.lorempixel.com/{width}/{height}',
         'https://dummyimage.com/{width}x{height}',
+        'https://baconmockup.com/{width}/{height}',
```
fcurella: I am a lover of bacon myself, but I'm afraid the images returned by this service might be too meaty, and I'd like to avoid any issue related to content. Could you remove it?

faker/providers/internet/__init__.py
```diff
@@ -115,6 +115,13 @@ class Provider(BaseProvider):
...
+        'https://loremflickr.com/{width}/{height}',
```
fcurella: My understanding is that this service returns random images from flickr, without content curation. I'm not sure I'm feeling comfortable with having random content returned. I'd rather not have it.

faker/providers/internet/__init__.py
```diff
@@ -115,6 +115,13 @@ class Provider(BaseProvider):
...
+        'https://placekitten.com/{width}/{height}',
+        'https://www.placecage.com/{width}/{height}',
```
fcurella: I'm not sure the images on site have used according to their license. In fact, I don't see any mention of licensing or attribution anywhere on this website.

faker/providers/internet/__init__.py
```diff
@@ -115,6 +115,13 @@ class Provider(BaseProvider):
...
+        'https://placeimg.com/{width}/{height}/any',
+        'https://placebear.com/{width}/{height}',
```
fcurella: Same concern as for `placecage.com`

faker/providers/internet/__init__.py
clarmso: Placekitten's license: https://placekitten.com/attribution.html

placeimg's license is creativecommon.

faker/providers/internet/__init__.py
clarmso: I've removed most of the new URLs, just to be safe.

faker/providers/internet/__init__.py
clarmso: Sure. That's why we have code review. :)

fcurella:
Looks Great! Thank you! ✨

PR #826, opened 2018-10-01
#491 Add replacements for non-ascii characters in pt_BR.

### LINKED ISSUE #491: pt_BR email not returning valid email addresses
When creating a fake Factory with the pt_BR it is not returning valid email addresses.
Example:
```
melocauã@bol.com.br
joão-gabrielferreira@ig.com.br
lavíniarodrigues@sales.org
vitória78@example.br
```

clarmso:
### What does this changes

Add replacements tuples to specify the mapping between non-ascii characters to ascii characters.

### What was wrong

Some non-ascii characters shows up in ascii_* methods.

### How this fixes it

Added replacements tuples in `faker/providers/internet/pt_BR/__init__.py`.

Added some unit tests.

### Note

In addition, ``string = unidecode(string)`` from ``_to_ascii(str)`` should cover any characters that are missed in the replacements.  The replacements tuples basically force the character mappings that may be overlooked by unidecode.

Fixes #491

### DIFF
--- a/faker/providers/internet/pt_BR/__init__.py
+++ b/faker/providers/internet/pt_BR/__init__.py
@@ -13,3 +13,11 @@ class Provider(InternetProvider):
         'bol.com.br',
         'ig.com.br')
     tlds = ('com', 'com', 'com', 'net', 'org', 'br', 'br', 'br')
+    replacements = (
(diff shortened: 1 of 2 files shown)


fcurella:
Thank you! ✨

PR #827, opened 2018-10-02
Added usage example for providers

Noir-:
### What does this changes

A short example how to use a provider.

### What was wrong

This information was missing in the whole documentation.


fcurella:
Thank you! ✨

PR #828, opened 2018-10-03
Locale cs_CZ should differentiate between male and female names

### LINKED ISSUE #625: Locales for cs_CZ fake.name_male() shows a female name
Looking at the very last example [here](https://faker.readthedocs.io/en/latest/locales/cs_CZ.html#faker-providers-misc). The name is actually a female name.

```
fake.name_male()
# 'Ing. Sára Mašková CSc.'
```

clarmso:
### What does this changes

The cs_CZ locale should contain male and female name formats.  The calls to name_female() should not return male name and name_male() should not return female names.

### What was wrong

There was no formats_male and formats_female in the person's locale cs_CZ.

### How this fixes it

Add formats_male and format_female templates.

Fixes #625

### DIFF
--- a/faker/providers/person/cs_CZ/__init__.py
+++ b/faker/providers/person/cs_CZ/__init__.py
@@ -1,26 +1,26 @@
 # coding=utf-8
 from __future__ import unicode_literals
+from collections import OrderedDict
 from .. import Provider as PersonProvider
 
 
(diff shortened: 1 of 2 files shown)


fcurella:
Looks good! Thank you! ✨

PR #830, opened 2018-10-05
687 Invalid domain name in zh_CN locale

### LINKED ISSUE #687: zh_CN (and possible other providers) return invalid results for domain_name

### Steps to reproduce

f = Faker('zh_CN')
f.domain_name()
>he.cn

### Expected behavior

it should return valid domain names

### Actual behavior

he.cn is not a valid domain. he is in the list of second-level domains (https://en.wikipedia.org/wiki/.cn#Second-level_domains_of_provinces) and will cause failures in places where domains are being validated.

it's unclear where this change should/could be made after briefly reviewing the source code.

clarmso:
### What does this changes
This PR adds more patterns on domain_word.  From my personal observation, some domains are numbers that sound like a specific phrase. Some domains looks like random letters that are actually abbreviations of some sort.

The second level domains such as the ones denoting the provinces are added.

If the domain word is one of the second level domains, it cannot be used if the levels is set to 1.

### What was wrong
Domains such as he.cn can be generated.  'he' is a valid second level domain and is a legit last name.

### How this fixes it
* More combinations on the pattern for domain_word()
* Add second level domains.

Fixes #687

### DIFF
--- a/faker/providers/internet/zh_CN/__init__.py
+++ b/faker/providers/internet/zh_CN/__init__.py
@@ -1,5 +1,6 @@
 # coding=utf-8
 from __future__ import unicode_literals
+from collections import OrderedDict
 from .. import Provider as InternetProvider
 from faker.utils.decorators import slugify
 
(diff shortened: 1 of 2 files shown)


fcurella:
Thank you! ✨

PR #833, opened 2018-10-09
Add french (fr_FR) word list for lorem provider

tristandeborde:
### What does this changes

The lorem provider now supports the fr_FR locale.

### What was wrong

Lorem provider only returned latin sentences for the fr_FR locale :(

### How this fixes it

Use the `lorem/en_US` directory as a template. The word list comes from the National Education Ministry's website Eduscol, and was made by a lexicologist. More info can be found here: http://eduscol.education.fr/cid50486/liste-de-frequence-lexicale.html


fcurella:
Thank you! ✨

PR #835, opened 2018-10-11
Handle OSError: Closes #829

### LINKED ISSUE #829: OSError: [Errno 22] Invalid argument on Windows
In some cases random dates generated by faker fail on Windows.
There was another issue on this, but it was closed and a stack trace was asked. It was issue #460 
I have the stack trace included.
Unfortunately I don't have the exact line where it fails, because there are several in my code, and the stack trace doesn't say which statement is the direct cause, but here are all of the fake dates in my code:

```
dateofbirth = factory.Faker('date_between', start_date="-70y", end_date="-20y")
created = factory.Faker('date_time_this_decade', before_now=True, after_now=False)
dateLastModified = factory.Faker('date_time_this_year', before_now=True, after_now=False)
password_changed = factory.Faker('date_time_this_year', before_now=True, after_now=False)
status_date = factory.Faker('date_time_this_year', before_now=True, after_now=False)
```

### Steps to reproduce
This occurs randomly, propably when the randomly chosen date is out of specific bounds.
Please refer to #460 for further details.

### Expected behavior
A random date should be assigned

### Actual behavior
OSError: [Errno 22] Invalid argument

### stack trace

```
  File "C:\projects\clubassistant\features\current\ca5\tests\businesslogic\reminderlogic.py", line 20, in test_amount_too_high_ok
    member = MemberFactory()
  File "C:\projects\clubassistant\features\current\ca5\.venv\lib\site-packages\factory\base.py", line 46, in __call__
    return cls.create(**kwargs)
  File "C:\projects\clubassistant\features\current\ca5\.venv\lib\site-packages\factory\base.py", line 563, in create
    return cls._generate(enums.CREATE_STRATEGY, kwargs)
  File "C:\projects\clubassistant\features\current\ca5\.venv\lib\site-packages\factory\base.py", line 500, in _generate
    return step.build()
  File "C:\projects\clubassistant\features\current\ca5\.venv\lib\site-packages\factory\builder.py", line 272, in build
    step.resolve(pre)
  File "C:\projects\clubassistant\features\current\ca5\.venv\lib\site-packages\factory\builder.py", line 221, in resolve
    self.attributes[field_name] = getattr(self.stub, field_name)
  File "C:\projects\clubassistant\features\current\ca5\.venv\lib\site-packages\factory\builder.py", line 375, in __getattr__
    extra=context,
  File "C:\projects\clubassistant\features\current\ca5\.venv\lib\site-packages\factory\faker.py", line 56, in evaluate
    return self.generate(extra or {})
  File "C:\projects\clubassistant\features\current\ca5\.venv\lib\site-packages\factory\faker.py", line 53, in generate
    return subfaker.format(self.provider, **kwargs)
  File "C:\projects\clubassistant\features\current\ca5\.venv\lib\site-packages\faker\generator.py", line 76, in format
  File "C:\projects\clubassistant\features\current\ca5\.venv\lib\site-packages\faker\providers\date_time\__init__.py", line 1580, in date_between
    return self.date_between_dates(date_start=start_date, date_end=end_date)
  File "C:\projects\clubassistant\features\current\ca5\.venv\lib\site-packages\faker\providers\date_time\__init__.py", line 1680, in date_between_dates
    return self.date_time_between_dates(date_start, date_end).date()
  File "C:\projects\clubassistant\features\current\ca5\.venv\lib\site-packages\faker\providers\date_time\__init__.py", line 1665, in date_time_between_dates
    pick = datetime.fromtimestamp(timestamp, tzlocal())
OSError: [Errno 22] Invalid argument
```

BrodyUnderwood1:
Handles issue #829

### What does this changes

faker/providers/date_time/__init__.py     date_time_between_dates()

### What was wrong

Windows unable to handle certain dates outside of range

### How this fixes it

Returns new date instead of out of range date.

Fixes #829

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -1669,6 +1669,8 @@ class Provider(BaseProvider):
               pick = datetime.fromtimestamp(timestamp, tzinfo)
         except OSError:
           return self.date_time_between_dates(datetime_start, datetime_end)
+        except RecursionError as err:
+          raise err  
         return pick
 
     def date_between_dates(self, date_start=None, date_end=None):
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -1669,8 +1669,10 @@ class Provider(BaseProvider):
               pick = datetime.fromtimestamp(timestamp, tzinfo)
         except OSError:
           return self.date_time_between_dates(datetime_start, datetime_end)
-        except RecursionError as err:
-          raise err  
+        except RuntimeError as re:
+          if re.args[0] != 'maximum recursion depth exceeded':
+            raise('Recursion Error: datetime_start and datetime_end likely out of range')
+          raise re
         return pick
 
     def date_between_dates(self, date_start=None, date_end=None):
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -7,7 +7,7 @@ import re
 import sys
 
 from calendar import timegm
-from datetime import timedelta, MAXYEAR, date
+from datetime import timedelta, MAXYEAR
 from time import time
 
 from dateutil import relativedelta
@@ -1576,14 +1576,12 @@ class Provider(BaseProvider):
         :example Date('1999-02-02')
         :return Date
         """
-        
         start_date = self._parse_date(start_date)
-        
         end_date = self._parse_date(end_date)
-        
+
         if sys.platform == 'win32':
-          if (start_date <= date(1970, 1, 1)) | (end_date >= date(3000, 1, 1)):
-            raise ValueError("Windows does not support using this method with date values before 1970-1-1 or after {0}".format(date(3000, 1, 1)))
+            if (start_date <= date(1970, 1, 1)) | (end_date >= date(3000, 1, 1)):
+                raise ValueError("Windows doen't support using this method with values before 1970-1-1 or after 3000-1-1")
 
         return self.date_between_dates(date_start=start_date, date_end=end_date)
 
@@ -1670,11 +1668,10 @@ class Provider(BaseProvider):
             datetime_to_timestamp(datetime_end),
         )
         if tzinfo is None:
-          pick = datetime.fromtimestamp(timestamp, tzlocal())
-          pick = pick.astimezone(tzutc()).replace(tzinfo=None)
+            pick = datetime.fromtimestamp(timestamp, tzlocal())
+            pick = pick.astimezone(tzutc()).replace(tzinfo=None)
         else:
-          pick = datetime.fromtimestamp(timestamp, tzinfo)
-        
+            pick = datetime.fromtimestamp(timestamp, tzinfo)
         return pick
 
     def date_between_dates(self, date_start=None, date_end=None):
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -1578,11 +1578,7 @@ class Provider(BaseProvider):
         """
         start_date = self._parse_date(start_date)
         end_date = self._parse_date(end_date)
-
-        if sys.platform == 'win32':
-            if (start_date <= date(1970, 1, 1)) | (end_date >= date(3000, 1, 1)):
-                raise ValueError("Windows doen't support using this method with values before 1970-1-1 or after 3000-1-1")
-
+        print(start_date + timedelta(days=-365))
         return self.date_between_dates(date_start=start_date, date_end=end_date)
 
     def future_datetime(self, end_date='+30d', tzinfo=None):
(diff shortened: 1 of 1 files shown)


fcurella:
Hi @BrodyUnderwood1 

Thanks for your help.

MY understanding for #829 is that Windows has different bounds than *nix systems. Specifically, it only supports dates from `1970-01-01T00:00:01` through 2038-12-31T23:59:59`.

Instead of forcing it and try to make it work, I'd rather put in the hands of the user. I would check that the the method is not called with 'out of bounds' arguments, and if it is, raise a `ValueError` with a helpful message.

BrodyUnderwood1:
Sounds good I'll play with it when I have time again but anyone who can get to it first feel free to beat me to it if you like :P

faker/providers/date_time/__init__.py
```diff
@@ -1661,11 +1661,14 @@ def date_time_between_dates(
...
-        else:
-            pick = datetime.fromtimestamp(timestamp, tzinfo)
+        try:
+          if tzinfo is None:
+              pick = datetime.fromtimestamp(timestamp, tzlocal())
+              pick = pick.astimezone(tzutc()).replace(tzinfo=None)
+          else:
+              pick = datetime.fromtimestamp(timestamp, tzinfo)
+        except OSError:
+          raise OSError('Argument(s) out of bounds.  On windows machines dates provided must occur in the range of 1970-01-01T00:00:01 and 2038-12-31T23:59:59')
```
fcurella: I think we should prevent out-of-bounds arguments from ever happening. Can we have the check happen _before_ we even try to instantiate `timestamp`? Just check if `datetime_start` or `datetime_end` are within the bounds.

BrodyUnderwood1: (in advance forgive me if I'm mistaken, still a student). While that would improve performance on Windows machine wouldn't the check process just be unnecessary overhead on all other systems that can have values outside that range?

fcurella: You would check the boundaries only on windows. To do that, you can use `sys.platform`:

```python
import sys

if sys.platform == 'win32':
    # if datetime_start ....
    # if datetime_end ....
```

Besides, I'm not really worried about the impact of `if` statements. I look out mainly for for-loops, especially when nested.

kungfu71186: @fcurella I feel like that would be a really bad idea to only limit between 1970 and now. This issue was already resolved in another PR as well.

See https://github.com/joke2k/faker/blob/cb333b3b2950860ee6f3e57cb604f107ddda6fd3/faker/providers/date_time/__init__.py#L1543

Can't use fromtimestamp, instead need to use a timedelta. for times before 1970, it would be a negative number.

fcurella: @BrodyUnderwood1 have you verified that the issue still exists?

BrodyUnderwood1: It did yesterday. I'm out getting lunch now, I'll test again when I'm back.

BrodyUnderwood1: Change of plans probably can't today.  Maybe tomorrow

BrodyUnderwood1:
After testing again I'm unable to reproduce the error I got last week

BrodyUnderwood1:
still a WIP but closer, need to correct max date that can be generated.

kungfu71186:
@BrodyUnderwood1 @fcurella Again, I would highly discourage setting a boundary. There is already a workaround in place to resolve this issue, so not sure why there is a boundary.

Please see https://github.com/joke2k/faker/blob/ba4994a9c1f84bab3aa3b21c7cc2a1e0e76d3f1c/faker/providers/date_time/__init__.py#L1561

As you can see on Windows this does work fine:

```
from datetime import datetime
datetime(1970, 1, 1)
datetime(1970, 1, 1) - timedelta(1000)
datetime(1970, 1, 1) + timedelta(1000)
datetime(1970, 1, 1) - timedelta(100000)
datetime(1970, 1, 1) + timedelta(100000)
datetime(1970, 1, 1) - timedelta(719162) //year 1
```

BrodyUnderwood1:
@kungfu71186 Was only doing it because he suggested.  Threw your proposed changes in for consideration as well

fcurella:
@BrodyUnderwood1 Thank you for all your hard work, but if there isn't an issue, I'd rather avoid any changes. I took a lot of work to get the `date_time` to where it's working across platforms, and I don't won't to run the risk of introducing other bugs.

BrodyUnderwood1:
@fcurella as is, there is an issue.  the most recent fix is literally a copy-paste-ish solution from the function above when had been changed in the same way (presumably to fix the same error).  The method we are working on does still throw an error when running on windows. Run the 3 lines of code I pasted above and you'll see that.  What @kungfu71186 is saying isn't that this method doesn't throw an error, merely that we could implement the exact same fix that had to be implemented in another method already.

fcurella:
@BrodyUnderwood1 Oh I see now. Thanks for clarifying.

@kungfu71186 you're more knowledgeable than me on this issue. Would you mind review the current status of this PR? If it looks good to you, I will merge it.

kungfu71186:
@fcurella Yes, please see #563 for more information. It is a bug in the windows code. 

Now with that said, I do feel like the methods are duplicates of each other. I'd have to test again, but I think you can put in a date or even something like x days.

date_time_between and date_time_between_dates are basically the same function and if memory serves me right, these can be used interchangeably. 

For instance,

```
date_time_between(self, start_date='-30y', end_date='now', tzinfo=None):
```
or

```
date_time_between(self, start_date=DateTime('1999-02-02 11:42:52'), end_date='now', tzinfo=None):
```
I'd have to test to verify, but I'm pretty sure you can do either or with that same function.

If that's the case, I would suggest merging the 2 and just have one call the other or get rid of one completely. But for compatibility with code, we could probably just call the `date_time_between` function from `date_time_between_dates`

BrodyUnderwood1:
@kungfu71186 @fcurella they did seem INCREDIBLY similar if not exactly the same.  might be best just to add the functionality of one to the other and merge the two if they are different at all

PR #836, opened 2018-10-15
Add method to generate a cell phone number to pt-BR localization

### LINKED ISSUE #760: Add method to generate a cell phone number to pt-BR
Faker doesn't have a function to generate a cellphone to Brazilian.

Steps to reproduce
Create fake instance using localization "pt_BR"
Call fake.msisdn() or fake.phone_number()
Expected behavior
It should generate a cell phone number.

Actual behavior
Sometimes these methods return a "residential" numbers.

Reference difference between cell phones and residential numbers:

http://www.teleco.com.br/num.asp

Newman101:
### What does this changes

As per #760, Brazilian localization method for ```cellphone_number()``` method was displaying landline number.

### What was wrong

The main problem was that the ```pt_BR``` localization was missing a ```cellphone_number()``` method. As such, the format was defaulting to landline.

### How this fixes it

The method ```cellphone_number``` has been now added to main localization with a relevant unit test. Tested with ```python setup.py test```, all green.

Fixes #760.

### DIFF
--- a/faker/providers/phone_number/pt_BR/__init__.py
+++ b/faker/providers/phone_number/pt_BR/__init__.py
@@ -71,6 +71,7 @@ class Provider(PhoneNumberProvider):
         '#### ####',
         '####-####',
     )
+    
     msisdn_formats = (
         '5511#########',
         '5521#########',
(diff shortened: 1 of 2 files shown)


fcurella:
Thank you! ✨

PR #837, opened 2018-10-15
[fixed] image_url generates invalid urls.

tsiaGeorge:
### What does this changes
fixes wrong formatted `image_url`s

### What was wrong
In the latest version there was added a ','
in the list `image_placeholder_services`,
which introduces an invalid url in the second element of the list.

e.g. '?txtsize=55&txt={width}x{height}&w={width}&h={height}'
### How this fixes it
Restored the list by removing the wrong ','  
Now the first element of the list is:
'https://placeholdit.imgix.net/~text?txtsize=55&txt={width}x{height}&w={width}&h={height}'


fcurella:
Nice catch, Thank you! ✨

tsiaGeorge:
Your welcome! 🍺

PR #839, opened 2018-10-16
Add phone number format in Thailand :bear:

zkan:
### What does this changes

* Add phone number format in Thailand.
* Add test to check the format.


fcurella:
Thank you! ✨

PR #840, opened 2018-10-16
Add bank provider for pl_PL locale

andrzej3393:
### What does this changes

Adds bank provider for pl_PL locale.

### What was wrong

Bank provider for pl_PL locale didn't exist before.

### How this fixes it

Adds bank provider for pl_PL locale.


fcurella:
Thank you! ✨

PR #841, opened 2018-10-16
Add lorem provider for pl_PL locale

andrzej3393:
### What does this changes

Adds lorem provider for pl_PL locale.

### What was wrong

Lorem provider for pl_PL locale didn't exist before.

### How this fixes it

Adds lorem provider for pl_PL locale.


faker/providers/lorem/pl_PL/__init__.py
```diff
@@ -0,0 +1,208 @@
+from __future__ import unicode_literals
```
fcurella: you'll need to add `# coding=utf-8` at the top of the file

andrzej3393: Fixed

fcurella:
Thank you! ✨

PR #842, opened 2018-10-17
Update README.rst to improve consistency

grrttedwards:
Update the README to be internally consistent. 
Some return values were denoted with the `>` character, while some were `#`-style Python comments. Also, a few were quoted, and now they are all quoted, to improve clarity of the return value.


fcurella:
Thank you! ✨

PR #846, opened 2018-10-20
Feature/tax ids

### LINKED ISSUE #844: How to write external provider with locale support?

### Steps to reproduce

1. Build an external provider package
2. Set the locale for Faker
3. Generate a fake using your own provider

### Expected behavior

The output should adjust depending on the locale

### Actual behavior

Only the default output is produced, the localized subclasses are completely ignored

### See also
I've tried to build some european tax IDs (and German tax numbers) into an external package:
https://github.com/mastacheata/faker_tax

The test for the danish locale fails all but the first testcase in the test. (i.e.: The length is not 10, and the prefix is not DK)

mastacheata:
### What does this changes

This adds VAT IDs for European Union Locales to the SSN provider

### What was wrong

It was simply a feature I was missing and thought would be worth to share.

### How this fixes it

Create VAT ID providers for most european countries according to this page: http://ec.europa.eu/taxation_customs/vies/faq.html#item_11

Fixes #844

### DIFF
--- a/faker/providers/ssn/fr_CH/__init__.py
+++ b/faker/providers/ssn/fr_CH/__init__.py
@@ -31,3 +31,20 @@ class Provider(SsnProvider):
                          + digits[7:11] + '.' \
                          + digits[11:]
         return ssn
+
+    def vat_id(self):
+        """
+        :return: Swiss UID number
+        """
+        def _checksum(digits):
+            code = ['8', '6', '4', '2', '3', '5', '9', '7']
+            digits = [*digits]
+            remainder = 11-(sum(map(lambda x, y: int(x) * int(y), code, digits)) % 11)
+            if remainder == 10:
+                return 0
+            elif remainder == 11:
+                return 5
+            return remainder
+
+        vat_id = self.bothify('########')
+        return 'CHE' + vat_id + _checksum(vat_id)
--- a/faker/providers/ssn/bg_BG/__init__.py
+++ b/faker/providers/ssn/bg_BG/__init__.py
@@ -1,0 +1,21 @@
+# coding=utf-8
+from .. import Provider as BaseProvider
+
+
+class Provider(BaseProvider):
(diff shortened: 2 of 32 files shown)


mastacheata:
Apparently this PR fails to pick up the 4th commit I pushed just a few minutes ago and which should fix all 4 problems. (3 failed tests due to last minute changes after I ran my local tests, sorry about that and one more build error I didn't see coming due to no implicit type conversion in python)

mastacheata:
Apparently this PR fails to pick up the 4th commit I pushed just a few minutes ago and which should fix all 4 problems. (3 failed tests due to last minute changes after I ran my local tests, sorry about that and one more build error I didn't see coming due to no implicit type conversion in python)

mastacheata:
Apparently this PR fails to pick up the 4th commit I pushed just a few minutes ago and which should fix all 4 problems. (3 failed tests due to last minute changes after I ran my local tests, sorry about that and one more build error I didn't see coming due to no implicit type conversion in python)

faker/providers/ssn/fr_CH/__init__.py
```diff
@@ -31,3 +31,20 @@ def _checksum(digits):
...
                          + digits[11:]
         return ssn
+
+    def vat_id(self):
+        """
+        :return: Swiss UID number
+        """
+        def _checksum(digits):
+            code = ['8', '6', '4', '2', '3', '5', '9', '7']
+            digits = [*digits]
```
fcurella: what's the purpose of making a copy of the argument and assigning it the same name?

mastacheata: I'm new to python (about 2.5 months since I first touched a piece of Python code) 
I thought this was for converting the string into an array, but now that you say it, I guess python treats strings as arrays of chars anyways and this isn't actually needed.

PR #848, opened 2018-10-23
Provider improvements for cs_CZ address provider

Newman101:
### What does this changes

Improvements for ```cs_CZ``` provider:

- Add unit test class for the address provider
- Add more methods for address provider

### What was wrong

Issues: 

- The provider was missing unit test class completely.
- The provider was missing city_with_postcode method.

### How this fixes it

Implemented the features mentioned above.

Unit tests executed locally - all green.


fcurella:
Thank you! ✨

PR #850, opened 2018-10-25
Add more thai job

lifez:
Add more thai job


fcurella:
Thank you @lifez.

We try to avoid making up our own lists, and prefer to rely on lists created by other sources.

This list in particular was originally taken from https://th.wikipedia.org/wiki/หมวดหมู่:บุคคลแบ่งตามอาชีพ, and it doesn't include the items you added.

If you really want to add them, you could update the list on wikipedia, wait a couple of weeks for the wikipedia's team edits, and then update the provider with the new content from the list.

fcurella:
Closing for inactivity

PR #852, opened 2018-10-26
Provider improvements for de_AT address provider

Newman101:
### What does this changes

Improvements for de_AT provider:

- Add more methods for address provider

### What was wrong

Issues:

- The provider was missing city_with_postcode method.

### How this fixes it

Implemented the features mentioned above.

Unit tests executed locally - all green.


fcurella:
Thank you! ✨

PR #855, opened 2018-10-27
Add format to numbers with 2 indicators

simaolemos:
### What does this changes

I've added a new format to phone numbers with 2 indicators, used for example in Porto, Lisbon...


faker/providers/phone_number/pt_PT/__init__.py
```diff
@@ -19,6 +20,7 @@ class Provider(PhoneNumberProvider):
         '(351) 93# ### ###',
         '(351) 96# ### ###',
         '(351) 2## ### ###',
+        '(351) 2# ### ### #)',
```
fcurella: is that trailing `)` a typo?

faker/providers/phone_number/pt_PT/__init__.py
```diff
@@ -9,6 +9,7 @@ class Provider(PhoneNumberProvider):
         '(351) 93# ### ###',
         '(351) 96# ### ###',
         '(351) 2## ### ###',
+        '(351) 2# ### ### #',
```
fcurella: while we're here, shouldn't all those be `+351` instead of `(351)`?

faker/providers/phone_number/pt_PT/__init__.py
fcurella: If I'm reading https://en.wikipedia.org/wiki/Telephone_numbers_in_Portugal correctly, it's always `21`, ie:  there isn't any number with a prefix if `22`, `23`, etc. Can you confirm?

simaolemos: yep, my bad, i'll solve this

faker/providers/phone_number/pt_PT/__init__.py
simaolemos: +351 is equal to (351) basicly, we can use only the +351 however im only following the already created content

faker/providers/phone_number/pt_PT/__init__.py
simaolemos: this is not truth, we have many prefix's, for example 21 is used to the Lisbon, 22 is to Porto, 252 is to Vila Nova de Famalicão and so one so forth

faker/providers/phone_number/pt_PT/__init__.py
fcurella: Looking at other providers, I think we should have both, unless it's really uncommon to see `+351` in Portugal.

Do you feel like adding `+351` versions of all the `(351)` patterns? That is, unle

PR #861, opened 2018-10-30
Added enhancement that allows uuid4 generator to return UUID object

### LINKED ISSUE #858: uuid4 should return UUID instead of str
__Context__: I'm using faker in combination with [FactoryBoy](https://github.com/FactoryBoy/factory_boy) to test [Django](https://www.djangoproject.com/) models. I have this test to make sure some data dict generated from model data is JSON-serializable. The test passed, but not the code in production. (`Object of type UUID is not JSON serializable`).

Now, of course the fix to the offending code is simple:

``` diff
- data['uuid'] = model.uuid
+ data['uuid'] = str(model.uuid)
```

but the problem is two-fold:

1. __The fact that `uuid4` returns `str` is confusing__. I didn't pay attention and couldn't test properly.
2. __There is no provider for actual `UUID` objects__. I can understand if some people want strings that are in fact uuid representations, after all that's [how AutoFixture generates strings](https://github.com/AutoFixture/AutoFixture/wiki/Cheat-Sheet#completely-autogenerated-string) (Autofixture does approximately the same job as faker, but for C#). But there should be a provider for true `UUID`s as well.

### Steps to reproduce

1. Call `fake.uuid4()`

### Expected behavior

`fake.uuid4()` should return an `UUID` object

### Actual behavior

`fake.uuid4()` returns a `str`

ediblesushi:
### What does this changes

This allows faker.uuid4() the option to return a UUID object instead of a string. faker.uuid4() still returns a string as its default functionality.

### What was wrong

faker.uuid4() only was able to return a string

### How this fixes it

Fixes #858

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -75,7 +75,7 @@ class Provider(BaseProvider):
     tlds = (
         'com', 'com', 'com', 'com', 'com', 'com', 'biz', 'info', 'net', 'org',
     )
-
+    hostname_prefixes = ('db', 'srv', 'desktop', 'laptop', 'lt', 'email', 'web')
     uri_pages = (
         'index', 'home', 'search', 'main', 'post', 'homepage', 'category',
         'register', 'login', 'faq', 'about', 'terms', 'privacy', 'author',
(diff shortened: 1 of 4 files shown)


ThibaultLemaire:
You should have mentioned you were taking this issue, I was about to do it myself :smile:

Issue #860, opened 2018-10-30
fake.uuid4() doesn't return a version 4 UUID


ThibaultLemaire:
[A UUID should have 4bits indicating its version](https://en.wikipedia.org/wiki/Universally_unique_identifier#Format). Meaning a [uuid4 isn't 128 bits of entropy, but only 122](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)).

The current implementation of `uuid4` returns a truly random UUID, so if any code is checking the version number of the UUID it would fail.

for reference, here is the [current offending code](https://github.com/joke2k/faker/blob/d26db45eebb9dcd02eb73099bb98b660f0e03aad/faker/providers/misc/__init__.py#L138), and here is the [cpython implementation for `uuid4`](https://github.com/python/cpython/blob/54752533b2ed1c898ffe5ec2e795c6910ee46a39/Lib/uuid.py#L757).

Luckily, the fix is trivial:

``` diff
- UUID(int= ... )
+ UUID(int= ... , version=4)
```

### Steps to reproduce

``` python
from faker import Faker
fake = Faker()
fake.seed(1)
uuid = fake.uuid4()
uuid
# 'cd613e30-d8f1-6adf-91b7-584a2265b1f5'
```

### Expected behavior

``` python
uuid[14] == '4'
```

The digit at this position should indicate the version of the uuid, in our case `4`.

### Actual behavior

``` python
uuid[14] == '6'
```

PS: Apparently nobody needed this so far, otherwise I wouldn't be the one to report this. I actually only noticed when looking at the source code. (Who checks version numbers anyway?)

ThibaultLemaire:
Fixed by #863

PR #863, opened 2018-10-30
Added enhancement that allows uuid4 generator to return UUID object

### LINKED ISSUE #858: uuid4 should return UUID instead of str

ediblesushi:
### What does this changes
This allows faker.uuid4() the option to return a UUID object instead of a string. faker.uuid4() still returns a string as its default functionality.

### What was wrong
faker.uuid4() only was able to return a string

### How this fixes it

Fixes #858

### DIFF
--- a/faker/providers/misc/__init__.py
+++ b/faker/providers/misc/__init__.py
@@ -133,7 +133,8 @@ class Provider(BaseProvider):
     def uuid4(self, cast_to=str):
         """
         Generates a random UUID4 string.
-        @param cast_to: String. What date type you would like the UUID to be cast to. Default is string.
+        :param cast_to: Specify what type the UUID should be cast to. Default is `str`
+        :type cast_to: callable, None
         """
         # Based on http://stackoverflow.com/q/41186818
         value = uuid.uuid4(int=self.generator.random.getrandbits(128))
--- a/faker/providers/misc/__init__.py
+++ b/faker/providers/misc/__init__.py
@@ -134,13 +134,10 @@ class Provider(BaseProvider):
         """
         Generates a random UUID4 string.
         :param cast_to: Specify what type the UUID should be cast to. Default is `str`
-        :type cast_to: callable, None
+        :type cast_to: callable
         """
         # Based on http://stackoverflow.com/q/41186818
-        value = uuid.UUID(int=self.generator.random.getrandbits(128), version=4)
-        if cast_to:
-            return cast_to(value)
-        return value
+        return cast_to(uuid.UUID(int=self.generator.random.getrandbits(128), version=4))
 
     def password(
             self,
--- a/faker/providers/misc/__init__.py
+++ b/faker/providers/misc/__init__.py
@@ -130,11 +130,14 @@ class Provider(BaseProvider):
     def language_code(self):
         return self.random_element(Provider.language_locale_codes.keys())
 
-    def uuid4(self):
+    def uuid4(self, uuid_object=False):
         """
         Generates a random UUID4 string.
+        @param uuid_object: Boolean. Whether to return UUID object or string
         """
         # Based on http://stackoverflow.com/q/41186818
+        if uuid_object:
+            return uuid.UUID(int=self.generator.random.getrandbits(128))
         return str(uuid.UUID(int=self.generator.random.getrandbits(128)))
 
     def password(
--- a/faker/providers/misc/__init__.py
+++ b/faker/providers/misc/__init__.py
@@ -136,7 +136,7 @@ class Provider(BaseProvider):
         @param cast_to: String. What date type you would like the UUID to be cast to. Default is string.
         """
         # Based on http://stackoverflow.com/q/41186818
-        value = uuid.UUID(int=self.generator.random.getrandbits(128))
+        value = uuid.uuid4(int=self.generator.random.getrandbits(128))
         if cast_to:
             return cast_to(value)
         return value
--- a/faker/providers/misc/__init__.py
+++ b/faker/providers/misc/__init__.py
@@ -130,15 +130,16 @@ class Provider(BaseProvider):
     def language_code(self):
         return self.random_element(Provider.language_locale_codes.keys())
 
-    def uuid4(self, uuid_object=False):
+    def uuid4(self, cast_to=str):
         """
         Generates a random UUID4 string.
         @param uuid_object: Boolean. Whether to return UUID object or string
         """
         # Based on http://stackoverflow.com/q/41186818
-        if uuid_object:
-            return uuid.UUID(int=self.generator.random.getrandbits(128))
-        return str(uuid.UUID(int=self.generator.random.getrandbits(128)))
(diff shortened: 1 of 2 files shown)


faker/providers/misc/__init__.py
```diff
@@ -130,11 +130,14 @@ def locale(self):
     def language_code(self):
         return self.random_element(Provider.language_locale_codes.keys())
 
-    def uuid4(self):
+    def uuid4(self, uuid_object=False):
```
fcurella: I would rather have a `cast_to` argument for more flexibility. Something like:

```python
def uuid(self, cast_to=str):
    value = uuid.uuid4()
    if cast_to is not None:
        return cast_to(value)
    return value
```

This would allow users to cast to `int`, or to `hex`:

```python
# currently UUIDs can't be casted with `hex`. See https://bugs.python.org/issue35115
from operator import attrgetter

faker.uuid4(cast_to=int)
faker.uuid4(cast_to=attrgetter('hex'))

Comment (ediblesushi) [reply faker/providers/misc/__init__.py]: Good idea I'll start working on changing it

File: faker/providers/misc/__init__.py [hunk 2]
```diff
@@ -130,12 +130,16 @@ def locale(self):
...
 
-    def uuid4(self):
+    def uuid4(self, cast_to=str):
         """
         Generates a random UUID4 string.
+        @param uuid_object: Boolean. Whether to return UUID object or string
         """
         # Based on http://stackoverflow.com/q/41186818
-        return str(uuid.UUID(int=self.generator.random.getrandbits(128)))
+        value = uuid.UUID(int=self.generator.random.getrandbits(128))
```
fcurella: let's use `uuid.uuid4()`. See https://github.com/joke2k/faker/issues/860

faker/providers/misc/__init__.py
```diff
@@ -130,12 +130,16 @@ def locale(self):
     def language_code(self):
         return self.random_element(Provider.language_locale_codes.keys())
 
-    def uuid4(self):
+    def uuid4(self, cast_to=str):
         """
         Generates a random UUID4 string.
+        @param uuid_object: Boolean. Whether to return UUID object or string
```
fcurella: remember to update the docstring

faker/providers/misc/__init__.py
```diff
@@ -130,12 +130,16 @@ def locale(self):
     def language_code(self):
         return self.random_element(Provider.language_locale_codes.keys())
 
-    def uuid4(self):
+    def uuid4(self, cast_to=str):
         """
         Generates a random UUID4 string.
+        @param cast_to: String. What date type you would like the UUID to be cast to. Default is string.
```
fcurella: `cast_to` is supposed to be a callable, and the default is the built-in callable `str`

ediblesushi: Im not too familiar with docstring syntax and best practices. How should I document the parameter correctly?

fcurella: You can use the [Sphinx syntax](http://www.sphinx-doc.org/en/1.4.8/domains.html#info-field-lists), like this:

```
:param cast_to:  What type you would like the UUID to be cast to. Default is `str`
:type cast_to: callable, None
```

faker/providers/misc/__init__.py
```diff
@@ -130,12 +130,17 @@ def locale(self):
...
-    def uuid4(self):
+    def uuid4(self, cast_to=str):
         """
         Generates a random UUID4 string.
+        :param cast_to: Specify what type the UUID should be cast to. Default is `str`
+        :type cast_to: callable, None
         """
         # Based on http://stackoverflow.com/q/41186818
-        return str(uuid.UUID(int=self.generator.random.getrandbits(128)))
+        value = uuid.uuid4(int=self.generator.random.getrandbits(128))
```
fcurella: sorry, my bad. `uuid4` doesn't support the `int` argument. Can you change this line to `value = uuid.UUID(int=self.generator.random.getrandbits(128), version=4)`?

faker/providers/misc/__init__.py
```diff
@@ -130,12 +130,17 @@ def locale(self):
     def language_code(self):
         return self.random_element(Provider.language_locale_codes.keys())
 
-    def uuid4(self):
+    def uuid4(self, cast_to=str):
```
ThibaultLemaire: I find the use of `cast_to` a (tiny) bit confusing in case of `None`. Consider

``` python
fake.uuid4(cast_to=str)
fake.uuid4(cast_to=None)
```
The first one reads as "cast to string", and the second reads as "cast to none", but I don't want my uuid to be `None`!

What if it's up to the user to provide an identity function when they want a plain UUID?

``` python
fake.uuid4(cast_to=lambda x: x)
```

fcurella: Hi agree with you @ThibaultLemaire , let's remove the special case for `None`.

tests/providers/test_misc.py
```diff
@@ -0,0 +1,25 @@
...
+        assert uuid4
+        assert isinstance(uuid4, six.string_types)
+
+    def test_uuid4_int(self):
+        uuid4 = self.factory.uuid4(cast_to=int)
+        assert uuid4
+        assert isinstance(uuid4, six.integer_types)
+
+    def test_uuid4_uuid_object(self):
+        uuid4 = self.factory.uuid4(cast_to=None)
```
fcurella: we're not special-casing anymore. Can you change this line to `uuid=self.factory.uuid(cast_to=lambda x: x)`?

tests/providers/test_misc.py
```diff
@@ -0,0 +1,25 @@
+import unittest
+import uuid
+import six
+
+from faker import Faker
+
```
fcurella: could you add 1 more blank line here?

fcurella:
Thank you! ✨

PR #867, opened 2018-11-06
More fi unit tests

Newman101:
### What does this changes

Added unit test for state() method in fi_FI locale.

### What was wrong

Unit test was missing.

### How this fixes it

Added the relevant unit test to the test file. 

Unit tests executed locally - all green. Thanks.

### DIFF
--- a/tests/providers/test_address.py
+++ b/tests/providers/test_address.py
@@ -167,7 +167,9 @@ class TestFiFI(unittest.TestCase):
         assert suffix in FiProvider.street_suffixes
 
     def test_state(self):
-        # TODO
+        state = self.factory.state()
+        assert isinstace(state, string_types)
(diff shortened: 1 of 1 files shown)


tests/providers/test_address.py
```diff
@@ -166,6 +166,11 @@ def test_street_suffix(self):
         assert isinstance(suffix, string_types)
         assert suffix in FiProvider.street_suffixes
 
+    def test_state(self):
+        state = self.factory.state()
+        assert isinstace(state, string_types)
```
fcurella: typo: `isinstance` (last `n` is missing)

Newman101: Thanks for notifying me. That's been now fixed. :+1:

PR #869, opened 2018-11-14
Add ranges to pyint

rafenden:
### What does this changes

It adds two new parameters to `pyint()` to specify `min` and `max` values.

Example usage:
```
fake.pyint(0, 1000000)
```

### What was wrong

There was no way to specify ranges for random integers generated by faker, usually, I ended using `random.randint()`.

### How this fixes it

Devs can now specify ranges when calling `pyint()`


fcurella:
Thanks @rafenden !

Could you add some tests?

fcurella:
Done at #937

PR #871, opened 2018-11-21
Correct number of digits in norwegian phonenumbers

aleksanb:
The first format had 9 digits, but norwegian phone numbers have 8 digits (excluding country code).
I was writing tests for some code, where libphonenumber suddenly starter failing the region formatting check, as norwegian phone numbers have 8 digits and not 9.


fcurella:
Thank you! ✨

PR #872, opened 2018-11-27
Address Unit Tests: fr_FR provider

Newman101:
### What does this changes

Adds unit tests for ```fr_FR``` address provider.

### What was wrong

The unit test module for ```fr_FR``` address provider was missing.

### How this fixes it

Implemented the unit tests.

Unit tests executed locally - all green. Thank you.


fcurella:
Thank you! ✨

Issue #773, opened 2018-06-17
Pipeline: Coverage report does not reflect what actually changed in a PR


crd:
Brief summary of the issue goes here.

### Steps to reproduce

Consider https://coveralls.io/builds/17537497, which corresponds to commit 6908bad3b2056c1733fc04ceed2aad76bb8b1e27 in PR #772. Coveralls reports a drop in coverage of 1.4%, in files not touched by the PR. I suspect that state may be kept between PR coverage evaluations. Instead, coverage should be compared against the current state of the `master` branch.

crd:
I did some digging and found this: https://github.com/lemurheavy/coveralls-public/issues/1019

crd:
I think this issue is exacerbated by the fact that so many of the tests fail to specify a seed value. This means that different branches of the codebase will be exercised randomly. For example, I ran `coverage` a few times and noticed this inconsistent behavior:

```
$ coverage report -m faker/providers/company/fr_FR/__init__.py
Name                                        Stmts   Miss  Cover   Missing
-------------------------------------------------------------------------
faker/providers/company/fr_FR/__init__.py      40      2    95%   109, 128
```

A few seconds later, with no code changes:

```
$ coverage report -m faker/providers/company/fr_FR/__init__.py
Name                                        Stmts   Miss  Cover   Missing
-------------------------------------------------------------------------
faker/providers/company/fr_FR/__init__.py      40      1    98%   128
```

PR #874, opened 2018-12-02
Add categories to Japanese company provider

shirakia:
### What does this changes

Added variety of Japanese company names.

The category list is from Ruby Faker. https://github.com/stympy/faker/blob/master/lib/locales/ja.yml#L28

### What was wrong

Nothing was wrong.

### How this fixes it

Added categories and format patterns.

### DIFF
--- a/faker/providers/company/ja_JP/__init__.py
+++ b/faker/providers/company/ja_JP/__init__.py
@@ -5,15 +5,15 @@ from .. import Provider as CompanyProvider
 
 class Provider(CompanyProvider):
     formats = (
-        '{{company_prefix}}{{last_name}}{{category}}',
-        '{{last_name}}{{category}}{{company_prefix}}',
+        '{{company_prefix}}{{last_name}}{{company_category}}',
+        '{{last_name}}{{company_category}}{{company_prefix}}',
     )
 
     company_prefixes = ('株式会社', '有限会社', '合同会社')
-    categories = ('水産', '農林', '鉱業', '建設', '食品', '印刷', '電気', 'ガス', '情報', '通信', '運輸', '銀行', '保険')
+    company_categories = ('水産', '農林', '鉱業', '建設', '食品', '印刷', '電気', 'ガス', '情報', '通信', '運輸', '銀行', '保険')
 
     def company_prefix(self):
         return self.random_element(self.company_prefixes)
 
-    def category(self):
-        return self.random_element(self.categories)
+    def company_category(self):
+        return self.random_element(self.company_categories)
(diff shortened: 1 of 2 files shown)


shirakia:
Coveralls says coverage decreased. But this part is completely unrelated to this PR.
Can someone who has WRITE access rebuild travis? It will solve coverage decreasing.

The problem is issued here https://github.com/joke2k/faker/issues/773

faker/providers/company/ja_JP/__init__.py
```diff
@@ -5,10 +5,15 @@
...
+        '{{last_name}}{{category}}{{company_prefix}}',
     )
 
     company_prefixes = ('株式会社', '有限会社', '合同会社')
+    categories = ('水産', '農林', '鉱業', '建設', '食品', '印刷', '電気', 'ガス', '情報', '通信', '運輸', '銀行', '保険')
 
     def company_prefix(self):
         return self.random_element(self.company_prefixes)
+
+    def category(self):
```
fcurella: Unfortunately, `category` is too generic of a word.

The problem is that in the Faker API, provider methods are not scoped by their provider, but they are all accessed from the root `Faker` instance. This means that people will call `faker.category()`, and I'm not sure we should return a list of categories that are restricted to companies.

Maybe we can just call it `company_category`?

faker/providers/company/ja_JP/__init__.py
```diff
@@ -5,10 +5,15 @@
 
 class Provider(CompanyProvider):
     formats = (
-        '{{company_prefix}} {{last_name}}',
+        '{{company_prefix}}{{last_name}}{{category}}',
+        '{{last_name}}{{category}}{{company_prefix}}',
     )
 
     company_prefixes = ('株式会社', '有限会社', '合同会社')
+    categories = ('水産', '農林', '鉱業', '建設', '食品', '印刷', '電気', 'ガス', '情報', '通信', '運輸', '銀行', '保険')
```
fcurella: Cor easier testing, could you rename this `company_categories`?

faker/providers/company/ja_JP/__init__.py
shirakia: Thanks. I understand it.
Just renamed them `company_category` and `company_categories`

tests/providers/test_company.py
```diff
@@ -45,7 +45,7 @@ def test_company(self):
         company = self.factory.company()
         assert isinstance(company, six.string_types)
         assert any(prefix in company for prefix in prefixes)
-        assert any(company.startswith(prefix) for prefix in prefixes)
+        assert any(company.startswith(prefix) or company.endswith(prefix) for prefix in prefixes)
```
fcurella: can you also add a test that calls the new `company_category` method?

shirakia: @fcurella Added some tests 👍

fcurella:
Thank you! ✨

Issue #875, opened 2018-12-03
Wrong class in localization text of README.rst


jocassid:
In README.rst the text under the Localization says faker.Factory, but the code example uses faker.Faker

### Steps to reproduce

1. Read localization section of README.rst

### Expected behavior

should read faker.Faker

### Actual behavior

reads faker.Factory

jianbin-allseeing:
should this issue be closed?  The test failed due to deleted branch. @fcurella

fcurella:
Yes, it can be closed. Thank you @jianbin-allseeing

PR #876, opened 2018-12-03
#875 localization text fix

jocassid:
The class in the test didn't match the sample code

### What does this changes

Brief summary of the changes

### What was wrong

### How this fixes it

Fixes #...


fcurella:
Thank you! ✨

PR #880, opened 2018-12-11
Add trunk prefix for ru_RU phone numbers

pishchalnikov:
### What does this changes
Added a new phone number formats with trunk prefix in ru_RU.


fcurella:
Thank you! ✨

PR #887, opened 2018-12-26
Remove installation of pep8 from Travis CI

jdufresne:
Linting is handled by the flake8 tox target which manages its own dependencies. Installing pep8 every run is unnecessary overhead.


fcurella:
Thank you! ✨

PR #888, opened 2018-12-31
Change id_ID states_abbr list

dt-ap:
### What does this changes

Change Indonesian states ('provinsi' in Indonesian) abbreviation. The list is from Indonesian Standardization Agency, which used ISO 3166-2 standard

[link](https://id.wikipedia.org/wiki/Daftar_provinsi_di_Indonesia)


fcurella:
Thank you so much! ✨

PR #892, opened 2019-01-15
Generating a zipcode within a given state for EnUS

mattyg:
### What does this change

New functions to generate a zipcode within a given state for EnUs locale: postcode_in_state(state_abbr) and aliases functions zipcode_in_state and postalcode_in_state, and tests for those functions.

### What was wrong

No way to specify the state of a generated zipcode.

### How this fixes it

Implemented functionality to generate zipcode that is within a given state


faker/providers/address/en_US/__init__.py
```diff
@@ -353,6 +373,24 @@ def zipcode_plus4(self):
         return "%s-%04d" % (self.zipcode(),
                             self.generator.random.randint(1, 9999))
 
+    def postcode_in_state(self, state_abbr='AL'):
```
fcurella: I would make `state_abbr` mandatory. Same for the other new methods.

faker/providers/address/en_US/__init__.py
```diff
@@ -279,6 +279,26 @@ class Provider(AddressProvider):
         'WV', 'WI', 'WY',
     )
 
+    states_postcode = {
+        'AL': (35004,36925), 'AK': (99501, 99950), 'AZ': (85001,86556),
```
fcurella: PEP8: can you add a space after every comma in the tuples? eg: `'AL': (35004, 36925)`

faker/providers/address/en_US/__init__.py
mattyg: Sorry if this is dumb -- but it's not clear to me the best way to go about this. None of the provider functions have mandatory arguments. Removing the default value causes the doc generation to fail.

I was thinking of replacing the line with this, which seems to be in line with how the other provider functions are setup:

` 
def postcode_in_state(self, state_abbr=None):
    if state_abbr is None or state_abbr in self.states_abbr:
         state_abbr = state_abbr or self.random_element(self.states_abbr)
         ...
`

Does that make sense?

faker/providers/address/en_US/__init__.py
```diff
@@ -373,13 +373,18 @@ def zipcode_plus4(self):
...
         """
         :returns: A random postcode within the provided state abbreviation
 
         :param state_abbr: A state abbreviation
         """
+        if state_abbr is None:
+            state_abbr = self.random_element(self.states_abbr)
+
         if state_abbr in self.states_abbr:
+            state_abbr = state_abbr or self.random_element(self.states_abbr)
```
fcurella: I don't this we need this check. We already normalized on L382

faker/providers/address/en_US/__init__.py
```diff
@@ -279,6 +279,26 @@ class Provider(AddressProvider):
...
+        'MA': (1001, 2791), 'MI': (48001, 49971), 'MN': (55001, 56763),
+        'MS': (38601, 39776), 'MO': (63001, 65899), 'MT': (59001, 59937),
+        'NE': (68001, 68118), 'NV': (88901, 89883), 'NH': (3031, 3897),
+        'NJ': (7001, 8989), 'NM': (87001, 88441), 'NY': (6390, 6390),
+        'NC': (27006, 28909), 'ND': (58001, 58856), 'OH': (43001, 45999),
+        'OK': (73001, 73199), 'OR': (97001, 97920), 'PA': (15001, 19640),
+        'RI': (2801, 2940), 'SC': (29001, 29948), 'SD': (57001, 57799),
+        'TN': (37010, 38589), 'TX': (73301, 73301), 'UT': (84001, 84784),
+        'VT': (5001, 5495), 'VA': (20040, 20041), 'WA': (98001, 99403),
+        'WV': (24701, 26886), 'WI': (53001, 54990), 'WY': (82001, 83128)
```
fcurella: can you add a trailing comma at the end of this line?

fcurella:
Thank you! ✨

mattyg:
@fcurella Thanks for being patient with me! How do you typically run PEP8 linting on your local machine? I set up flake8 and linter-flake8 on atom and just installed flake8-commas to warn me of missing trailing commas. Do you use a different linter? Are there other flake8 packages people typically use?

fcurella:
@mattyg flake-8 is what I use. I tend to follow this [setup](https://www.jonafato.com/2018/03/21/my-flake8-setup).

PR #893, opened 2019-01-16
Added Vietnamese Person Names

JRBusiness:
### What does this changes

Brief summary of the changes

Added Vietnamese Person Names

Adding missing languages

Added First, last and middle names in Vietnamese


JRBusiness:
Fixed some name also added the source for some of the first names.
Thanks

faker/providers/person/vi_VN/__init__.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+from .. import Provider as PersonProvider
+
+#Some of the firstname was base on this list: https://en.m.wikipedia.org/wiki/Vietnamese_name
+#Most of them was from my personal experience.
```
fcurella: I'm sorry but we only accept data coming from statistical sources.

JRBusiness:
Hi, sources are dead and missing many names. Also you can compare mine with the link I included, the source missing names and also the words from source were written in Vietnamese which will not display well for some people like "H?ng, C?ng"

fcurella:
@JRBusiness What do you mean with 'sources are dead'? https://en.m.wikipedia.org/wiki/Vietnamese_name loads fine to me

JRBusiness:
Sorry for missunderstanding, I meant the source is not updated or does not contains enough information. Using static source for address, phone number is great, but for person names we should not follow the source 100%, we can base on it though. Part of my dara was base on it but I added more names and also get rid of the Vietnamese punctuation mark so that the words wont break when display on certain devices

fcurella:
> I meant the source is not updated or does not contains enough information. 

>  for person names we should not follow the source 100%, we can base on it though.

That's not our policy. `Faker` just uses what data is publicly available and building new datasets is out of the scope for this project. 

In very few cases we have been able to combine multiple datasets, but they were from the same agency and for the same year.

We are not going to make up any new data points. There's a lot of issues that can come out of that, selective bias being the first that comes to mind.

Ultimately, we just don't want to be responsible for what's in our data; if something comes up, we want to be able to point to the source and say "that's wherewe got it from, take the issue there".

I'd suggest you look for more and better quality datasets. Is there a page on Vietnamese Wikipedia? Does the Vietnamese Government collect and publish such data?

fcurella:
Could you get to your original source using the [wayback machine](https://web.archive.org/)?

The list on wikipedia seem to have taken from one of [these books](https://web.archive.org/web/20080131062351/http://www.minhkhai.com.vn/Search.aspx?s=H%E1%BB%8D%20V%C3%A0%20T%C3%AAn%20Ng%C6%B0%E1%BB%9Di%20Vi%E1%BB%87t%20Nam). If you could get access to a book like those, we can get get the dataset from there and using it as a source.

JRBusiness:
One thing I need to firgure it out is a way to include the Vietnamese punctuation mark without breaking the character on some device. I think they using Unicode to type in Vietnamese....

fcurella:
@JRBusiness that would work, although it does not have distributions.

I'm not seeing any `?` when opening the file in Apple Pages. Encoding is the least of my worries, and I can help you out with that.

PR #894, opened 2019-01-16
Add more french last_name

RAndre:
I need more unique name in my large set of data

### What does this changes

It adds more french name

### What was wrong

Not enough different names

### How this fixes it

Add values

Fixes #...


faker/providers/person/fr_FR/__init__.py
```diff
@@ -259,45 +259,997 @@ class Provider(PersonProvider):
...
+        'Sorel',
+        'Quenneville',
+        'Casgrain',
+        'Faure',
+        'Bourgeois',
+        'Laramée',
+        'Houde',
+        'Tremblay',
+        'Arpin',
+        'LAngelier',
```
153957: Is this a typo? should it not be `Langelier`?

153957:
It might be nicer if the names were alphabetically sorted, unless this is how they appear in the source, then it might be clearer to keep the current order.

faker/providers/person/fr_FR/__init__.py
bluesheeptoken: I think this can be both, either "L'Angelier" (without the single quote) or Langelier. Depends on the DataSource  I guess. (You might see the same "issue" on "LHiver", which can be written both ways)

PR #900, opened 2019-01-25
Update zh_TW person provider

TimeFinger:
add proportions for zh_TW person provider last_name  …
group first name by gender


tests/providers/test_person.py
```diff
@@ -251,83 +252,95 @@ def test_name_female(self):
         assert last_name in CsCZProvider.last_names_female
 
 
-class TestZhCN(unittest.TestCase):
+class TestZh(unittest.TestCase):
```
fcurella: I'd rather have separate TestCases. It makes it easier to maintain.

faker/providers/person/zh_TW/__init__.py
```diff
@@ -1,75 +1,431 @@
...
-                  "歐", "殳", "沃", "利", "蔚", "越", "夔", "隆",
-                  "師", "鞏", "厙", "聶", "晁", "勾", "敖", "融",
-                  "冷", "訾", "辛", "闞", "那", "簡", "饒",
-                  "曾", "毋", "沙", "養", "鞠", "須", "豐",
-                  "巢", "關", "蒯", "相", "查", "後", "荊", "紅",
-                  "游", "竺", "權", "逮", "盍", "益", "桓", "公",
-                  "司馬", "上官", "歐陽", "東方", "公羊",
-                  "公冶", "淳于", "公孫", "慕容", "司徒", "司空")
+    # From https://zh.wikipedia.org/wiki/%E4%B8%AD%E5%9B%BD%E5%A7%93%E6%B0%8F%E6%8E%92%E5%90%8D
+    last_names = OrderedDict((
```
fcurella: This appear to be the same list as https://github.com/joke2k/faker/blob/master/faker/providers/person/zh_CN/__init__.py#L29. You should use data specific to Taiwan for the Taiwanese provider.

TimeFinger: The data source is same, because the data source contains the last name data of all Chinese regions (China Mainland, Taiwan, Hong Kong and Macao ).
I just translated Simplified Chinese into Traditional Chinese

tests/providers/test_person.py
fcurella: Can you split the test cases in `TestZhCN` and `TestZhTW`? I realize it would result in duplicated code, but it will make it easier to maintain if the two locales will ever diverge in the future.

TimeFinger: ok

fcurella:
thank you! ✨

PR #905, opened 2019-02-05
exclude unnecessary files from the distribution

fcurella:


frispete:
Coming from the packaging front, I don't think, that exclding the tests is such a brilliant idea. 

Keeping it gives you a lot more regression tests in real world scenarios (production environments) at no extra cost, apart from delivering a few more files. 

At openSUSE, we try hard to exercise that tests on all builds, that provide tests.

BTW: I came here to complain about delivering stray __pycache__ directories in the current source tarball, because that results in some issues in the test rig...

https://build.opensuse.org/package/rdiff/home:frispete:python/python-Faker?linkrev=base&rev=2

fcurella:
Thank you @frispete !

I made a PR at https://github.com/joke2k/faker/pull/917/files. Would you mind take a look?

PR #908, opened 2019-02-08
Split formats into formats_male and formats_female

petro-zdebskyi:
### What does this changes

It splitted field `formats` into filelds `formats_male` and `formats_female`

### What was wrong

Originally methods `name_male()` and `name_female()` works uncorrect for German names, they works as same as `name()` method and returns men and women names randomly. This is caused because methods trying to find `formats_male` and `formats_female` fields and uses default `format` field if there are no such fields. For example method  `name_male()` form  _faker/providers/person/\_\_init\_\_.py_:

```   
def name_male(self):
        if hasattr(self, 'formats_male'):
            formats = self.formats_male
        else:
            formats = self.formats
        pattern = self.random_element(formats)
```

### How this fixes it

Methods `name_male()` and `name_female()` now can use fields `formats_male` and `formats_female` respectively.


faker/providers/person/de_DE/__init__.py
```diff
@@ -5,25 +5,30 @@
 
 
 class Provider(PersonProvider):
-    formats = (
+    formats_male = (
         '{{first_name_male}} {{last_name}}',
         '{{first_name_male}} {{last_name}}',
         '{{first_name_male}} {{last_name}}',
         '{{first_name_male}} {{last_name}}',
-        '{{first_name_male}} {{last_name}}-{{last_name}}',
```
fcurella: Are  hyphenated last names not used in Germany?

faker/providers/person/de_DE/__init__.py
```diff
@@ -5,25 +5,30 @@
...
+        '{{first_name_male}} {{last_name}}',
+        '{{prefix_male}} {{first_name_male}} {{last_name}}',
+        '{{first_name_male}} {{last_name}} {{suffix_male}}',
+        '{{prefix_male}} {{first_name_male}} {{last_name}} {{suffix_male}}',
+    )
+
```
fcurella: this line has some whitespace. Could you remove it so it can pass the linter?

fcurella:
Thank you! ✨

PR #910, opened 2019-02-13
Fix fr_FR postcodes_format

canarduck:
### What does this changes

It fixes the way french postcodes are formatted

### What was wrong

French postcodes should never be formatted ## ####
see https://datanova.laposte.fr/explore/dataset/laposte_hexasmal/table/
The "Official french database of postal codes"

### How this fixes it

Only ##### is used to format french postcodes


canarduck:
It looks like the pipelines failures are not related to my (tiny) change but to [this issue](https://github.com/erikrose/more-itertools/issues/272) in more-itertools and pytest ?

PR #911, opened 2019-02-13
Pin more-itertools to a version compatible with py2.7

canarduck:
### What does this changes

Force setup.py to install a version of more-itertools compatible with python 2.7

### What was wrong

Current test suite is failing for python 2.7, due to the fact that more_itertools (installed by pytest) dropped support for it, see https://github.com/erikrose/more-itertools/issues/253

### How this fixes it

Add `more-itertools<6.0.0` to `tests_require` in faker's setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,6 +73,7 @@ setup(
         "ukpostcodeparser>=1.1.1",
         "mock",
         "pytest>=3.8.0,<3.9",
+        "more-itertools<6.0.0",
     ],
     extras_require={
         ':python_version=="2.7"': [


fcurella:
Thank you!

mmlb:
Is there a nice way to specify `<6.0.0` if python version < 3 (similar to ipaddresses dep) but keep `more-itertools` for python3 versions? This came up in https://github.com/NixOS/nixpkgs/pull/61110, where I worked around it, but it'd be nice to avoid continued hacks.

canarduck:
Yes @mmlb we could replace `"more-itertools<6.0.0",` dependency by `"more-itertools<6.0.0 ; python_version< '3.0'",` did a quick test and it works.

Btw it's called Environment Markers [PEP 496](https://www.python.org/dev/peps/pep-0496/), I did not know about this before, so thanks for asking !

Making a new PR for this.

mmlb:
nice! :tada:

PR #913, opened 2019-02-18
Locale/master/add sa phone provider

mabrikan:
### What does this changes

* Adds phone provider for _ar_SA_ locale.
* Adds weightings for names formats in person provider in _ar_AA_ locale.

### DIFF
--- a/faker/providers/person/ar_AA/__init__.py
+++ b/faker/providers/person/ar_AA/__init__.py
@@ -1,20 +1,23 @@
 # coding=utf-8
 from __future__ import unicode_literals
+from collections import OrderedDict
+
 from .. import Provider as PersonProvider
 
 
 class Provider(PersonProvider):
-    formats_female = (
-        '{{first_name_female}} {{last_name}}',
-        '{{prefix_female}} {{first_name_female}} {{last_name}}',
-    )
+    formats_female = OrderedDict((
+        ('{{first_name_female}} {{last_name}}', 0.98),
+        ('{{prefix_female}} {{first_name_female}} {{last_name}}', 0.02),
+    ))
 
-    formats_male = (
-        '{{first_name_male}} {{last_name}}',
-        '{{prefix_male}} {{first_name_male}} {{last_name}}',
-    )
+    formats_male = OrderedDict((
+        ('{{first_name_male}} {{last_name}}', 0.98),
+        ('{{prefix_male}} {{first_name_male}} {{last_name}}', 0.02),
+    ))
 
-    formats = formats_male + formats_female
+    formats = formats_male.copy()
+    formats.update(formats_female)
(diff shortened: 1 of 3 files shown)


faker/providers/person/ar_AA/__init__.py
```diff
@@ -1,20 +1,23 @@
...
 
 
 class Provider(PersonProvider):
-    formats_female = (
-        '{{first_name_female}} {{last_name}}',
-        '{{prefix_female}} {{first_name_female}} {{last_name}}',
-    )
+    formats_female = OrderedDict((
+        ('{{first_name_female}} {{last_name}}', 0.98),
+        ('{{prefix_female}} {{first_name_female}} {{last_name}}', 0.02),
```
fcurella: is there any statistical basis for the assumption that 2% of people will use a prefix?

faker/providers/phone_number/ar_SA/__init__.py
```diff
@@ -0,0 +1,113 @@
+# coding=utf-8
+from __future__ import unicode_literals
+import re
+
+from .. import Provider as PhoneNumberProvider
+
+
+class Provider(PhoneNumberProvider):
```
fcurella: can you add unit tests for the new provider?

faker/providers/person/ar_AA/__init__.py
mabrikan: No, unfortunately there were none. However, since `formats_female`, `formats_male` and `formats` are all used in their corresponding _name_ functions, that is `name_male`, `name_female` and `name`, it is assumed that a person's name, most of the time, indicates his/her first and last names only, without the prefix nor the suffix (if available). Also, similar implementation was found in _en_US_ person provider.

faker/providers/phone_number/ar_SA/__init__.py
mabrikan: Will do.

faker/providers/person/ar_AA/__init__.py
fcurella: Our [policy](https://github.com/joke2k/faker/blob/master/docs/coding_style.rst#data-sets) is to introduce new data only based on statistical evidence. Some provider don't have a source for the distribution of prefixes because they didn't have it before the introduction of this policy.

I would love to have default distributions of names vs names and prefix, but I don't know what the most common distribution is among countries. Just assuming it's like the `en_US` seems a little too biased.

mabrikan: Fair point. Should I revert that commit then?

fcurella: Yeah. I'm sorry, I wish there was better data available.

PR #914, opened 2019-02-20
Fix the code of YellowGreen color

hovikman:
### What does this changes

This changes the code of YellowGreen color from '9ACD3' to '9ACD32'.

### What was wrong

The code of YellowGreen color was wrong for en-US, ar_PS, hr_HR and ru_RU locales. For other locales it is correct.


fcurella:
Thank you! ✨

Issue #366, opened 2016-07-22
TypeError: pystruct() got multiple values for keyword argument 'count' ,  what is the correct value_types?  


eromoe:
example doesn't show the value_types
http://faker.readthedocs.io/en/master/providers/faker.providers.python.html

I want generate a ramdon json with all key and value is string.
But only `fake.pystruct(count=10)` work 

I look into the provider.python and find:

```
if not value_types:
    value_types = ['str', 'str', 'str', 'str', 'float', 'int', 'int', 'decimal', 'date_time', 'uri', 'email']
```

But still error

```
>>>value_types = ['str', 'str', 'str', 'str', 'float', 'int', 'int', 'decimal', 'date_time', 'uri', 'email']
>>>fake.pystruct(count=10, *value_types)
TypeError: pystruct() got multiple values for keyword argument 'count'
```

This `fake.pystruct(10, 'str')` work,  you should change the doc with a correct

ThibaultLemaire:
:+1: for documenting this. The valid values for `value_types` are nowhere documented.

My own testing yielded that you can pass built-in types (`int`, `str`, `float`, etc. without `""`) or names of providers (e.g. `'email'`)

PR #916, opened 2019-02-28
Close #525. Make `value_types` a keyword argument

### LINKED ISSUE #525: python providers should use value_types as a keyword argument
Some of the python providers use a slightly unusual function call signature like this:

```
fake.pydict(nb_elements=10, variable_nb_elements=True, *value_types)
fake.pylist(nb_elements=10, variable_nb_elements=True, *value_types)

```
This leads to some usage problems with factory_boy (https://github.com/FactoryBoy/factory_boy/issues/387).

IMHO it would be much more convenient to be able to call the python providers with `value_types` as a keyword argument like so:

```
fake.pydict(nb_elements=10, variable_nb_elements=True, value_types=['str'])
fake.pylist(nb_elements=10, variable_nb_elements=True, value_types=['str'])

```

I know this would introduce a braking change to the python provider api, but if this is acceptable I can try to provide a PR.

fcurella:
### What does this changes

In the `python` provider, `value_types` arguments are changed to be non-variadic.

An additional `allowed_types` variadic argument has been added to achieve backward compatibility.  The deprecation plan is:
1. on the next major version, promote the `PendingDeprecationWarning` to a `DeprecationWarning`
2. on the second next major version, remove `allowed_types` completely.

### What was wrong

* The method signatures are ambiguous and non intuitive. #366
* Tools like `factoryboy` can't easily integrate with Faker. #525, https://github.com/FactoryBoy/factory_boy/pull/456

### How this fixes it

It is now possible to pass the value types either as multiple arguments, or as a single argument which could be a list or tuple.

Fixes #525

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -3,14 +3,32 @@
 from __future__ import unicode_literals
 
 from decimal import Decimal
-import sys
 
 import six
+import sys
+import warnings
 
 from .. import BaseProvider
 
 
 class Provider(BaseProvider):
+    default_value_types = (
+        'str', 'str', 'str', 'str', 'float', 'int', 'int', 'decimal',
+        'date_time', 'uri', 'email',
+    )
+
+    def _check_signature(self, value_types, allowed_types):
+        if value_types is not None and not isinstance(value_types, (list, tuple)):
+            value_types = [value_types]
+            warnings.warn(
+                'Passing value types as positional arguments is going to be '
+                'deprecated.  Pass them as a list or tuple instead.',
+                PendingDeprecationWarning,
+            )
+        if value_types is None:
+            value_types = ()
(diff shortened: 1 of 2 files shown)


swehba:
Why are you not using the asserts your class inherits from `unitest` instead of calling the plain ol' Python `assert` statement? In other words, shouldn't you be doing `self.assertEqual(expected, actual)` instead of `assert expected == actual`?

fcurella:
@swehba IThre isn't really a technical reason, it's just a stylistic choice we made for our tests

jrheard:
This PR seems like a good idea to me! Are you still planning on merging it?

slafs:
I've rebased this PR against current `master` and opened a new PR - see https://github.com/joke2k/faker/pull/1143

PR #917, opened 2019-03-05
Exclude cache files and re-include tests for packaging

fcurella:


frispete:
Thanks @fcurella for your prompt support.

Yes, the patch looks good to me. I would keep CONTRIBUTING.rst and/or (at least) CHANGELOG.rst as well.

Will attempt a package build, when you committed this.

frispete:
Hi @fcurella, git as of now packages fine, here, including the tests. Great.

Thanks again for the quick response and care. Much appreciated.

frispete:
BTW, it looks, like faker's tests run fine with pytest 3.10.1. We actively remove the upper limit requirement. Thanks god, we're not in ruby land...

PR #919, opened 2019-03-08
Brazilian RG Feature

### LINKED ISSUE #918: Brazilian RG (identity card)
Add Generator to Brazilian RG (identity card)

### Steps to reproduce
fake = Faker('pt_Br')
fake.rg()

### Expected behavior
return like this rules:
https://www.ngmatematica.com/2014/02/como-determinar-o-digito-verificador-do.html
8 digits + 1 checksum digit
### Actual behavior
New feature

davizucon:
### What does this changes
Address Brazilian RG (identity card)

### What was wrong
New feature

### How this fixes it

Fixes  #918

### DIFF
--- a/tests/providers/test_ssn.py
+++ b/tests/providers/test_ssn.py
@@ -423,6 +423,14 @@ class TestPtBR(unittest.TestCase):
         for _ in range(100):
             assert re.search(r'\d{3}\.\d{3}\.\d{3}-\d{2}', self.factory.cpf())
 
+    def test_pt_BR_rg(self):
+        for _ in range(100):
+            to_test = self.factory.rg()
+            if 'X' in to_test:
+                assert re.search(r'^\d{8}X', to_test)
+            else:
+                assert re.search(r'^\d{9}$', to_test)
+
 
 class TestNlNL(unittest.TestCase):
     def setUp(self):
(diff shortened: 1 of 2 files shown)


tests/providers/test_ssn.py
```diff
@@ -423,6 +423,14 @@ def test_pt_BR_cpf(self):
         for _ in range(100):
             assert re.search(r'\d{3}\.\d{3}\.\d{3}-\d{2}', self.factory.cpf())
 
+    def test_pt_BR_rg(self):
+        for _ in range(100):
```
fcurella: I know there's plenty of existing test using this loop pattern, but we're trying to move away from it.

A better way is to use `factory.seed` to have the provider generate known values. Something like:
```python
    def test_pt_BR_rg(self):
        self.factory.seed(0)
        to_test = self.factory.rg()
        assert to_test.endswith('X')

        self.factory.seed(1)
        to_test = self.factory.rg()
        assert to_test[-1] not in ('X', 0)

        self.factory.seed(5)
        to_test = self.factory.rg()
        assert to_test.endswith('0')
```

fcurella: Ignore my comment. Testing with `factory.seed` is not currently that easy. I have a PR up that would make it easier, but it shouldn't be on you.

fcurella:
Thank you! ✨

PR #924, opened 2019-03-13
Add min and max values for pyfloat and pydecimal

### LINKED ISSUE #870: Enables min and max values for pydecimal
Currently is not possible to set min or max values to `pydecimal` or `pyfloat`. It would be nice if we could pass these parameters.

If it makes senses I can open a PR.

Lrcezimbra:
Fixes #870


fcurella:
Thank you! ✨

PR #926, opened 2019-03-15
Request ipaddress on python 3.2/3.1/3.0 too

scarabeusiv:
Even tho not directly supported it matches up reality of where the ipaddress is really required rather than 2.7 only

In reality this is no-op change, I just need to change all the == in the setup.py prior distributing it to >= as we are not hardcoding versions in distribution and this makes my life easier.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,7 +78,7 @@ setup(
         "freezegun==0.3.11",
     ],
     extras_require={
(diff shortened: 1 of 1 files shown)


fcurella:
Trivial enough to merge, but we're not going to officially support Python 3.2/3.1/3.0. We just don't have the necessary resources.

Issue #891, opened 2019-01-14
packing with pyinstaller problems


vin7ent:
When I use pyinstaller to  pack the file to exe, it turns out the problem:
ImportError: No module named providers.

Then I added the hidden imports parameters, errors changes to
Invalid configuration for faker locale 'en_US'

So how to fix it?

arossert:
I have the same issue, there are some debaites regarding this on the PyInstaller github but I have not been able to make it work yet.
You can follow it here [#3465](https://github.com/pyinstaller/pyinstaller/pull/3465) and here [#3839](https://github.com/pyinstaller/pyinstaller/issues/3839).

It seems that there is a code in faker that check if the code is "frozen" or not.

Any help with making this work will be higly appriciated.

fcurella:
@arossert I've created a branch at https://github.com/joke2k/faker/tree/hardcode-locales that removes the dynamic loading of some values. There's still some dynamic loading happening, but hopefully hardcoding the values in `config` is enough to work. Could you give it a try?

arossert:
@fcurella Thanks, I have tested your branch and it is working with this hook:
```
from PyInstaller.utils.hooks import collect_submodules, collect_data_files

hiddenimports = collect_submodules('faker.providers')
datas = collect_data_files('text_unidecode')
```
I'm not sure that the creator will want to change this values to be hardcoded, if so it will make the like much easier.

Following this approch #704, we might need some code to detect the useage of PyInstaller (maybe check for `sys._MEIPASS` attribute) and look for the files relative to it's path, then in the PyInstaller hook we can include the src files.

arossert:
@joke2k Any update regarding this? I really need to pack faker with my project using PyInstaller and don't want to use a branch.

fcurella:
@arossert there's a problem with my PR.

As it's written, it would break some existing feature, specifically the command-line option to include custom providers.

The PR is a good starting point, but it needs some more work, an my time is limited. Any help would be appreciated :)

arossert:
@fcurella thanks, I would love to help with this but don't know how to make it work.
I have tried to create a hook in PyInstaller for that and got some progress until I realized that `pkgutil.iter_modules()` is not supported in the frozen version.
Is there another way that you know to load the package list dynamically? In my hook, all the providers code is available so I can iterate the directory tree.

arossert:
@fcurella I really need this to work, is my changes seem reasonable?

fcurella:
[...]

PR #927, opened 2019-03-18
Fix PyInstaller issues

arossert:
Following the discussion in #891, This PR is listing the modules directly from the filesystem.
 
According to this thread [#1905](https://github.com/pyinstaller/pyinstaller/issues/1905), PyInstaller is not supporting the `pkgutil.iter_modules` so the dynamic module loading is not working.

I have added a code to list the modules directly from the source code that needs to be included in the bundle for this to work.
It is using a small PyInstaller hook to fetch the providers source files:
```
from PyInstaller.utils.hooks import collect_submodules, collect_data_files

hiddenimports = collect_submodules('faker.providers')
datas = (collect_data_files('text_unidecode') +
         collect_data_files('faker.providers', include_py_files=True))
```
If this PR will be approved, I will submit a PR in PyInstaller to include this hook.

It is not working for me using cx_freeze, I have no experience with it and I could not get it to work, always give me an exception:
```
ImportError: No module named providers
```
From a small test, it seems that `pkgutil.iter_modules` is working in cx_freeze but I still got this error (even when I try to explicitly include the module).

I think that it will be better to find a generic way for all the "code freezers" work but not sure that it is possible.

### DIFF
--- a/faker/utils/loading.py
+++ b/faker/utils/loading.py
@@ -7,8 +7,15 @@ import pkgutil
 def get_path(module):
     if getattr(sys, 'frozen', False):
         # frozen
-        base_dir = os.path.dirname(sys.executable)
-        lib_dir = os.path.join(base_dir, "lib")
+
+        if getattr(sys, '_MEIPASS', False):
+            # PyInstaller
+            lib_dir = getattr(sys, '_MEIPASS')
+        else:
+            # others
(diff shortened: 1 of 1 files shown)


fcurella:
Thank you @arossert 

I wouldn't worry too much about `cx_freeze`. From what I see, your change should be fully backward compatible.

arossert:
Thanks

PR #933, opened 2019-03-26
Fix some invalid French phone numbers starting with "+33 8x"

stephane:
The full range of phone number starting with 08 isn't valid (eg. 083xx is not valid) so this patch limits the range to valid numbers only.

French validator used by the phonenumbers project:
https://github.com/daviddrysdale/python-phonenumbers/blob/dev/python/phonenumbers/data/region_FR.py#L9

### What does this changes

Limits the possible French phone numbers.

### What was wrong

+3383xxx isn't valid.

### How this fixes it

Patch provided.


fcurella:
Thank you! ✨

PR #937, opened 2019-04-13
Add bounds to pyint

francoisfreitag:
### What does this changes

Allow specifying bounds for `pyint()`.

### What was wrong

`Faker().pyint()`'s range was forced to `[0, 9999]`

---

Superseds https://github.com/joke2k/faker/pull/869.


francoisfreitag:
It looks like CI error is not related to this change.

fcurella:
Thank you!

PR #938, opened 2019-04-13
Recommend running tests with tox

francoisfreitag:
### What does this changes

Recommend running tests with tox. Simplify testing instructions.

### What was wrong

Testing requires `pytest` and `freezegun`, which are not installed with instruction `pip install -e .`. 

### How this fixes it

`tox` is closer to what the CI tool does, installs dependencies automatically, runs linters and
checks the manifest.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -356,23 +356,11 @@ you pinned the version of ``Faker`` down to the patch number.
 Tests
 -----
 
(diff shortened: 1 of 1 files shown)


fcurella:
Thank you! ✨

PR #940, opened 2019-04-13
Accept step argument in random_int()

francoisfreitag:
### What does this changes

Allows selecting a subset of integers in a range, for example even integers.

### What was wrong

Could not select a random even number with `pyint()`.

### How this fixes it

Accept step argument in random int


fcurella:
Thank you! ✨

fabien-michel:
Keyword args (min, and max) are not consistent with pyfloat and pydecimal (min_value and max_value)...

francoisfreitag:
That's right, I'll look into harmonizing these.

francoisfreitag:
> Keyword args (min, and max) are not consistent with pyfloat and pydecimal (min_value and max_value)

https://github.com/joke2k/faker/pull/957

PR #941, opened 2019-04-13
Add basic tests for pyint()

francoisfreitag:
### What does this changes

Add tests for `pyint()`.

### What was wrong

That method was not tested.


fcurella:
already added in #937

PR #943, opened 2019-04-14
Add missing commas to company/nl_NL provider

francoisfreitag:
### What does this changes

Add missing commas to company/nl_NL provider.

### What was wrong

Commas were missing, resulting in two different strings being concatenated, forming nonexistent companies such as `'KruidvatVan Merksteijn Steel'`.


fcurella:
Thank you! ✨

Issue #946, opened 2019-04-21
image_placeholder_services contains dead service


Monstrofil:
Looks like https://placeholdit.imgix.net/ does not work anymore, I propose to remove it.

### Steps to reproduce

1. use fake.image_url()
2. sometimes you get url like https://placeholdit.imgix.net/~text?txtsize=55&txt=320x320&w=320&h=32

### Expected behavior

Url must be valid

### Actual behavior

Link does not work

PR #947, opened 2019-04-21
Remove dead url from image_placeholder_services

Monstrofil:
Looks like placeholdit.imgix.net is no longer available, so remove it from code.

### What does this changes
Remove dead image stubs service from code.

### What was wrong
placeholdit.imgix.net does not work anymore

Fixes: #946


fcurella:
Thank you! ✨

PR #950, opened 2019-04-28
Adds catalan, adds doi/nie/nif/cif to spain ssn

kingbuzzman:
### What does this changes

Adds Catalan language
Adds doi / nif / nie / cif

### What was wrong

- adds functionality

### How this fixes it

...


faker/providers/person/es_CA/__init__.py
```diff
@@ -0,0 +1,61 @@
+# coding=utf-8
+from __future__ import unicode_literals
+from ..es_ES import Provider as PersonProvider
+
+
+class Provider(PersonProvider):
+    first_names_male = tuple(set(PersonProvider.first_names_male + (
```
fcurella: what's the source for the dataset? Please consult our [Coding style](https://github.com/joke2k/faker/blob/master/docs/coding_style.rst#data-sets) for guidelines about data sources

kingbuzzman: @fcurella Yes, i’ve documented where i get the data from.

kingbuzzman:
I've moved all the validation (`is_nie`, `is_cif`, `is_doi`, ...) code to: https://github.com/kvesteri/validators/pull/121 -- how do you guys feel about having that as a test dependency?

faker/providers/person/es_CA/__init__.py
```diff
@@ -0,0 +1,65 @@
+# coding=utf-8
+from __future__ import unicode_literals
+from ..es_ES import Provider as PersonProvider
+
+
+class Provider(PersonProvider):
+    """
+    Adds popular Catalan names to the spanish ones.
```
fcurella: The added names come from a proper source, but by adding the Spanish one we're effectively build our own dataset, which goes outside of the scope of Faker. Could we stick to _just_ what's on the data source?

fcurella:
@kingbuzzman We can add `validators` as test dependency, but that means this PR will be blocked by your PR on `validators`.

faker/providers/person/es_CA/__init__.py
kingbuzzman: Spain is complicated from an external point of view. It has Spanish, Catalan, Galician, Basque, Occitan languages under one country. Each region has a lot of mixing and crossover, i don't think that having _JUST_ catalan names is correct, since you will find a lot of "normal" spanish names in Catalunya.

fcurella: Wouldn't such complexity be reflected in the data collected by the statistics?

kingbuzzman: Thats a VERY good point. I've amended the list using two sources: baby names, popular names in the region.

faker/providers/person/es_CA/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+    Adds popular Catalan names.
+    https://www.idescat.cat/pub/?id=aec&n=946&lang=es&t=2018
+    https://www.idescat.cat/pub/?id=aec&n=947&lang=es&t=2018
+    """
+    first_names_male = (
+        'Adam',
+        'Albert',
+        'Aleix',
+        'Àlex',
+        'Álex',
```
fcurella: this two names are treated like one and the same in the original dataset. Including them both would result in having them picked twice as often. I'd just toss a coin and pick one, unless there's one deemed more 'canonical'.

kingbuzzman: Another good point. Amended.

kingbuzzman:
@fcurella want me to squash? Or would you do that on merge?

fcurella:
@kingbuzzman I'll squash on merge. I'm just waiting on the CI to complete.

fcurella:
Thank you ✨

PR #951, opened 2019-05-08
Pin more-itertools only for python < 3.0

canarduck:
### What does this changes

Force setup.py to install a version of `more-itertools` compatible with python 2.7 when python < 3.0

### What was wrong

Nothing *wrong* but this is better than https://github.com/joke2k/faker/pull/911 because we can use a more up to date version of `more-itertools` when  python >= 3.0.
Also it looks like it was causing problems on https://github.com/NixOS/nixpkgs/pull/61110 which were reported by @mmlb https://github.com/joke2k/faker/pull/911#issuecomment-490300392 and required a workaround. Hope this can fix it properly.

### How this fixes it

It makes use of [PEP 496 -- Environment Markers](https://www.python.org/dev/peps/pep-0496/) to conditionality install a specific version of `more-itertools` when python < 3.0

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,3 +1,6 @@
+* Faker version:
+* OS:
+
 Brief summary of the issue goes here.
 
 ### Steps to reproduce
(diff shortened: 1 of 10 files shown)


fcurella:
Thank you! ✨

PR #956, opened 2019-05-13
fix(negative_float): Corrected behavior for pyfloat

### LINKED ISSUE #813: pyfloat, pydecimal cannot be negative when left_digits=0
I am trying to fake random floats in the range `(-1, 1)`.  Currently, `pyfloat` and `pydecimal` will never be negative as long as `left_digits=0`, even if `positive=False`.

This seems to be because [the `sign` (-1, 1) is multiplied by only the left digits](https://github.com/joke2k/faker/blob/6e32e07/faker/providers/python/__init__.py#L62) (which in this case is 0, canceling out the `sign`), rather than by the entire numeric value.

### Steps to reproduce

```python
>>> import faker
>>> fake = faker.Faker()
>>> any(fake.pyfloat(left_digits=0, positive=False) < 0 for _ in range(10000))
False
```

### Expected behavior

I'd expect approximately half of the floats to be negative, when `positive` is not `True`

ariksu:
### What does this changes

This change modified behavior of pyfloat and pydecimal providers on negative numbers with left_digits=0

### What was wrong

When min and max was omitted, positive was not set to True and there was no integral part negative numbers were not generated for provider.pyfloat and provider.pydecimal.

### How this fixes it

Integral part of float number was generated either positive or negative. In case where integral part had 0 digits negative integral dissapeared as -1*0==0. Separated sign from integral part.
Fixes #813

### DIFF
--- a/tests/test_factory.py
+++ b/tests/test_factory.py
@@ -467,7 +467,6 @@ class FactoryTestCase(unittest.TestCase):
         assert any(factory.pyfloat(left_digits=0, positive=False) < 0 for _ in range(100))
         assert any(factory.pydecimal(left_digits=0, positive=False) < 0 for _ in range(100))
 
-
     def test_us_ssn_valid(self):
         from faker.providers.ssn.en_US import Provider
 
--- a/tests/test_factory.py
+++ b/tests/test_factory.py
@@ -462,12 +462,11 @@ class FactoryTestCase(unittest.TestCase):
 
     def test_negative_pyfloat(self):
         # tests for https://github.com/joke2k/faker/issues/813
-        from faker.providers.python import Provider
-        provider = Provider(self.generator)
+        factory = Faker()
+        factory.seed_instance(32167)
+        assert any(factory.pyfloat(left_digits=0, positive=False) < 0 for _ in range(100))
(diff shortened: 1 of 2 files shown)


ariksu:
Frankly, I have no idea why coverage reduced. Looks like there are some uncovered or luck-based case for domain testing and cli exceptions. I believe that some additional tests would fix that, but those are out of scope for this commit.

tests/test_factory.py
```diff
@@ -460,6 +460,15 @@ def test_random_pyfloat(self):
         with pytest.raises(ValueError):
             provider.pyfloat(left_digits=0, right_digits=0)
 
+    def test_negative_pyfloat(self):
+        # tests for https://github.com/joke2k/faker/issues/813
+        from faker.providers.python import Provider
+        provider = Provider(self.generator)
+
+        assert any(provider.pyfloat(left_digits=0, positive=False) < 0 for _ in range(100))
```
fcurella: can you make these tests deterministic by using [`Provider.seed_instance()`](https://github.com/joke2k/faker#seeding-the-generator)?

ariksu: I can and will. I omitted that because the probability of failure is somewhere around 2^-99.

fcurella:
Thank you! ✨

PR #957, opened 2019-05-16
Rename pyint min and max to min_value and max_value

francoisfreitag:
### What does this changes

Rename `pyint`'s `min` and `max` arguments to `min_value` and `max_value`.

### What was wrong

Other generators `pyfloat` and `pydecimal` already name their arguments `min_value` and `max_value`, `pyint` should follow that convention.


fcurella:
Thank you! ✨

Issue #961, opened 2019-05-22
No function to return an invalid US format SSN.


darrylwhiting:
* Faker version: v0.8.17+
* OS: N/A

When an invalid en_US format SSN is desired, there is no function to return an invalid en_US format SSN.

### Steps to reproduce

1. N/A

### Expected behavior

Along with generating valid United States' EIN, ITIN, and SSNs, there should be a function to generate an invalid US format SSN, which is also not a valid ITIN. An invalid SSN assist teams with negative testing for applications or code which consume en_US format SSNs.

### Actual behavior

No function for returning an invalid en_US format SSN is available.

malefice:
@fcurella it looks like this can be closed.

PR #962, opened 2019-05-22
Added an Invalid SSN generator to the en_US SSN Provider for use case…

darrylwhiting:
…s where an Invalid en_US format SSN is needed.

### What does this change

In use cases where a genuine United States' en_US format SSN may pull a credit report, an invalid SSN can be substituted, preventing a live credit pull. Additionally, by ensuring many thousands of random invalid SSNs can be generated,  this can reduce or eliminate reusing the same invalid SSN during data driven testing, or development purposes.

### What was wrong

No Invalid United States en_US format SSN function.

### How this fixes it

Adds "invalid_ssn" en_US format SSN function to en_US Provider, which returns an invalid SSN.

Fixes # [961](https://github.com/joke2k/faker/issues/961).

### DIFF
--- a/tests/providers/test_ssn.py
+++ b/tests/providers/test_ssn.py
@@ -114,83 +114,94 @@ class TestEnUS(unittest.TestCase):
             assert area != '666'
 
     def test_invalid_ssn(self):
-        self.factory.seed(19030)
-        for _ in range(100):
-            ssn = self.factory.ssn(taxpayer_identification_number_type='INVALID_SSN')
+        # Magic Numbers used generate '666-44-9370', '000-19-8911', '978-45-0926', '241-00-1691',
+        # and 812-11-0000 respectively:
+        #
+        # Ensure that generated SSNs are 11 characters long
+        # including dashes, consist of dashes and digits only, and the tested number
+        # violates the requirements below, ensuring an INVALID SSN is returned:
+        #
+        # A United States Social Security Number
+        # (SSN) is a tax processing number issued by the Internal
+        # Revenue Service with the format "AAA-GG-SSSS".  The
+        # number is divided into three parts: the first three
+        # digits, known as the area number because they were
+        # formerly assigned by geographical region; the middle two
+        # digits, known as the group number; and the final four
+        # digits, known as the serial number. SSNs with the
+        # following characteristics are not allocated:
+        #
+        # 1) Numbers with all zeros in any digit group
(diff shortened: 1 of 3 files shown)


tests/providers/test_ssn.py
```diff
@@ -113,6 +113,85 @@ def test_ssn(self):
             assert 1 <= int(serial) <= 9999
             assert area != '666'
 
+    def test_invalid_ssn(self):
+        self.factory.seed(19030)
+        for _ in range(100):
```
fcurella: after setting `self.factory.seed()`, `self.factory.ssn` should always return the same value. I don't think this loop is necessary.

darrylwhiting: I can set the seed() to 2432 to return an SSN that begins with 666, but removing the loop then tests only one SSN, and many asserts are not processed. Removing the seed(), and keeping the loop allows for 100 random Invalid SSNs to be tested, similar to the "test_ssn" unittest. Is that an acceptable approach?

fcurella: I'd rather set the seed multiple times for the specific assertions. Something like:

```python
self.factory.seed_instance(242)
ssn = self.factory.ssn(taxpayer_identification_number_type='INVALID_SSN')

assert ssn.startswith('666')

self.factory.seed(19030)
ssn = self.factory.ssn(taxpayer_identification_number_type='INVALID_SSN')
# assert something else
```

darrylwhiting: > I'd rather set the seed multiple times for the specific assertions. Something like:
> 
> ```python
> self.factory.seed_instance(242)
> ssn = self.factory.ssn(taxpayer_identification_number_type='INVALID_SSN')
> 
> assert ssn.startswith('666')
> 
> self.factory.seed_instance(19030)
> ssn = self.factory.ssn(taxpayer_identification_number_type='INVALID_SSN')
> # assert something else
> ```

I believe I have corrected the assertions as shown above:

[New Commit](https://github.com/darrylwhiting/faker/commit/3dc30c69da300f90b6bbb4db2eacd06c8f771733)

fcurella:
The build is failing because the random module on Python 3 outputs a different value than on Python 2 for the same seed.

You'll need to import the `random2` module (it's already in the `test_requirements` and set the factory to use it. Here's an example: https://github.com/joke2k/faker/blob/master/tests/providers/test_ssn.py#L315

fcurella:
Thank you! ✨

Issue #944, opened 2019-04-15
Inconsistency in unix_time handling start and end time parameters


navtej:
There is an inconsistency the way Faker handles `unix_time()` when parameter `start_datetime` is provided as 'now'. `start_datetime` is parsed using class method `_parse_start_datetime` 
https://github.com/joke2k/faker/blob/6db5503cf070c0b09da8e98cd662c10230cea22f/faker/providers/date_time/__init__.py#L1374

which is eventually parsed by `_parse_date_time`, which can take argument 'now'. When 'now' is the provided,  `datetime.now` , which will use local time,
https://github.com/joke2k/faker/blob/6db5503cf070c0b09da8e98cd662c10230cea22f/faker/providers/date_time/__init__.py#L1525

However, end time parameter is parsed by `_parse_end_datetime`, which uses `time.time()`. `time.time()` returns time since epoc ie it is utc timestamp. 
https://github.com/joke2k/faker/blob/6db5503cf070c0b09da8e98cd662c10230cea22f/faker/providers/date_time/__init__.py#L1477

on machines with negetive time offset, faker will run into problems because of this inconsistancy. Example
```
>>> import faker
>>> f = faker.Faker()
>>> f.unix_time(start_datetime='-6h')
1555352053
>>> f.unix_time(start_datetime='now')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/dist-packages/faker/providers/date_time/__init__.py", line 1376, in unix_time
    return self.generator.random.randint(start_datetime, end_datetime)
  File "/usr/lib/python3.7/random.py", line 222, in randint
    return self.randrange(a, b+1)
  File "/usr/lib/python3.7/random.py", line 200, in randrange
    raise ValueError("empty range for randrange() (%d,%d, %d)" % (istart, istop, width))
*ValueError: empty range for randrange() (1555373558,1555353759, -19799)*
>>>

```

Perhaps 'now' should be `utcnow` here
https://github.com/joke2k/faker/blob/6db5503cf070c0b09da8e98cd662c10230cea22f/faker/providers/date_time/__init__.py#L1525

fcurella:
Thank you for the report @navtej !

Would you mind submit a Pull Request with a failing test case?

erayerdin:
I currently run across problems using `future_datetime` and `past_datetime` providers. Might be related to this.

[...]

PR #964, opened 2019-05-27
add pwz generator and date_of_brith and sex asargument to pesel (`pl_PL`)

tomaszrycz:
What does this changes

- `Person` in `pl_PL`

What was wrong

- it wasn't possible to generate pesel number for given date of birth and/or sex
- generator for a `pwz` number was unavailable 

How this fixes it

- adds pwz_doctor and pwz_nurse generator to `Person` (pl_PL)
- adds date_of_birth and sex as parameter to `Person.pesel` (pl_PL)

Fixes #931

PR #965, opened 2019-05-29
Issue/944/throw error in unix time generation

bluesheeptoken:
### What does this changes

Replace the `time.time()` by `datetime_to_timestamp(datetime.now())` to avoid problems with negative offsets in machine as specified in #944

### What was wrong

These lines of code used to throw an error
```python
import faker
f = faker.Faker()
f.unix_time(start_datetime='now')
```

### How to fix it
Changed the calculation with `end_datetime=None`. Added a test to set an offset during tests to avoid this :bug: to happen again.

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -6,7 +6,6 @@ import re
 
 from calendar import timegm
 from datetime import timedelta, MAXYEAR
-from time import time
 
 from dateutil import relativedelta
 from dateutil.tz import tzlocal, tzutc
@@ -1474,7 +1473,7 @@ class Provider(BaseProvider):
     @classmethod
     def _parse_end_datetime(cls, value):
         if value is None:
-            return int(time())
+            return datetime_to_timestamp(datetime.now())
 
         return cls._parse_date_time(value)
 
--- a/tests/providers/test_date_time.py
+++ b/tests/providers/test_date_time.py
@@ -3,10 +3,13 @@ from __future__ import unicode_literals
 
 from datetime import date, datetime, timedelta, tzinfo
 from datetime import time as datetime_time
-import time
-import unittest
+import os
+import platform
+import pytest
 import random
 import sys
+import time
+import unittest
 
 import six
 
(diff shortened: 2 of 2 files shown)


fcurella:
Have you tried setting `os.environ["TZ"]` temporarily to a negative offset as suggested in https://stackoverflow.com/a/1301528 ?

bluesheeptoken:
@fcurella Thanks for pointing out, I have read the documentation too fast.

This is fixed, however, I do not know what I messed up for the appveyor integration... I might need some help here.

bluesheeptoken:
@fcurella 
I am strill struggling with the appveyor integration with python3.5. Apparently, he cannot find [importlib-metadata](https://ci.appveyor.com/project/joke2k/faker/builds/25000045/job/ukj46n9hlcgacuej#L261) which is required by pluggy. Pluggy is required by Pytest.

Any idea where can this come from? Any idea how to fix that?

tomaszrycz:
Hi @bluesheeptoken 

I have the same issue with my pull request #964.
I've analysed the problem and I think it can be something wrong in  `importlib-metadata` building setup. There is a warning in log:
```
C:\Python35\lib\distutils\dist.py:261: UserWarning: Unknown distribution option: 'use_scm_version'
  warnings.warn(msg)
```
and then builder is copying a file with a strange name 
```
Copying unknown-0.0.0-py3.5.egg to c:\projects\faker\.eggs
```
In my opinion ` unknown-0.0.0-py3.5.egg` is actually `importlib-metadata` package but due to wrong name `pkg_resources` can't find it.

Unfortunately, I don't know how to resolve this problem.

bluesheeptoken:
Hi @tomaszrycz, thanks for telling me !

I arrived at the same conclusion... (But I did not know about pkg_resources!)

I tried to check pytest's issues to see if it came from here. But did not find anything either... 
I will investigate again tomorrow and keep you in touch !

Thanks for replying !

fcurella:
@bluesheeptoken appveyor is fixed on master. Do you mind rebasing your branch?

bluesheeptoken:
@fcurella , thank you this is done.
No idea why coverage reacts like this though

fcurella:
@bluesheeptoken don't worry about the coverage. It's flaky because some tests are non-deterministic.

Thank you so much! ✨

PR #968, opened 2019-06-11
Added SSN provider for es_MX

### LINKED ISSUE #967: Missing logic for generating Mexican SSN
The format for the Mexican SSN should be 11 digits:

* The first two digits represent the office where the person was registered.
* The next two digits are the year when the person was registered to social security.
* The next two digits are the year of birth for the person.
* The next four digits are a sequential number assigned by the registration office.
* The last digit is a checksum.

### Steps to reproduce

```python
from faker import Faker
fake = Faker('es_MX')
print(fake.ssn()  # 105-90-6991)
```
The [docs](https://faker.readthedocs.io/en/master/locales/es_MX.html#faker-providers-ssn) reflect this error.

### Expected behavior

The SSN for locale `"es_MX"` should follow the format described above.

```python
from faker import Faker
fake = Faker('es_MX')
print(fake.ssn()  # 92919084431)
```

### Actual behavior

The SSN format for `"es_MX"` is currently `'##-##-####'`.

mrfunnyshoes:
### What does this changes

Added a Mexican SSN provider, along with Unique Population Registry Code (CURP) and Federal Taxpayer Registry ID (RFC).

### What was wrong

The was no provider for these numbers numbers / codes.

### How this fixes it

This adds providers for SSN, CURP and RFC (both for natural and legal persons):

```python
from faker import Faker
fake = Faker("es_MX")
print(gen.ssn())  # '20142713988'
print(gen.curp())  # 'TOGG411122HGRPQI01'
print(gen.rfc())  # 'TEBS650328SZG'
print(gen.rfc(natural=False))  # 'UDO460509MU1'
```

Fixes #967

### DIFF
--- a/faker/providers/ssn/es_MX/__init__.py
+++ b/faker/providers/ssn/es_MX/__init__.py
@@ -8,6 +8,7 @@ from .. import Provider as BaseProvider
 
 
 ALPHABET = string.ascii_uppercase
+ALPHANUMERIC = string.digits + ALPHABET
 VOWELS = "AEIOU"
 CONSONANTS = [
     letter
@@ -209,3 +210,41 @@ class Provider(BaseProvider):
         random_curp += str(control_digit)
 
         return random_curp
+
+    def rfc(self, natural=True):
+        """
+        See https://es.wikipedia.org/wiki/Registro_Federal_de_Contribuyentes
+
+        :param natural: Whether to return the RFC of a natural person.
+            Otherwise return the RFC of a legal person.
+        :type natural: bool
+        :return: a random Mexican RFC
+        """
+        birthday = self.generator.date_of_birth()
+
+        if natural:
+            first_surname = random.choice(ALPHABET) + random.choice(VOWELS)
+            second_surname = random.choice(ALPHABET)
+            given_name = random.choice(ALPHABET)
+            name_initials = first_surname + second_surname + given_name
+        else:
+            name_initials = (
+                self.random_uppercase_letter() +
+                self.random_uppercase_letter() +
+                self.random_uppercase_letter()
+            )
+
+        birth_date = birthday.strftime("%y%m%d")
+        disambiguation_code = (
+            random.choice(ALPHANUMERIC) +
+            random.choice(ALPHANUMERIC) +
+            random.choice(ALPHANUMERIC)
+        )
+
+        random_rfc = (
+            name_initials +
+            birth_date +
+            disambiguation_code
+        )
+
+        return random_rfc
--- a/faker/providers/ssn/es_MX/__init__.py
+++ b/faker/providers/ssn/es_MX/__init__.py
@@ -195,7 +195,7 @@ class Provider(BaseProvider):
 
         # This character is assigned to avoid duplicity
         # It's normally '0' for those born < 2000
-        # and 'A' for those botn >= 2000
+        # and 'A' for those born >= 2000
         assigned_character = "0" if birthday.year < 2000 else "A"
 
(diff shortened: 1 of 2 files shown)


faker/providers/ssn/es_MX/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+ALPHABET = string.ascii_uppercase
+ALPHANUMERIC = string.digits + ALPHABET
+VOWELS = "AEIOU"
+CONSONANTS = [
+    letter
+    for letter in ALPHABET
+    if letter not in VOWELS
+]
```
fcurella: I wonder if using sets would make this faster

bluesheeptoken: If you think about `list(set(ALPHABET) - set(CONSONANTS))`, this is taking approximately 20% more time, the set conversion comes at a "huge cost" (We speak about some µs). Note, we need to convert it back to a list because we cannot call `random.choice` on sets.

faker/providers/ssn/es_MX/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+    Sum of digits of a number until sum becomes single digit.
+
+    Example:
+        658 => 6 + 5 + 8 = 19 => 1 + 9 = 10 => 1
+    """
+    if n == 0:
+        return 0
+    if n % 9 == 0:
+        return 9
+    else:
```
bluesheeptoken: Pylint does not approve the `if-else-return`. Not sure it is important.

faker/providers/ssn/es_MX/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+            birth_date +
+            gender +
+            state +
+            first_surname_inside +
+            second_surname_inside +
+            given_name_inside +
+            assigned_character
+        )
+
+        # control_digit = str(self.random_digit())
```
bluesheeptoken: Is this comment left is intended?

faker/providers/ssn/es_MX/__init__.py
fcurella: doesn't cost much to make pylint happy, let's fix it

faker/providers/ssn/es_MX/__init__.py
mrfunnyshoes: Not really, I removed it.

faker/providers/ssn/es_MX/__init__.py
mrfunnyshoes: pylint should be happier now :)

bluesheeptoken:
The only problem now is the build. I do not know what's going on. @tomaszrycz is also on it.

This might be a problem from libmetadata setup.

faker/providers/ssn/es_MX/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+        birth_date = birthday.strftime("%y%m%d")
+        gender = random.choice("HM")
+        state = random.choice(STATES_RENAPO)
+        first_surname_inside = random.choice(CONSONANTS)
+        second_surname_inside = random.choice(CONSONANTS)
+        given_name_inside = random.choice(ALPHABET)
+
+        # This character is assigned to avoid duplicity
+        # It's normally '0' for those born < 2000
+        # and 'A' for those botn >= 2000
```
bluesheeptoken: Typo here I think, this should be `born` I think

mrfunnyshoes: Thanks, fixed!

fcurella:
Thank you! ✨

Issue #970, opened 2019-06-19
Overly restrictive freezegun and random2 dependencies


timokau:
Currently freezegun and random2 are pinned to exact versions:

https://github.com/joke2k/faker/blob/c412d56e6a8f3681cfa1b75ef6ed47bc9798825d/setup.py#L77
https://github.com/joke2k/faker/blob/c412d56e6a8f3681cfa1b75ef6ed47bc9798825d/setup.py#L78

Is there any reason for that? Would there really be problems with `freezegun` 0.3.12? This kind of version pinning is hard to match for distribution packagers.

fcurella:
The reason we pin the dependencies is that we don't want to break the build in case one of them drops support for python 2.7. I think we can relax them and just pin to the minor version, so something like  `random2<1.1` and `freezegun<0.4`

timokau:
That would be helpful, as long as they adhere to something semver-like they shouldn't drop python2 support in a minor update.

fcurella:
@timokau I'm currently on vacation, so I don't have much time to work on this.

But if you could write a PR that passes the build, I should have enough time to merge and release

timokau:
Sure, https://github.com/joke2k/faker/pull/972.

No big hurry though, after you vacation is also perfectly fine.

PR #972, opened 2019-06-20
Losen version restrictions on freezegun and random2

timokau:
### What does this changes

Loosen version restrictions as discussed in https://github.com/joke2k/faker/issues/970.

### What was wrong

Unnecessarily restrictive version requirements.

### How this fixes it

Fixes https://github.com/joke2k/faker/issues/970


fcurella:
Thank you! ✨

Issue #935, opened 2019-04-10
android ios User-Agent


LooEv:
can you add  some  user-agents of android and ios?

fcurella:
@LooEv feel free to add them to https://github.com/joke2k/faker/blob/master/faker/providers/user_agent/__init__.py and submit a Pull Request :)

malefice:
@fcurella looks like this one can be closed as well.

PR #985, opened 2019-08-06
add ios and android to user_agent provider

gsilvan:
### Update to user_agent provider

This update adds Android and iOS user agent to Firefox and Chrome as requested in #935 .
There is also a small bug fixed in randomization of WebKit's version number where a dot was missing (https://github.com/joke2k/faker/commit/50b40738b70d1de2189807a1489125f9e31df18a)

### DIFF
--- a/faker/providers/user_agent/__init__.py
+++ b/faker/providers/user_agent/__init__.py
@@ -50,8 +50,8 @@ class Provider(BaseProvider):
 
     def chrome(self, version_from=13, version_to=63,
                build_from=800, build_to=899):
-        saf = str(self.generator.random.randint(531, 536)) +  \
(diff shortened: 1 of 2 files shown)


faker/providers/user_agent/__init__.py
```diff
@@ -5,6 +5,7 @@
 from datetime import datetime
 
 from .. import BaseProvider
+import string
```
fcurella: can you move this to the top of the file? (Above `from datetime import datetime)`

fcurella: it's just a stylistical convention. I like to have imports from the standard lib at the top.

fcurella:
Thank you ✨

PR #989, opened 2019-08-22
Add pytimezone for tzinfo objects

### LINKED ISSUE #988: suggestion: pytimezone
In addition to the currently available [timezone](), it would be very useful* to have a similar python object faker, something like:
```python
import pytz

class TheExistingPythonProvider:

    def pytimezone(self, **kwargs):
        return pytz.timezone(self.generator.timezone(**kwargs))
```

For an example use-case, see:

\* FactoryBoy/factory_boy#643

OJFord:
### What does this changes

Adds a python object faker around the existing `timezone` name faker.

### What was wrong

It was a bit of a pain (multiple steps) to fake dates with fake tzinfo.

### How this fixes it

It removes one of the steps!

Fixes #988 

NB:

I thought about also doing something like `tzinfo=True` (or a new kwarg) as an argument to the other datetime methods, that makes them use `pytimezone` to generate a tzinfo instead of expecting one. That's more work though, so I thought I'd leave it without agreement on the principle.

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -4,7 +4,7 @@ from calendar import timegm
 from datetime import MAXYEAR, timedelta
 
 from dateutil import relativedelta
-from dateutil.tz import tzlocal, tzutc
+from dateutil.tz import gettz, tzlocal, tzutc
 
 from faker.utils.datetime_safe import date, datetime, real_date, real_datetime
 
@@ -1961,6 +1961,17 @@ class Provider(BaseProvider):
     def timezone(self):
         return self.generator.random.choice(
             self.random_element(self.countries)['timezones'])
+
+    def pytimezone(self, *args, **kwargs):
+        """
+        Generate a random timezone (see `faker.timezone` for any args)
+        and return as a python object usable as a `tzinfo` to `datetime`
+        or other fakers.
+
+        :example faker.pytimezone()
+        :return dateutil.tz.tz.tzfile
+        """
+        return gettz(self.timezone(*args, **kwargs))
 
     def date_of_birth(self, tzinfo=None, minimum_age=0, maximum_age=115):
         """
--- a/tests/providers/test_date_time.py
+++ b/tests/providers/test_date_time.py
@@ -138,6 +138,16 @@ class TestDateTime(unittest.TestCase):
         today_back = datetime.fromtimestamp(timestamp, utc).date()
         assert today == today_back
 
+    def test_pytimezone(self):
+        import dateutil
+        pytz = self.fake.pytimezone()
+        assert isinstance(pytz, dateutil.tz.tz.tzfile)
+
+    def test_pytimezone_usable(self):
+        pytz = self.fake.pytimezone()
+        date = datetime(2000, 1, 1, tzinfo=pytz)
+        assert date.tzinfo == pytz
+
     def test_datetime_safe(self):
         from faker.utils import datetime_safe
         # test using example provided in module


faker/providers/date_time/__init__.py
```diff
@@ -1967,6 +1967,16 @@ def timezone(self):
         return self.generator.random.choice(
             self.random_element(self.countries)['timezones'])
 
+    def pytimezone(self, **kwargs):
+        """
+        Generate a random timezone object; taking the same arguments
+        as `timezone`.
+
+        :example tzfile('/usr/share/zoneinfo/Europe/London')
```
fcurella: Could we make this a little bit more user-friendly and have users pass timezones like `"Europe/London" instead of the full path?

OJFord: Sorry, I just tried to copy another - the usage simply looks like `faker.pytimezone()`, but the constructed result looks like that example. What should the example read?

The goal here is *not* to pass in a timezone, but to have faker generate a random timezone object.

fcurella: `self.timezone()` does not take any arguments, and so shouldn't `pytimezone`

The example should simply be `:example faker.pytimezone()`

OJFord: Thanks, I'll update the example. I tried to imitate the one below which was `Date('some str')` and seemed to be the `repr` of the returned object.

The idea with accepting kwargs is that if `self.timezone()` is modified for any future args (whatever they may be), no change is necessary here. I can't imagine `pytimezone` args would diverge from `timezone`'s, so to me it makes sense to tie the two together even if there's none today. Thoughts?

fcurella: >  I tried to imitate the one below 

Yeah, those are completely wrong and need to be fixed

> The idea with accepting kwargs is that if self.timezone() is modified for any future args (whatever they may be), no change is necessary here. 

As long as `kwargs` doesn't show up in the genecerated docs, I'm ok with having it. My priority is to not confuse the user.

OJFord: Thanks, I'll fix the example and check the generated docs.

OJFord: I've sorted the example, reworded the docstring to be less confusing than 'the same arguments', and updated the signature so that it matches others in the generated docs (particularly several in the internet provider).

malefice:
@fcurella Based on the error logs, it appears that the docker image's locale has not been set to something that supports `UTF-8`. To verify, I tried to replicate the environment by setting up an `i386/ubuntu` docker image, installed `locales`, and then ran the command `locale-gen en_US.UTF-8`. When I checked the default locale, none was set.

```
root@4cf92a376f00:/# cat /etc/default/locale
#  File generated by update-locale
```

The `build32bit.sh` file needs to be updated to set the locale manually. Try appending the command `update-locale LANG=en_US.UTF-8` after the `locale-gen` command.

malefice:
@fcurella @OJFord So, I finally got the time to set up travis for my fork. I applied my recommendations on top of the work done in this pull request, and the 32bit build passes. You can see it [here](https://travis-ci.com/malefice/faker/builds/136106896).

Just add `export LANG='en_US.UTF-8' in `build32bit.sh` as shown [here](https://github.com/malefice/faker/blob/89584bcc8cfb5db8ef00975ae083bf7b7b210534/build32bit.sh#L13).

fcurella:
Thank @malefice, I've committed the fix to master.

@OJFord could you rebase your branch against `master`?

OJFord:
Sorry for the delay @fcurella, sure; thanks @malefice!

fcurella:
@OJFord Could you addo some test for the new method?

OJFord:
@fcurella Anything further needed here?

fcurella:
Looks Good! Thank you! ✨

PR #990, opened 2019-08-24
Revert "Close #628. Switch to text-unidecode"

### LINKED ISSUE #628: Unidecode dependency GPL license
In commit adc1b1b a dependency on the Unidecode python package was introduced.

Unidecode is released under the GPL license while Faker is using MIT. I'm concerned that this might not be permitted by the GPL license and that users of the Faker library might not realise they are introducing GPL code into their dependencies. It could be argued that any code using the Faker library now has to be licensed under the GPL too.

See here for some opinions on using GPL libraries:

https://opensource.stackexchange.com/questions/1640/if-im-using-a-gpl-3-library-in-my-project-can-i-license-my-project-under-mit-l
https://softwareengineering.stackexchange.com/questions/87446/using-a-gplv3-python-module-will-my-entire-project-have-to-be-gplv3-licensed
https://opensource.stackexchange.com/questions/2139/can-i-license-python-project-under-3-clause-bsd-while-it-has-gpl-based-dependenc
https://opensource.stackexchange.com/questions/6062/using-gpl-library-with-mit-licensed-code

I'm not an expert in this area at all, just a concerned user!

### LINKED ISSUE #727: text-unidecode is released under the Artistic license

jorti:
This reverts commit e32a0cd73d7e1f42e12593bf6024ba7607454f54.

### What does this changes

Reverts the the use of text-unidecode to unidecode

### What was wrong

text-unidecode is under Artistic license which is not FOSS compatible

### How this fixes it

This change reverts the commit that changes to text-unidecode

Fixes #727

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -1,7 +1,7 @@
 # coding=utf-8
 from __future__ import unicode_literals
 
-from text_unidecode import unidecode
+import unidecode
 
 from .. import BaseProvider
 
@@ -123,7 +123,7 @@ class Provider(BaseProvider):
         for search, replace in self.replacements:
             string = string.replace(search, replace)
 
-        string = unidecode(string)
+        string = unidecode.unidecode(string)
         return string
 
     @lowercase
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,7 @@ setup(
     install_requires=[
         "python-dateutil>=2.4",
         "six>=1.10",
-        "text-unidecode==1.2",
+        "unidecode",
     ],
     tests_require=[
         "validators>=0.13.0",


fcurella:
Hi @jorti 

we used to use `unidecode`, but we had to switch precisely because it's GPL and that was too restrictive (see #628).

Could you provide any documentation to your statement the the artistic license is not FOSS compatible? I see the artistic license listed as approved on [opensource.org](https://opensource.org/licenses/Artistic-2.0)

jorti:
I'm the Fedora packager of faker and the artistic license is forbidden by the Fedora policy:

https://fedoraproject.org/wiki/Licensing:Main#Bad_Licenses

More info:

http://www.gnu.org/licenses/license-list.html#ArtisticLicense

fcurella:
If I'm understanding this correctly, `text-unidecode` is under Artistic License v1, which is not OK with Fedora, while Artistic Clarified or Artistic v2 would be. Is that a correct interpretation?

Ultimately, the issue is that we don't want `unidecode` because it's GPL, and we don't want `text-unicode` because it's FSF approved.

Looks like `text-unidecode` is in the process of changing their license (see https://github.com/kmike/text-unidecode/issues/9 - last updated 10 days ago). Maybe you can help them out?

tbrlpld:
@fcurella Yes, looks like they clarified the license which is now [Artistic 1.0 or GPL 2+](https://github.com/kmike/text-unidecode/commit/ec280038347b79e3c9f635de6e0c12d7c126f041).

So now it would fail both the requirements stated above. It is now GPL and not FSF approved.

kmike:
@tbrlpld Perl license (dual Artistic and GPL) should be fine for Fedora: https://fedoraproject.org/wiki/Licensing:Main#Good_Licenses, this is exactly a license text-unidecode is using.

fcurella:
@jorti Can you confirm if  GPLv2 + Artistic V1 works for Fedora?

kmike:
These are good questions @tbrlpld. 

My feeling is that the licenses may work out ok. Fedora prohibits packaging Artistic 1.0-only packages as they're not considered free - but Perl license it not prohibited, as users of the package can use GPL terms if they want. python-faker also can use text-unidecode, even in Fedora, as using an Artistic-licensed or Perl-licensed package doesn't put restrictions on python-faker, according to the license terms. I don't see a conflict here. Having text-unidecode in Fedora doesn't change text-unidecode license to GPL, the license is still a dual license, even if text-unidecode is executed under Fedora.

There must be zillions of Perl modules packaged for Fedora, with a different mixes of licenses, there could be similar cases to look at. 

Sorry, I won't be able to participate in deeper license discussions, it is such a rabbit hole, and I have neither knowledge time nor time to get this knowledge :)

And anyways, I can't change text-unidecode license, because the license is inherited, it was not my decision to use

tbrlpld:
@kmike Haha, yea I know what you are talking about. It is a rabbit hole, and I have fallen in I am afraid... 

Regarding you being forced to keep the Perl dual license model, this article would suggest otherwise:
> In 1991, Wall carried out what he later called a“cultural hack”(quoted in Tamiya, 2001), inventing the dual license by releasing Perl under both the GPL and The ArtisticLicense. The latter was authored by Wall, and distinguished itself from the GPL in two important ways. First, it laid out a number of provisions to ensure the relative integrity of Perl, allowing modifications and reuse but maintaining“some semblance of artistic control”over“the standard distribution”(Wall,1991, emphasis in original). Second, it made explicit that Perl could be used to create commercial products that could be released as proprietary software.
> 
> The Artistic License thus emphasized authorship and control at the same time that it was more permissive than the GPL. If the GPL enforces a form of solidarity in which any modifications or extensions are returned to the commons, the Artistic License seeks to maintain Perl’s integrity and Wall’s reputation while simultaneously allowing for commercial use. **The fact that users could choose which license they wanted to use Perl under was itself innovative.** Rather than remove the GPL and risk alienating the free software community, Wall’s solution meant having it both ways: Perl belonged to the collective and to the author; it was an expression of GNU solidarity while also a tool for commercial software production.
— [Full article: Having it both ways: Larry Wall, Perl and the technology and culture of the early web](https://www.tandfonline.com/doi/full/10.1080/24701475.2018.1495810)

The fact that your dependency it is dual licensed with GPL **OR** Artistic should mean that you as the "user" can choose to use the dependency under the Artistic license which does not impose the limitation to reuse the same license. 


> a) the GNU General Public License as published by the Free Software Foundation; either version 1, or (at your option) any later version, or
>
> b) the "Artistic License". 
-- https://dev.perl.org/licenses/

But I guess the same would then count for Faker since you are licensing with GPL **OR** Artistic as well would give Faker the freedom to use `text-unidecode` under the Artistic license. Fedora on the other hand can choose to look at `text-unidecode` as being GPL licensed which fulfills their requirement for a "good license".

Sorry for ranting. I guess I am more talking to myself than anything. 

Just to clarify for myself: if I depend on Faker (which pulls in `text-unidecode`) I am a user of `text-unidecode` and I can choose to use `text-unidecode` under the Artistic license. I guess...

fcurella:
The only thing that really matters for the purpose of this issue is if the original author (@jorti ) is OK with the new license.

jorti:
> @jorti Can you confirm if GPLv2 + Artistic V1 works for Fedora?

Yes, the new dual licensing of text-unidecode allows us to choose GPLv2+. This is the same case as Perl, so there's no problem now.

Thank you.

fcurella:
Thank you @jorti ! That's all we need! :)

PR #991, opened 2019-08-25
Add option to return list of n unique companies

### LINKED ISSUE #971: Return list of n unique companies (n_unique keyword for all providers?)
* Faker version: 1.0.7
* OS: Ubuntu 16.04

I was wondering if it is possible to generate a list of 'n'  unique company names? I saw some PR's which gave a unique keyword for 'words' but doesn't seem to extend to other providers? I understand i could just keep regenerating and dropping duplicates until I got a unique set of length n, but would be nice to just have a keyword for that (plus this may be suboptimal depending on how Faker implemented under the hood, which i don't know enough about i confess)

alaminopu:
### What does this changes
Add option to return list of n unique companies

### What was wrong
N/A

### How this fixes it
Add a method `companies()` to return unique list

Fixes #971

### DIFF
--- a/faker/providers/company/__init__.py
+++ b/faker/providers/company/__init__.py
@@ -494,6 +494,95 @@ class Provider(BaseProvider):
          'web services',
          'methodologies'))
 
+    # Taken the list nl_NL providers
+    company_names = [
+        'Shell', 'Coolblue', 'ASML', 'Ahold', 'Tata Steel', 'KLM', 'Bol.com', 'BP Nederland', 'De Efteling', 'Eneco',
+        'De Persgroep', 'ING', 'Royal HaskoningDHV', 'Randstad', 'Google', 'Ikea', 'Rockwool', 'BAM', 'Achmea',
+        'Damen Shipyard', 'ABN Amro', 'Remeha Group', 'TenneT', 'Coca-Cola', 'Van Leeuwen Buizen', 'Wavin', 'Rabobank',
+        'AkzoNobel', 'Arcadis', 'AFAS', 'Cisco', 'DAF Trucks', 'DHL', 'Hanos', 'Boon Edam', 'BMW Nederland',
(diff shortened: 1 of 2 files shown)


153957:
I would expect companies to be based on the company provider. And why is it specifically useful to have a list of companies? The same could be useful for most other providers, e.g. 'addresses', 'cities', 'countries', 'license_plates', 'jobs', etc.. A more general solution for this would be preferable.

fcurella:
I would agree with @153957. This is a case common enough a more general solution would be preferable, although I currently don't know what that would like.

@153957 Do you have something in mind?

PS: I find it unnecessary to 'thumb down' a PR. It's discouraging from the PR's author perspective and does not add to the discussion. Just a comment like the one you posted will do :)

153957:
> I would agree with @153957. This is a case common enough a more general solution would be preferable, although I currently don't know what that would like.
> 
> @153957 Do you have something in mind?

I would have to dive into the code a bit, but I would suggest something similar to this PR. I.e. reusing random_choices/random_sample, but then instead of 'elements' have the option to provide another provider as source for the the elements. For the case where the resulting elements need to be unique it might be a challenge since the elements are randomly generated, so it is more difficult to know in advance how many options are even available for that provider.

So something like `random_elements(provider=company, length=10)`.

> PS: I find it unnecessary to 'thumb down' a PR. It's discouraging from the PR's author perspective and does not add to the discussion. Just a comment like the one you posted will do :)

👍 Good point. I started with the thumb down and they decided to elaborate on why I disagreed with the PR, the thumb down was no longer necessary.

alaminopu:
Thanks for the feedback guys. I appreciate it and also agree that a more general solution would be helpful.

153957:
One more thing about the suggested `random_elements(provider=company, length=10)`, it would be useful if `provider` also supported a string reference, to allow calling it with factory_boy like `company_list = factory.Faker('random_elements', provider='company', length=10, unique=True)`

PR #993, opened 2019-09-05
Bump required version text-unidecode to 1.3

### LINKED ISSUE #727: text-unidecode is released under the Artistic license

moggers87:
### What does this changes

Bumps to required version of text-unidecode to 1.3

### What was wrong

text-unidecode was previously licensed under the Artistic License v1 which is incompatible with the GPL

### How this fixes it

text-unidecode version 1.3 is licensed under the terms of both the GPLv2 (or later) and the Artistic License v1, thus removing this incompatibility.

Fixes #727

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,7 @@ setup(
     install_requires=[
         "python-dateutil>=2.4",
         "six>=1.10",
-        "text-unidecode==1.2",
+        "text-unidecode==1.3",
     ],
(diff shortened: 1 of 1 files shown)


moggers87:
When running tests locally there were some flake8 errors, but they also appear in master so won't be fixed here.

fcurella:
Looks good! Thank you everyone who worked on this!

PR #995, opened 2019-09-06
Bug fix for pyfloat going over max_value (issue #994)

### LINKED ISSUE #994: pyfloat producing values above max_value
* Faker version: 2.0.1
* OS: Windows 10

Calling `pyfloat` with the `max_value` set can produce values above `max_value` leading to issues.

### Steps to reproduce

Run my example code below.

### Expected behavior

Given my following example, I would expect to _only_ see numbers in the range `1 <= x <= 2`.

```
for _ in range(25):
    print(faker.pyfloat(min_value=1, max_value=2))
```

### Actual behavior

However, my example will produce values in the range `1 <= x < 3`.  This behavior is very confusing and unexpected.  Output I got from running my above example:

```
1.200055
2.709
2.319785
2.773763416717
1.521
1.9
2.52454
2.91
1.87016
2.35457
1.92215
1.7
2.461453
1.922252943
1.416632
1.448
2.0
1.5
2.31
2.5114
2.18
1.8
2.12503540581
2.0
```

fgs-dbudwin:
### What does this changes

When calling `pyfloat` with a `max_value`, it was possible for the `pyfloat` function to generate a random number within the range of `min_value <= x < (max_value + 1)` instead of the expected `min_value <= x <= max_value`.

### What was wrong

`pyfloat` was generating numbers above `max_value` sometimes.

### How this fixes it

Subtracts 1 from the `max_value` so that it doesn't go over by 1.  `left_number` was being inclusive when it should be exclusive for floating point numbers.

Fixes #994

### DIFF
--- a/tests/test_factory.py
+++ b/tests/test_factory.py
@@ -466,6 +466,16 @@ class FactoryTestCase(unittest.TestCase):
         with pytest.raises(ValueError):
             provider.pyfloat(left_digits=0, right_digits=0)
 
+    def test_pyfloat_in_range(self):
+        # tests for https://github.com/joke2k/faker/issues/994
+        factory = Faker()
+
+        factory.seed_instance(5)
+        result = factory.pyfloat(min_value=0, max_value=1)
+
+        assert result >= 0.0
+        assert result <= 1.0
+
     def test_negative_pyfloat(self):
         # tests for https://github.com/joke2k/faker/issues/813
         factory = Faker()
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -60,7 +60,7 @@ class Provider(BaseProvider):
             if max_value is None:
                 max_value = min_value + self.random_int()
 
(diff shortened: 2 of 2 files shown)


fgs-dbudwin:
Ok to test

tests/test_factory.py
```diff
@@ -466,6 +466,15 @@ def test_random_pyfloat(self):
         with pytest.raises(ValueError):
             provider.pyfloat(left_digits=0, right_digits=0)
 
+    def test_pyfloat_in_range(self):
+        from faker.providers.python import Provider
+        provider = Provider(self.generator)
+
+        for _ in range(99):
```
fcurella: Let's find a seed value that would've failed the test before the fix an use that, instead of iterating 99 times. See ` test_negative_pyfloat` below for an example

fgs-dbudwin: Good call!  A seed of 5 works, pushing fix now.

fcurella:
Thank you! ✨

PR #997, opened 2019-09-18
Use the provider's RNG instead of the random module in invalid_ssn

### LINKED ISSUE #996: `invalid_ssn` provider doesn't yield reproducible values with the same seed
* Faker version: 2.0.1
* OS: macOS 10.14.6

The `invalid_ssn` provider doesn't yield reproducible values when used with the same random seed. It's not immediately noticeable but it does show up if you generate a number of values.

### Steps to reproduce

```
Python 3.7.2 (default, Aug 29 2019, 16:32:19) 
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
>>> from faker import Faker
>>> f1 = Faker()
>>> f2 = Faker()
>>> f1.seed_instance(1); f2.seed_instance(1)
<faker.generator.Generator object at 0x10e36dcf8>
<faker.generator.Generator object at 0x10e286320>
>>> f1.invalid_ssn()
'137-97-0000'
>>> f2.invalid_ssn()
'137-97-0000'
>>> [f1.invalid_ssn() for _ in range(1000)] == [f2.invalid_ssn() for _ in range(1000)]
False
>>> f1.seed_instance(1); f2.seed_instance(1)
<faker.generator.Generator object at 0x10e36dcf8>
<faker.generator.Generator object at 0x10e286320>
>>> while True:
...   a = f1.invalid_ssn()
...   b = f2.invalid_ssn()
...   if a != b:
...     print(f"{a} != {b}")
...     break
... 
923-67-5200 != 923-69-5200
```

### Expected behavior

With the same seed `invalid_ssn` should yield the same series of values.

### Actual behavior

The series of values differ.

From briefly looking at the code I'd guess the problem is the use of `random.choice` here which would use the global RNG instead of the instance RNG:
https://github.com/joke2k/faker/blob/cfffeb82086db9d8519c8db4f93b833316062629/faker/providers/ssn/en_US/__init__.py#L199

luser:
This change makes the `invalid_ssn` provider use `self.random_element` instead of `random.choice` so that it will correctly honor the provider's random seed.

Fixes #996

### DIFF
--- a/faker/providers/ssn/en_US/__init__.py
+++ b/faker/providers/ssn/en_US/__init__.py
@@ -1,7 +1,6 @@
 # coding=utf-8
 from __future__ import unicode_literals
 
-import random
 from .. import Provider as BaseProvider
 
 
@@ -28,7 +27,7 @@ class Provider(BaseProvider):
         serial = self.random_int(min=0, max=9999)
 
         # The group number must be between 70 and 99 inclusively but not 89 or 93
-        group = random.choice([x for x in range(70, 100) if x not in [89, 93]])
+        group = self.random_element([x for x in range(70, 100) if x not in [89, 93]])
 
         itin = "{0:03d}-{1:02d}-{2:04d}".format(area, group, serial)
         return itin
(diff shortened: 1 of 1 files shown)


fcurella:
Thank you! ✨

PR #998, opened 2019-09-26
fix randomize_nb_elements max param

jorrit-wehelp:
### What does this changes
The `max` argument for randomize_nb_elements is undocumented (and only the min is every used internally)....but it does exist and is broken :)

### What was wrong

Tiny typo/copy-paste error, `> min` instead of `>max`


fcurella:
Nice Catch! Thank you! ✨

PR #1006, opened 2019-10-06
Spanish Unit Tests

Newman101:
### What does this changes

Adds unit tests for address module of es_ES localization.

### What was wrong

The class was missing unit tests. Also, the the test classes which name was starting with letter E was after test classes starting with H. As far as I know, these test classes are supposed to be in alphabetical order.

### How this fixes it

1. Added the unit tests.
2. Moved the test classes in proper alphabetical order.

All unit tests passed locally. Thanks.


Newman101:
Looks like flake8 error. I'll fix this as soon as I can.

Newman101:
That should be fixed now. Let's see what CI says.

PR #1007, opened 2019-10-07
Added pt_PT automotive plates

rubenandre:
### What does this changes

Added pt_PT automotive plates implementation

### What was wrong

\-

### How this fixes it

\-


rubenandre:
Failed checks will be fixed in next pull request

PR #1010, opened 2019-10-09
Greek jobs

athaks:
I just added 445 jobs in Greek


faker/providers/job/el_GR/__init__.py
```diff
@@ -0,0 +1,453 @@
+# coding=utf-8
+from __future__ import unicode_literals
+from .. import Provider as BaseProvider
+
+
+class Provider(BaseProvider):
+    jobs = [
+        'Αγγειοπλάστης',
```
fcurella: do you have a source for this list? If so, could you include it as comment?

faker/providers/ssn/el_GR/__init__.py
```diff
@@ -4,17 +4,31 @@
 
 class Provider(BaseProvider):
     """
-    A Faker provider for the Greek VAT IDs
+    A Faker provider for the Greek VAT IDs and Greek police IDs
```
fcurella: this is out of scope for this PR. Could  you submit it as separate PR?

faker/providers/job/el_GR/__init__.py
athaks: there is no specific source, I was browsing the web gathering sample data to populate the job descriptions list

faker/providers/ssn/el_GR/__init__.py
athaks: apologies, this was meant to be my next PR but somehow slipped in this one

athaks:
Added tests and removed accidentally slipped source file

fcurella:
Thank you! ✨

PR #1011, opened 2019-10-11
Added Greek police ID under SSN provider for el_GR

athaks:
### What does this changes

Adds Greek police ID "numbers"

### What was wrong

Greek police ID "numbers" were missing

### How this fixes it

Added generator and test


fcurella:
Thank you!

PR #1012, opened 2019-10-13
Implemented jobs provider pt_PT

rubenandre:
### What does this changes
Added jobs provider for Portugal.
Implemented respective unit tests

### What was wrong

\-

### How this fixes it

\-


fcurella:
Thank you! ✨

PR #1019, opened 2019-10-17
Added provider for Finnish IBAN numbers

sitomani:
### What does this changes

This change adds Finnish IBAN number provider.

### What was wrong

Faker did not produce Finnish IBAN numbers (would generate GB entries for 'fi_FI' locale)

### DIFF
--- a/faker/providers/bank/fi_FI/__init__.py
+++ b/faker/providers/bank/fi_FI/__init__.py
@@ -1,5 +1,6 @@
 from .. import Provider as BankProvider
 
 class Provider(BankProvider):
-    bban_format = '##############'
+    bban_format = '################'
     country_code = 'FI'
+
--- a/faker/providers/bank/fi_FI/__init__.py
+++ b/faker/providers/bank/fi_FI/__init__.py
@@ -1,0 +1,5 @@
+from .. import Provider as BankProvider
+
+class Provider(BankProvider):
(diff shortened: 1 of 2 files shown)


faker/providers/bank/fi_FI/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .. import Provider as BankProvider
+
```
fcurella: could you add a blank line here for PEP8?

sitomani: done

sitomani: I also added tests, and they run fine on my local machine
> python setup.py test 
results in all green. Am I missing something here?

tests/providers/test_bank.py
```diff
@@ -16,6 +16,19 @@ def test_bban(self):
         bban = self.factory.bban()
         assert re.match(r"\d{11}", bban)
 
+class TestFiFi(unittest.TestCase):
```
fcurella: same here re: blank lines: you need 1 more above and one more below

sitomani: Should be OK now, pep8 reports no problems on the test_bank.py

fcurella:
Thank you! ✨

PR #1021, opened 2019-10-17
Faker en NG

djunehor:
### What does this changes

It added support Nigerian locale

### What was wrong

There is presently no support for nigerian locale in the package

### How this fixes it

Now, users can specify locale `en_NG` when instantiating the Faker Class to get Nigerian specific data

Fixes #...
Added Nigerian locale specific person, address and phone number data


djunehor:
I've added sources to the list

faker/providers/person/en_NG/__init__.py
```diff
@@ -0,0 +1,241 @@
...
+        '{{first_name_female}} {{last_name}}',
+        '{{first_name_female}} {{last_name}}-{{last_name}}',
+        '{{prefix_male}} {{first_name_male}} {{last_name}}',
+        '{{prefix_female}} {{first_name_female}} {{last_name}}',
+        '{{prefix_male}} {{first_name_male}} {{last_name}}',
+        '{{prefix_female}} {{first_name_female}} {{last_name}}',
+    )
+
+    # Names from
+    # https://andela-celisha-wigwe.github.io/names.html
```
fcurella: This link results in a 404 for me. We can't use sources that are not public.

mastacheata: The user on GitHub changed their username and since GitHub doesn't reserve the old name there's no automatic redirect. The user formerly known as `andela-celisha-wigwe` is now known as `elchroy`:

```suggestion
    # https://elchroy.github.io/names.html
```

fcurella: This appear to be just a list of names that someone compiled without any mention of its statistical basis.

Are there any more "official" sources available? A governmental office of statistics, or a research paper would work great.

djunehor: There's really no official list online, not even on Wikipedia. I've added 2 sources though. But for some weird reason, this pull request is not been updated with my new commits. Had to create another PR https://github.com/joke2k/faker/pull/1042

fcurella:
superseded by #1042

PR #1022, opened 2019-10-17
Fix/master/swedish ssn

### LINKED ISSUE #1005: Proposal for enhancement of Swedish SSN values (Pers ID, Org ID & MOMS (VAT) ID)
* Faker version: 2.0.2
* OS: *

I'd like to expand and correct some of the behaviour for Swedish ssn and VAT numbers, and add an option to return an organisation ID. 

Some info on how these codes work in Sweden:

1. **Personal number (SSN)** - the code for this is largely correct, but I would like to add a "long=False" parameter to the call. Default behaviour would remain the same, when set to true it would return a 12-digit SSN, which includes the century digits for the year. Most systems actually use the long version of the SSN and it would be handy to get this returned directly.

2. **Organisation number** - the code for this is mostly the same as the pers ID (SSN), but the "month" part of the number is 20 or higher. When taking 12 digit versions these are usually prefixed with "16".

3. **VAT number** (VAT is also called MOMS in Sweden) -  This number is not just a random sequence of numbers in Sweden. The organisation number is part of this. So the pattern is usually SEnnnnnnnnnnss, where n= org ID and s=sequence number (usually 01, unless you are a sole trader with multiple businesses)

..See wikipedia under Sweden for more info: https://en.wikipedia.org/wiki/VAT_identification_number

### Expected behavior

**Pers ID:**
1.  Request a Personal number without setting any params
2.  The 10 digit personal number is returned as it is today. 
3.  Request a personal number setting "long=True"
4.  A 12 digit personal number is returned, which passes Swedish validation checks. 

**Org ID:**
1.  Request an org ID setting no params or setting "long=False"
2.  A valid 10 digit org ID is returned (same checkdigit validation as for pers ID), which passes Swedish validation checks.
3.  Request an org ID setting "long=True"
4.  A valid 12 digit org ID is returned, which passes Swedish validation checks. 

**VAT numbers:**
1.  Request a VAT number
2.  The VAT number returned 
3.  The VAT number passes the validation that the first 10 digits after the "SE" are a valid org ID.

Note: Implementation of Org ID could either be as a new function, or using the existing ssn function with a parameter to set "org_id=True". Whichever is preferable.

### Actual behavior

**Pers ID:** Cannot return a long pers ID with 12 digits
**Org ID:** Not an option today
**VAT number:** currently returns a random string of digits

michjnich:
### What does this changes

Changes to Swedish SSN and VAT routines. 

`ssn` now allows 2 parms - `long` (default `False`) and `dash` (default `True`)
`long=True` returns a 12 digit Personal number, including the century, instead of a 10 digit number
`dash=False` returns the Personal number without the hyphen

Default behaviour remains as is.

new method `org_id` returns a valid Swedish organisation number. The format is the similar to Personal number, but the rules for the numeric values are different. This allows the same `long` and `dash` parameters as `ssn`, with the same defaults.

`vat_id` has been amended to use the `org_id` functionality to generate the core number, and base the VAT (MOMS) ID on that. 

A new method `org_and_vat_id` returns a consistent organisation number and VAT ID (as is the case in reality). The `long` and `dash` parameters can be passed here too, but will only affect the org ID, not the VAT number.

### What was wrong

SSN was not incorrect, it has just been extended to be more flexible. 
Org ID was missing. 
VAT number was invalid format and  did not pass validation checks (except occasionally by accident)

### How this fixes it

See above

Closes #1005

### DIFF
--- a/faker/providers/ssn/sv_SE/__init__.py
+++ b/faker/providers/ssn/sv_SE/__init__.py
@@ -71,7 +71,7 @@ class Provider(SsnProvider):
         first_digits = list(self.ORG_ID_DIGIT_1)
         random.shuffle(first_digits)
         onr_one = str(first_digits.pop())
-        onr_one += str(self.generator.random.randrange(0,9)).zfill(1)
+        onr_one += str(self.generator.random.randrange(0, 9)).zfill(1)
         onr_one += str(self.generator.random.randrange(20, 99))
         onr_one += str(self.generator.random.randrange(0, 99)).zfill(2)
         onr_two = str(self.generator.random.randrange(0, 999)).zfill(3)
--- a/faker/providers/ssn/sv_SE/__init__.py
+++ b/faker/providers/ssn/sv_SE/__init__.py
@@ -49,6 +49,12 @@ class Provider(SsnProvider):
 
         return pnr
 
+    def org_id(self, long=False, dash=True):
+        if long:
+            return '16559035-1077'
+        else:
+            return '559035-1077'
+
     vat_id_formats = (
         'SE############',
     )
--- a/faker/providers/ssn/sv_SE/__init__.py
+++ b/faker/providers/ssn/sv_SE/__init__.py
@@ -7,13 +7,16 @@ import datetime
 
 class Provider(SsnProvider):
 
-    def ssn(self, min_age=18, max_age=90):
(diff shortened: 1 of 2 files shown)


michjnich:
Damn, missed running flake8 and need some extra blank lines. Will fix tomorrow ...

fcurella:
Thank you! ✨

PR #1023, opened 2019-10-19
Added tests under test_address for hindi locale

type812:
### What does this changes

Brief summary of the changes

### What was wrong

### How this fixes it

Fixes #...

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,8 +8,11 @@ license_file = LICENSE.txt
 max-line-length = 80
 
 [aliases]
-test=pytest
+test = pytest
 
 [tool:pytest]
 addopts = --verbose
 python_files = tests/*.py
+
+[easy_install]
+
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -8,8 +8,11 @@ license_file = LICENSE.txt
...
 [aliases]
-test=pytest
+test = pytest
 
 [tool:pytest]
 addopts = --verbose
 python_files = tests/*.py
+
+[easy_install]
+
```
fcurella: Can you revert the changes in this file?

tests/providers/test_address.py
```diff
@@ -526,6 +526,14 @@ def test_state(self):
...
         assert isinstance(state, string_types)
 
+    def test_street_title(self):
+        street_title = self.factory.street_title()
+        assert isinstance(street_title, string_types)
+
+    def test_street_suffix(self):
+        suffix = self.factory.street_suffix()
+        assert isinstance(suffix, string_types)
+        assert suffix in FiProvider.street_suffixes
```
fcurella: `FiProvider` is a Finnish provider. I don't think you want to test against that one. You'll have to import the Hindi provider (if it's not imported already) and check against that.

PR #1025, opened 2019-10-19
Add license plate for India states

type812:
### What does this changes
Adds provider class for handling license plate numbers for different Indian states , not including union territories and diplomat license plates
Brief summary of the changes

### What was wrong

### How this fixes it

Fixes #...


faker/providers/person/hi_IN/__init__.py
```diff
@@ -237,4 +237,13 @@ class Provider(PersonProvider):
...
-        'नूरानी')
+        'नूरानी',
+        'श्रीवास्तव',
+        'सिन्हा',
+        'भटनागर',
+        'निगम',
+        'सक्सेना',
+        'माथुर',
+        'कर्ण',
+        'कुलश्रेष्ठ'
```
fcurella: this is out of scope for the PR, and already being fixed in #1024

PR #1028, opened 2019-10-24
split first names into male/female for person/de_AT

Jayday:
### What does this changes

Spit first name data for person/de_AT to be able to correctly create person fakes based on gender

### What was wrong

first names were aggregated in one list

### How this fixes it

changed the data structure (partly) according to the de_DE version

Fixes #...


fcurella:
Thank you! ✨

PR #1034, opened 2019-10-30
Add PH locales

malefice:
### What does this changes

This will add the following providers for the PH locales (en_PH, fil_PH, tl_PH):

- address
- automotive
- company
- date_time
- internet
- lorem
- misc
- phone_number
- ssn

### What was wrong

There are none for PH locales.


faker/providers/company/en_PH/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+from collections import OrderedDict
+from .. import Provider as CompanyProvider
+
+
+class Provider(CompanyProvider):
+    """
+    Provider for company names for en_PH locale
+
+    Company naming scheme and probabilities are inspired by and/or based on existing companies in the Philippines.
+    Parody names of popular local companies are also available.
```
fcurella: Jokes are notoriously misinterpreted across cultures. I would drop the parody names.

faker/providers/company/fil_PH/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+        ('Kahit kailan, {{random_good_service_adjective_chain}}!', 0.12),
+    ])
+    noun_ish_good_traits = (
+        'bida',
+        'ginhawa',
+        'haligi',
+        'karangalan',
+        'lingkod',
+        'liwanag',
+        'numero_uno',
```
fcurella: is this supposed to use an underscore rather than a hyphen?

faker/providers/misc/en_PH/__init__.py
```diff
@@ -0,0 +1,56 @@
+# coding=utf-8
+from __future__ import unicode_literals
+from .. import Provider as MiscProvider
+
+
+class Provider(MiscProvider):
```
fcurella: I dont see this as a use-case commmon enough to warrant inclusion in `Faker`, but you could package it as a [Community Provider](https://github.com/joke2k/faker/blob/master/docs/communityproviders.rst)

faker/providers/company/fil_PH/__init__.py
malefice: It should be a whitespace. Derp.

faker/providers/company/en_PH/__init__.py
malefice: Well, that is why it is locale-specific hahaha! But sure, I will drop those.

faker/providers/misc/en_PH/__init__.py
fcurella: Oh, I missed the fact that it was in the address provider. Let's leave it then. 👍

fcurella:
Thank you so much! ✨

PR #1036, opened 2019-11-04
Hide "Edit on Github" links in autogenerated pages

### LINKED ISSUE #264: "Edit on Github" link broken in ReadTheDocs
http://fake-factory.readthedocs.org/en/latest/locales.html

Clicking "Edit on Github" results in a 404 error.

EDIT: 
http://fake-factory.readthedocs.org/en/latest/ has a github link to `https://github.com/joke2k/faker/blob/docs/docs/index.rst` when the correct link is
`https://github.com/joke2k/faker/blob/master/docs/index.rst`
(Note the doubled up `docs/docs` instead of `master/docs`)

malefice:
### What does this change

This removes the "Edit on Github" links found in the autogenerated docs under providers and locales without affecting pages with actual sources in Github.

### What was wrong

The docs imply that the source .rst files will be available in Github, but clicking on those links will lead to a 404.

### How this fixes it

Basically, we set the `github_url` file-level metadata to a certain value (in this case `hide`), and then we extend the default `breadcrumbs.html` template of `sphinx_rtd_theme` to check for that value. If the value of `github_url` is set to `hide`, we skip the default template processing for that `block`, effectively disabling the link for that page. Otherwise, it would proceed as it normally would. Reference material used can be found [here](https://docs.readthedocs.io/en/stable/guides/remove-edit-buttons.html) and [here](https://sphinx-rtd-theme.readthedocs.io/en/latest/configuring.html#confval-github_url), and see it in action [here](https://maleficefakertest.readthedocs.io/en/issue-264/index.html).

Fixes #264

### DIFF
--- a/docs/_templates/breadcrumbs.html
+++ b/docs/_templates/breadcrumbs.html
@@ -1,0 +1,7 @@
+{%- extends "sphinx_rtd_theme/breadcrumbs.html" %}
+
+{% block breadcrumbs_aside %}
+  {% if not meta or meta.get('github_url') != 'hide' %}
+  {{ super() }}
+  {% endif %}
+{% endblock %}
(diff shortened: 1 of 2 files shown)


fcurella:
Thank you! ✨

jeffwidman:
Ohh, that's a clever way to solve it. ❤️

PR #1037, opened 2019-11-05
Optimize IPv4 address generation

### LINKED ISSUE #845: ipv4 faker very slow
Hi! We updated from Faker 0.9.0 to 0.9.2 and noticed our tests speed decreased by 5X. We tracked the bug to the changes in the ipv4 faker done in commit 3b847f48bbfc33aed3e86cffcedbcfd1c0cccec5

### Steps to reproduce

1. checkout to commit 3b847f48bbfc33aed3e86cffcedbcfd1c0cccec5
1. run the following: `time python -c 'import faker; fake = faker.Faker(); [fake.ipv4() for _ in range(100)]'`
1. Check its output to see it was slow. in my laptop it took about 6 seconds.
1. Run `git checkout HEAD~1` to go to a working commit
1. Run the command of step 2 again and it will work much faster (in my case it took 0.4 seconds)

### Expected behavior

Generating ivp4 addresses should be fast

### Actual behavior

Generating ipv4 addresses is slow

malefice:
### What does this change

Reduce IPv4 address generation time by around 33%. Using the master branch, the command below ran for 17.811s, and with the changes, it ran for 11.458s, a 6.353s reduction.

`time python -c 'import faker; fake = faker.Faker(); [fake.ipv4() for _ in range(1000)]'`

### What was wrong

Slow according to some reports

### How this fixes it

Short of using a totally different algorithm in the `_exclude_ipv4_networks` method (which I think is already pretty good), there are still some ways to speed the current algorithm up. One way is to remove the duplicates and the concatenation performed in the decalaration of `_excluded_networks`.   Another is to sort `networks_to_exclude` to exclude larger networks first i.e. networks with smaller prefix lengths. With how `_exclude_ipv4_networks` is written, when larger networks are excluded first, the resulting list of networks to check in the next iteration is guaranteed to be smaller, and that in turn leads to fewer overall operations.

### Other notes

Admittedly, there is a better way, and that is to generate the network lists only once per provider instantiation, and we can make probably do some sort of lazy loading to only create the relevant network list when needed. The bulk of the execution time is in the generation of the network lists, and we can make subsequent runs faster but with higher memory usage. If that is an acceptable approach, I can update this pull request.

Fixes #845

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -10,6 +10,7 @@ from ipaddress import ip_address, ip_network, IPV4LENGTH, IPV6LENGTH
 # from faker.generator import random
 # from faker.providers.lorem.la import Provider as Lorem
 from faker.utils.decorators import lowercase, slugify, slugify_unicode
+from faker.utils.distribution import choices_distribution
 
 
 localized = True
@@ -29,12 +30,6 @@ class _IPv4Constants:
         'c': ip_network('192.0.0.0/3'),
     }
 
-    _linklocal_network = ip_network('169.254.0.0/16')
-
-    _loopback_network = ip_network('127.0.0.0/8')
-
-    _multicast_network = ip_network('224.0.0.0/4')
-
     # Three common private networks from class A, B and CIDR
     # to generate private addresses from.
     _private_networks = [
@@ -49,8 +44,8 @@ class _IPv4Constants:
     _excluded_networks = [
         ip_network('0.0.0.0/8'),
         ip_network('100.64.0.0/10'),
-        ip_network('127.0.0.0/8'),
-        ip_network('169.254.0.0/16'),
+        ip_network('127.0.0.0/8'),  # loopback network
+        ip_network('169.254.0.0/16'),  # linklocal network
         ip_network('192.0.0.0/24'),
         ip_network('192.0.2.0/24'),
         ip_network('192.31.196.0/24'),
@@ -60,12 +55,9 @@ class _IPv4Constants:
         ip_network('198.18.0.0/15'),
         ip_network('198.51.100.0/24'),
         ip_network('203.0.113.0/24'),
+        ip_network('224.0.0.0/4'),  # multicast network
         ip_network('240.0.0.0/4'),
         ip_network('255.255.255.255/32'),
-    ] + [
-        _linklocal_network,
-        _loopback_network,
-        _multicast_network,
     ]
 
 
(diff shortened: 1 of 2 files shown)


fcurella:
Thank you so much @malefice !

I think I'm OK with a higher memory usage as long as it's not crazy high and it's only if the user requests some IP address. In other words, I wouldn't compile the list on instantiation, but the first time it's needed.

malefice:
Yeah, that is the plan. As for higher memory usage, it should be slightly higher at worst. We are talking about caching several lists with possibly a few hundred elements each, and they will only be generated during a cache miss. I just tested a rough implementation for private IPv4 networks, and generating 1000 such addresses from the same provider instance only takes 0.126s on my machine, and generating 100,000 only takes 1.4s. I will just need to polish/beautify it.

malefice:
In the process of polishing my initial implementation to improve performance, I also found out that current IPV4 generation can be biased against IP addresses in larger subnets, leading to more recycled IP's from smaller subnets when generating in bulk. This is because the equal distribution is applied on the subnet level, regardless of how large each subnet is. To demonstrate, suppose that the resulting network list is the following:

- A - ip_network('10.0.0.0/8')
- B - ip_network('192.168.0.0/24')
- C - ip_network('192.168.1.0/24')
- D - ip_network('192.168.2.0/24')

Under the current code, each of the networks A, B, C, and D have a 25% chance of being selected, but each IP in network A only has a 1 in 16,777,216 chance of being selected if network A was already chosen. On the other hand, each IP in network B, C, and D has a 1 in 256 chance of being selected if the corresponding network was already chosen.

Each IP in network A then only has around a 1 in 67 million chance of being generated (25% of 1 in 16,777,216), but each IP in networks B, C, and D has around 1 in 1024 chance (25% of 1 in 256). This proves that the IP addresses generated are not equally distributed.

To fix this, I applied weights in the selection process, and the weights are simply the sizes of the subnets, and each IP regardless of subnet has a 1 in X chance of being selected, where X is the total number of addresses across all the subnets.

With this bias correction and the caching behavior, I can generate 1 million addresses using `ipv4()` in around 23 to 24 seconds on my machine. Comparing it to 1000 addresses in 17.8 seconds of the old implementation, this is a 740x to 750x performance improvement.

fcurella:
NIce work @malefice ! ❤️ 

Could you also write some tests for the `except` clauses missed [here](https://coveralls.io/builds/26824112/source?filename=faker%2Fproviders%2Finternet%2F__init__.py#L360)?

kungfu71186:
This is fantastic. Great job! @malefice

fcurella:
LGTM! Thank you! 🚀

PR #1038, opened 2019-11-05
Add job in ja_JP.

shmokmt:
### What does this changes

Add Japanese jobs for provider.


fcurella:
Thank you! ✨

faker/providers/job/ja_JP/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+        '調理師',
+        '電気工事士',
+        '農家',
+        '配管工',
+        'バスガイド',
+        '花火師',
+        '漫画家',
+        'モデル',
+        '薬剤師',
+        'YouTuber',
```
shmokmt: YouTuber is a famous job in Japan.

PR #1041, opened 2019-11-07
Create separate docs on base provider methods

### LINKED ISSUE #642: Base provider methods are documented with the address provider.
In the docs the [BaseProvider methods](https://github.com/joke2k/faker/blob/master/faker/providers/__init__.py) are grouped with the [address provider](https://github.com/joke2k/faker/blob/master/faker/providers/address/__init__.py). Here is the documentation: http://faker.readthedocs.io/en/master/providers/faker.providers.address.html

I would expect those to be in their own category, and they have nothing to do with addresses.

### LINKED ISSUE #832: Documentation of utilities like random_choices
Is there any online documentation of the utility functions like random_choices? I always end up looking at the source code when I forget what they're named.

malefice:
### What does this change

Create separate docs on base provider methods

### What was wrong

Base provider methods are always documented under the 1st provider module which is currently `address`. This is not only "unsemantic" but it can also give the impression that the base provider methods are not documented.

### How this fixes it

This patch will write a new source file just for the base provider, and then exclude the base provider methods for subsequent provider modules. See it in action [here](https://maleficefakertest.readthedocs.io/en/issue-832/providers.html).

Fixes #642, fixes #832 .

### DIFF
--- a/faker/build_docs.py
+++ b/faker/build_docs.py
@@ -15,6 +15,12 @@ def write(fh, s):
     return fh.write(s.encode('utf-8'))
 
 
+def write_base_provider(fh, doc, base_provider):
+    formatters = doc.get_provider_formatters(base_provider)
+    write(fh, ':github_url: hide\n\n')
+    write_provider(fh, doc, base_provider, formatters)
+
+
 def write_provider(fh, doc, provider, formatters, excludes=None):
 
     if excludes is None:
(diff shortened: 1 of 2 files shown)


fcurella:
Eventually I would also like to list locale-specific methods on each provider separately. But this is already a big step forward, and we tackle that on a separate PR.

Thank you so much!

PR #1044, opened 2019-11-09
add more phone numbers

shervinmo:
### What does this changes

Add more phone number prefixes.

### What was wrong

Id dose not cover all of iranian mobile numbers

### How this fixes it

add new operator and 8 new prefixes 

Fixes #...


fcurella:
Thank you! ✨

PR #1045, opened 2019-11-09
Improve barcode provider

### LINKED ISSUE #456: EAN-13 starting with 0
ean13 function generates codes starting with zeroes.

According to http://www.barcodeisland.com/ean13.phtml ean13 has to duplicate the leading zero.
So basically UPC-A code "075678164125" in ean13 form should be "0075678164125".

malefice:
### What does this change

1. Add support for generating only EAN-13 barcodes that do not start with a zero
2. Add support for generating only EAN-13 barcodes that start with zero
3. Add support for generating random UPC-A and UPC-E barcodes
4. Add support for generating only UPC-A barcodes with UPC-E equivalents

### What was wrong

Current master does not have the aforementioned improvements

Fixes #456 as well

### DIFF
--- a/faker/providers/barcode/__init__.py
+++ b/faker/providers/barcode/__init__.py
@@ -17,7 +17,7 @@ class Provider(BaseProvider):
         r'(?P<mfr_code>\d{2})'              # the first 2 digits make up the manufacturer code,
         r'(?:(?P<extra>[012])0{4})'         # if immediately followed by 00000, 10000, or 20000,
         r'(?P<product_code>\d{3})'          # a 3-digit product code,
-        r'(?P<check_digit>\d)$'             # and finally a check digit.
+        r'(?P<check_digit>\d)$',            # and finally a check digit.
     )
     upc_ae_pattern2 = re.compile(
         r'^(?P<number_system_digit>[01])'   # The first digit must be 0 or 1
@@ -25,17 +25,17 @@ class Provider(BaseProvider):
         r'(?P<mfr_code>\d{3,4}?)'           # the first 3 or 4 digits make up the manufacturer code,
         r'(?:0{5})'                         # if immediately followed by 00000,
         r'(?P<product_code>\d{1,2})'        # a 2-digit or single digit product code,
-        r'(?P<check_digit>\d)$'             # and finally a check digit.
+        r'(?P<check_digit>\d)$',            # and finally a check digit.
     )
     upc_ae_pattern3 = re.compile(
         r'^(?P<number_system_digit>[01])'   # The first digit must be 0 or 1
         r'(?=\d{11}$)'                      # followed by 11 digits of which
         r'(?P<mfr_code>\d{5})'              # the first 5 digits make up the manufacturer code,
         r'(?:0{4}(?P<extra>[5-9]))'         # if immediately followed by 0000 and a 5, 6, 7, 8, or 9,
-        r'(?P<check_digit>\d)$'             # and finally a check digit.
+        r'(?P<check_digit>\d)$',            # and finally a check digit.
     )
 
-    def _ean(self, length=13, *, leading_zero=None):
+    def _ean(self, length=13, leading_zero=None):
         if length not in (8, 13):
             raise AssertionError("length can only be 8 or 13")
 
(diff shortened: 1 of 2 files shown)


malefice:
There are some flake8 errors and some errors caused by Python3-only syntax. I will fix all these errors once I have reviewer feedback.

faker/providers/barcode/__init__.py
```diff
@@ -23,8 +58,157 @@ def ean(self, length=13):
...
+        return ''.join(str(x) for x in code)
+
+    def ean(self, length=13):
+        return self._ean(length)
+
     def ean8(self):
-        return self.ean(8)
+        return self._ean(8)
+
+    def ean13(self, *, no_leading_zero=False, force_leading_zero=False):
```
fcurella: I'm not sure why we should use two arguments. Couldn't we just have `leading_zero`?

I think this will behavior would be more natural:

* `leading_zero=None`: the leading digit could be anything
* `leading_zero=True`: the leading digit will be `0`
* `leading_zero=False`: the leading digit will be 1-9

faker/providers/barcode/__init__.py
```diff
@@ -23,8 +58,157 @@ def ean(self, length=13):
...
+          may only start with 0 or 1.
+        - When `upc_ae_mode` is enabled, the `number_system_digit` may be explicitly set to 0 or 1.
+          If it is not set or is invalid, either values will be chosen at random.
+        - When `upc_ae_mode` is enabled, a 6-digit UPC-E string `base` may be explicitly set. If it
+          is not set or is invalid, a random 6 digit combination will be used.
+        - If `upc_ae_mode` is disabled, the values of `base` and `number_system_digit` are ignored.
+
+        :param upc_ae_mode: Set to True explicitly to enable
+        :param base: A 6-digit string
+        :param number_system_digit: 0, 1, '0', or '1'
```
fcurella: I'd rather avoid casting down the line and just make sure this argument is a number to begin with. We can add an `assert isinstance(number_system_digit, int)` as a guard on input

faker/providers/barcode/__init__.py
```diff
@@ -23,8 +58,157 @@ def ean(self, length=13):
...
+          same UPC-A equivalent. For example, any UPC-E barcode of the form abc0000d, abc0003d, and
+          abc0004d share the same UPC-A value abc00000000d, but that UPC-A value will only convert
+          to abc0000d because of (a) how UPC-E is just a zero-suppressed version of UPC-A and
+          (b) the rules around the conversion.
 
-    def ean13(self):
-        return self.ean(13)
+        :param base: A 6-digit string
+        :param number_system_digit: First digit of the barcode
+        :param safe_mode: True or False
```
fcurella: Could you add a description about `safe_mode` here too? I'm thinking people scanning through the code would probably expect to find it here.

malefice: There is actually an explanation in the docstring, but I honestly do not know how to describe it concisely. Maybe you can give me suggestions.

Basically, in accordance to a set of rules, there are UPC-A barcodes that may be zero-suppressed (kind of like how some zeroes may be omitted in IPv6), and those have a UPC-E equivalent. The problem is that if you generate certain UPC-E barcodes arbitrarily out of thin air and convert those to UPC-A and back again to UPC-E, there is a possibility the new UPC-E barcode will not be the same as the original.

You can try that quirk [here](https://www.morovia.com/education/utility/upc-ean.asp). The converter accepts a 6 digit UPC-E base, so try converting `100003` to UPC-A. Copy paste the result, and convert it back to UPC-E, and you will get `100000`.

Safe mode basically guarantees that every UPC-E barcode generated, even if such a base was provided explicitly, can be converted to UPC-A and back to UPC-E ad infinitum.

faker/providers/barcode/__init__.py
malefice: Alright, I will change that to an integer.

faker/providers/barcode/__init__.py
fcurella: I think something like this would work:

```
    :param safe_mode: True or False. Guarantees that every UPC-E barcode generated can be converted back-and forth between UPC-A and UPC-E.
```

malefice: I had something like that in mind, but I felt it was still long. As long as it has your blessing, then I am all for it as well. :+1:

faker/providers/barcode/__init__.py
malefice: I actually thought about that, and for some reason, it felt incredibly unnatural to me at the time. Now that you point it out, I do not know what in the world "past me" was thinking. I will fix this.

fcurella: Don't worry, we all have those moments 😃  That's what code reviews are for!

fcurella:
Thank you so much! Nice work! ✨

PR #1046, opened 2019-11-09
Fix occasional errors in isbn provider

### LINKED ISSUE #655: Fake ISBN10 causes "Registrant/Publication not found"
In rare cases the `fake.isbn10` method throws an exception with the following message: `Exception: Registrant/Publication not found in registrant rule list. `

A full exception message:
```
/usr/local/lib/python3.6/site-packages/faker/providers/isbn/__init__.py:70: in isbn10
    ean, group, registrant, publication = self._body()
/usr/local/lib/python3.6/site-packages/faker/providers/isbn/__init__.py:41: in _body
    registrant, publication = self._registrant_publication(reg_pub, rules)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
reg_pub = '64799998'
rules = [RegistrantRule(min='0000000', max='1999999', registrant_length=2), RegistrantRule(min='2000000', max='2279999', regis...'6480000', max='6489999', registrant_length=7), RegistrantRule(min='6490000', max='6999999', registrant_length=3), ...]

    @staticmethod
    def _registrant_publication(reg_pub, rules):
        """ Separate the registration from the publication in a given
            string.
            :param reg_pub: A string of digits representing a registration
                and publication.
            :param rules: A list of RegistrantRules which designate where
                to separate the values in the string.
            :returns: A (registrant, publication) tuple of strings.
            """
        for rule in rules:
            if rule.min <= reg_pub <= rule.max:
                reg_len = rule.registrant_length
                break
        else:
>           raise Exception('Registrant/Publication not found in registrant '
                            'rule list.')
E           Exception: Registrant/Publication not found in registrant rule list.

/usr/local/lib/python3.6/site-packages/faker/providers/isbn/__init__.py:59: Exception
```
### Steps to reproduce

Call `faker.providers.isbn.Provider._registrant_publication` with any of the following values for the `reg_pub` param: `64799998`, `39999999`. These values are valid randomly generated strings from [L34](https://github.com/joke2k/faker/blob/master/faker/providers/isbn/__init__.py#L37).

Code:
```python
from faker.providers.isbn import Provider
from faker.providers.isbn.rules import RULES

# Fails; throws an exception
Provider._registrant_publication('64799998', RULES['978']['0'])
Provider._registrant_publication('39999999', RULES['978']['1'])

# Works; but may be invalid
Provider._registrant_publication('64799998', RULES['978']['1'])
Provider._registrant_publication('39999999', RULES['978']['0'])
```

### Expected behavior

The `faker.providers.isbn.Provider._body` should generate valid `reg_pub` values.

### Actual behavior

It generates values for `reg_pub` that are not accepted by the rules defined in `faker.providers.isbn.rules`.

malefice:
### What does this changes

This will fix errors that may pop up when generating ISBN's. 

### What was wrong

The problem lies [here](https://github.com/joke2k/faker/blob/fd041439e5096290ee33cf2eb9f2576f3506c985/faker/providers/isbn/__init__.py#L55). The values for `rule.min` and `rule.max` are 7-digit strings, while the value of `reg_pub` is an 8-digit string. The line is comparing the values of strings, so if `rule.max` is `0000009`, any 8 digit string that starts with `0000009` will automatically be invalid, triggering the exception.

An interesting side effect of this string comparison is that a string of any length, as long as it starts with 6 zeroes and then any number from 0 to 8, will pass the check e.g. `0000008123456789foobar`.

### How this fixes it

Just compare the first 7 digits of `reg_pub`.

Fixes #655

### DIFF
--- a/faker/providers/isbn/__init__.py
+++ b/faker/providers/isbn/__init__.py
@@ -52,7 +52,7 @@ class Provider(BaseProvider):
         :returns: A (registrant, publication) tuple of strings.
         """
         for rule in rules:
-            if rule.min <= reg_pub <= rule.max:
+            if rule.min <= reg_pub[:-1] <= rule.max:
                 reg_len = rule.registrant_length
                 break
         else:
--- a/tests/providers/test_isbn.py
+++ b/tests/providers/test_isbn.py
@@ -43,8 +43,13 @@ class TestProvider(unittest.TestCase):
     def test_reg_pub_separation(self):
         r1 = RegistrantRule('0000000', '0000001', 1)
         r2 = RegistrantRule('0000002', '0000003', 2)
-        assert self.prov._registrant_publication('0000000', [r1, r2]) == ('0', '000000')
-        assert self.prov._registrant_publication('0000002', [r1, r2]) == ('00', '00002')
+        assert self.prov._registrant_publication('00000000', [r1, r2]) == ('0', '0000000')
+        assert self.prov._registrant_publication('00000010', [r1, r2]) == ('0', '0000010')
+        assert self.prov._registrant_publication('00000019', [r1, r2]) == ('0', '0000019')
+        assert self.prov._registrant_publication('00000020', [r1, r2]) == ('00', '000020')
+        assert self.prov._registrant_publication('00000030', [r1, r2]) == ('00', '000030')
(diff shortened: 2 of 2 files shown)


fcurella:
Thank you so much ✨

PR #1047, opened 2019-11-11
Add support for token-based string generation

### LINKED ISSUE #362: Would a faker.format() method make sense?
It could enable the syntax:
`print(faker.format('{name} considers a {catch_phrase} in {country}'))`

A sample implementation show how simple it would be:

``` python
def fake_fmt(fmt="{first_name}'s favorite number: {random_int}", fake=None):
    fake = fake or Faker()
    fields = [fld.split('}')[0].split(':')[0] for fld in fmt.split('{')[1:]]
    return fmt.format(**{field: getattr(fake, field)() for field in fields})
```

More test cases at: https://github.com/cclauss/Ten-lines-or-less/blob/master/fake_format.py

### LINKED ISSUE #550: Fake Part Numbers, Names & Descriptions
Is there any interest in adding fake part numbers, names and descriptions? I could see this used for seeding product tables. Examples:
 `fake.part_number()`
`# MHP-17854623`

`fake.part_name()`
`# MOSFET Oscillator`

`fake.part_description()`
`# Electrical Coil Sensor DC Block Type`

I recently had a project where I had to seed a product table, and I'd be happy to do a PR.

malefice:
### What does this change

Strings may now be generated from numerify, lexify, and provider method tokens, e.g. `{{address}} {{random_int}}????###`

### What was wrong

There was none.

Fixes #362 fixes #550

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -3,6 +3,7 @@
 from __future__ import unicode_literals
 
 from decimal import Decimal
+import string
 import sys
 
 import six
@@ -31,6 +32,9 @@ class Provider(BaseProvider):
                     length=self.generator.random.randint(min_chars, max_chars),
                 ),
             )
+
+    def pystr_format(self, string_format='?#-###{{random_int}}{{random_letter}}', letters=string.ascii_letters):
+        return self.bothify(self.generator.parse(string_format), letters=letters)
 
     def pyfloat(self, left_digits=None, right_digits=None, positive=False,
                 min_value=None, max_value=None):
--- a/faker/generator.py
+++ b/faker/generator.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 
 import random as random_module
 import re
+import six
 
 _re_token = re.compile(r'\{\{(\s?)(\w+)(\s?)\}\}')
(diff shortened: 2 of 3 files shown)


faker/providers/python/__init__.py
```diff
@@ -32,6 +33,9 @@ def pystr(self, min_chars=None, max_chars=20):
                 ),
             )
 
+    def pystr_format(self, string_format='?#-###{{random_int}}{{random_letter}}', letters=string.ascii_letters):
```
fcurella: What's the rationale behind this specific default? Shouldn't this argument be mandatory? I can't imagine anyone calling this method without a string to format

malefice: Because `build_docs` will fail if it is omitted entirely, and if you specify a blank value, the current docs that will be generated will not show what it is supposed to do. Do you have any suggestions?

fcurella: We should probably fix `build_docs`, but that's a much bigger project and out of scope. Let's keep this default.

malefice: Got it! It is getting late here, so I will try to fix the errors tomorrow.

fcurella:
Thank you! ✨

PR #1049, opened 2019-11-11
Fix for pyfloat empty randrange (issue #1048)

### LINKED ISSUE #1048: Empty range bug with pyfloat
* Faker version: 2.0.3
* OS: macOS 10.14.6 (Mojave)

Calling `pyfloat` with `min_value` or `max_value` can result in both being set to the same thing, and causing errors. May have been introduced by #995.

### Steps to reproduce
```
for _ in range(1000):
    faker.pyfloat(max_value=9999)
```

You may have to try this a few times, or increase the range

### Expected behavior

Faker should handle this gracefully

### Actual behavior

`ValueError: empty range for randrange() (9999,9999, 0)`

jcardali:
### What this changes

`pyfloat` should not error out anymore due to randomly min/max values.

### What was wrong

`random_int` calls `randrange` which chokes if `min` and `max` are the same.

### How this fixes it

Checks to see if `min_value == max_value` before calling `random_int`, if yes, generate again.

Fixes #1048

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -56,7 +56,7 @@ class Provider(BaseProvider):
             self.random_int(0, sys.float_info.dig - left_digits))
         sign = ''
         if (min_value is not None) or (max_value is not None):
-            left_number = self._safe_randint(min_value, max_value)
+            left_number = self._safe_random_int(min_value, max_value)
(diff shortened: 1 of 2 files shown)


fcurella:
Thank you! ✨

PR #1053, opened 2019-11-14
add Iranian credit_card

abtinmo:
### What does this changes

add localization to credit_card.
add Iranian banks and credit card providers. 

### What was wrong

lack of Iranian prefixes and expire date
10 years for master or visa but 3 years for Iranian banks and credit card providers, and reverse date-time format.

### How this fixes it

by enabling localization with 'fa_IR'. 

Fixes #...


fcurella:
I think the build is erroring out because of [this call to .format()](https://github.com/joke2k/faker/blob/7053bf98712cbaa14238662ee1aad2dd16dbb23a/faker/providers/credit_card/__init__.py#L90-L96)

https://anonbadger.wordpress.com/2016/01/05/python2-string-format-and-unicode/

You may have to patch that to use `%` formatting and see if it works.

abtinmo:
> I think the build is erroring out because of [this call to .format()](https://github.com/joke2k/faker/blob/7053bf98712cbaa14238662ee1aad2dd16dbb23a/faker/providers/credit_card/__init__.py#L90-L96)
> 
> https://anonbadger.wordpress.com/2016/01/05/python2-string-format-and-unicode/
> 
> You may have to patch that to use `%` formatting and see if it works.

it did not worked, i fixed the error with adding u' ' before provider names.

fcurella:
Thank you! ✨

PR #1055, opened 2019-11-15
Improve color provider

### LINKED ISSUE #104: Improve color provider
see: https://github.com/davidmerfield/randomColor

malefice:
### What does this changes

This will add a new color provider method `color()` that allows specifying a hue, a luminosity, and a color format to generate an matching color.

```
>>> from faker import Faker
>>> fake = Faker()

# Create random blue color in RGB format
>>> fake.color(hue='blue', color_format='rgb')
'rgb(5, 56, 175)'

# Create random bright red color in HSV format
>>> fake.color(hue='red', luminosity='bright', color_format='hsv')
'hsv(345, 96, 90)'

# Create random light pink color in HSL format
>>> fake.color(hue='pink', luminosity='light', color_format='hsl')
'hsl(294, 93, 71)'

# Create random color from yellow to green spectrum in hex format
>>> fake.color(hue=[50, 120], luminosity='dark', color_format='hex')
'#3f7c02'
``` 

### What was wrong

There is currently no support for such features.

### Notes

Personally, I prefer to return tuples of HSV, HSL, and RGB values instead of the `color_format(X, Y, Z)` string notation or any string for that matter, so that there is no need for users to parse the string for the values. I followed the current way `rgb_color()` and `rgb_css_color()` do this, but I am not a fan of those methods.

I also think that it would be better to provide more convenient methods and update old ones, so users will not need to remember the different `color_format` values while still being able to just specify 'blue' and 'dark' to get a color of that type.

Fixes #104

### DIFF
--- a/faker/providers/color/color.py
+++ b/faker/providers/color/color.py
@@ -1,0 +1,276 @@
+"""
+Code adapted from:
+- https://github.com/davidmerfield/randomColor  (CC0)
+- https://github.com/kevinwuhoo/randomcolor-py  (MIT License)
+
+Additional reference from:
+- https://en.wikipedia.org/wiki/HSL_and_HSV
+"""
+
+import colorsys
+import math
+import random
+import sys
+
+import six
+
+
+COLOR_MAP = {
+    'monochrome': {
+        'hue_range': [0, 0],
+        'lower_bounds': [
+            [0, 0], [100, 0],
+        ],
+    },
+    'red': {
+        'hue_range': [-26, 18],
+        'lower_bounds': [
+            [20, 100], [30, 92], [40, 89],
+            [50, 85], [60, 78], [70, 70],
+            [80, 60], [90, 55], [100, 50],
+        ],
+    },
+    'orange': {
+        'hue_range': [19, 46],
+        'lower_bounds': [
+            [20, 100], [30, 93], [40, 88], [50, 86],
+            [60, 85], [70, 70], [100, 70],
+        ],
+    },
+    'yellow': {
+        'hue_range': [47, 62],
+        'lower_bounds': [
+            [25, 100], [40, 94], [50, 89], [60, 86],
+            [70, 84], [80, 82], [90, 80], [100, 75],
+        ],
+    },
+    'green': {
+        'hue_range': [63, 178],
+        'lower_bounds': [
+            [30, 100], [40, 90], [50, 85], [60, 81],
+            [70, 74], [80, 64], [90, 50], [100, 40],
+        ],
+    },
+    'blue': {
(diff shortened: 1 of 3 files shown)


faker/providers/color/color.py
```diff
@@ -0,0 +1,276 @@
...
+        if generator:
+            self.random = generator.random
+        else:
+            self.seed = seed if seed else random.randint(0, sys.maxsize)
+            self.random = random.Random(self.seed)
+
+        for color_name, color_attrs in self.colormap.items():
+            lower_bounds = color_attrs['lower_bounds']
+            s_min = lower_bounds[0][0]
+            s_max = lower_bounds[len(lower_bounds) - 1][0]
```
fcurella: isn't this equivalent to `s_max = lower_bounds[-1][0]`?

faker/providers/color/color.py
```diff
@@ -0,0 +1,276 @@
...
+    def get_color_info(self, hue):
+        # Maps red colors to make picking hue easier
+        if 334 <= hue <= 360:
+            hue -= 360
+
+        for color_name, color in self.colormap.items():
+            if color['hue_range'][0] <= hue <= color['hue_range'][1]:
+                return self.colormap[color_name]
+        else:
+            raise ValueError('Value of hue is invalid.')
```
fcurella: can we also show the value itself in the exception message? Something like ``'Value of hue `%s` is invalid.' % hue``

faker/providers/color/color.py
```diff
@@ -0,0 +1,276 @@
...
+
+        elif isinstance(color_input, six.string_types) and color_input in self.colormap:
+            return self.colormap[color_input]['hue_range']
+
+        try:
+            v1, v2 = color_input
+            v1 = int(v1)
+            v2 = int(v2)
+        except (ValueError, TypeError):
+            return [0, 360]
```
fcurella: I'm not sure I'd want it to silently fallback to defaults. I'd rather have it throw an exception if the user is passing invalid values.

malefice: Alright, I will change this behavior.

faker/providers/color/color.py
malefice: This should only happen if something is wrong with `COLOR_MAP` itself, but sure.

faker/providers/color/color.py
malefice: You are right. I wonder why the author of `randomcolor-py` did not just use `[-1]` for this one. I will fix this, and I am sorry I did not notice.

malefice:
The flake8 errors are related to PR #1058.

fcurella:
Looks Good! Thank you so much!

PR #1062, opened 2019-11-22
Fix pyfloat out of min/max range

bryan-brancotte:
Ensure that pyfloat stays within range defined by min max

### What does this changes

Add tests, and shift min and max when they are negative

### What was wrong

When min was lower than 0, former code generated values between [min-1; max], so did for max

### How this fixes it

 * When the min is lower than `0`, we increase it of one (`-6` -> `-5`) as digits are appended later
 * When the max is lower than `0`, we increase it of one (`-6` -> `-5`) as the number before the comma is generated up to `max - 1`

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -61,9 +61,9 @@ class Provider(BaseProvider):
         sign = ''
         if (min_value is not None) or (max_value is not None):
             if max_value is not None and max_value < 0:
-                max_value += 1 # as the random_int will be generated up to max_value - 1
+                max_value += 1  # as the random_int will be generated up to max_value - 1
             if min_value is not None and min_value < 0:
-                min_value += 1 # as we then append digits after the left_number
+                min_value += 1  # as we then append digits after the left_number
             left_number = self._safe_random_int(min_value, max_value)
         else:
             sign = '+' if positive else self.random_element(('+', '-'))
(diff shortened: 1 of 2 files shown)


faker/providers/python/__init__.py
```diff
@@ -60,6 +60,10 @@ def pyfloat(self, left_digits=None, right_digits=None, positive=False,
             self.random_int(0, sys.float_info.dig - left_digits))
         sign = ''
         if (min_value is not None) or (max_value is not None):
+            if max_value is not None and max_value < 0:
+                max_value += 1 # as the random_int will be generated up to max_value - 1
```
fcurella: at least two spaces before the inline comment

tests/test_factory.py
```diff
@@ -474,11 +474,21 @@ def test_pyfloat_in_range(self):
         # tests for https://github.com/joke2k/faker/issues/994
         factory = Faker()
 
-        factory.seed_instance(5)
-        result = factory.pyfloat(min_value=0, max_value=1)
-
-        assert result >= 0.0
-        assert result <= 1.0
+        for i in range(20):
+            for min_value,max_value in [
```
fcurella: add a space: `fro min_value, max_value in [`

bryan-brancotte: thanks !

bryan-brancotte:
flake8 ci fixed, thanks to @fcurella . Only remains the dependency issue with python3.5 which I see in other PR:
> pkg_resources.DistributionNotFound: The 'python-dateutil>=2.4' distribution was not found and is required by the application

fcurella:
Thank you! ✨

bryan-brancotte:
you are welcome

Issue #1060, opened 2019-11-19
zipcode_in_state for NY returns only one result


arielkaluzhny:
* Faker version: 2.0.2
* OS: macOS Mojave (version 10.14.6)

Brief summary of the issue goes here.

### Steps to reproduce
```python
import faker

fake = faker.Faker()
fake.zipcode_in_state("NY")
fake.zipcode_in_state("NY")
```

### Expected behavior
Results from the two calls to `zipcode_in_state` should return two different zipcodes.

### Actual behavior
`zipcode_in_state("NY")` always returns '06390'. (Other states return different zipcodes as expected)

malefice:
I looked into this, and there is only one zipcode value listed for NY as shown:
https://github.com/joke2k/faker/blob/1eb101a81150b1926fb215f0b6dacd94e85886ef/faker/providers/address/en_US/__init__.py#L293

Unfortunately, I am not familiar with US zipcodes, so please consider submitting a pull request if you are.

arielkaluzhny:
> I looked into this, and there is only one zipcode value listed for NY as shown:
> https://github.com/joke2k/faker/blob/1eb101a81150b1926fb215f0b6dacd94e85886ef/faker/providers/address/en_US/__init__.py#L293
> 
> Unfortunately, I am not familiar with US zipcodes, so please consider submitting a pull request if you are.

Yes thank you, I looked into this as well. I'm planning on submitting one soon!

crd:
Should be considered closed by #1064

fcurella:
Thank you @crd!

PR #1064, opened 2019-11-25
Change NY zipcode range

arielkaluzhny:
### What does this change
 Add a range of zipcodes for New York state (SOURCE: https://www.zip-codes.com/state/ny.asp)

### What was wrong
There was only one NY zipcode available, making every `zipcode_in_state("NY")` return the same thing. **(Issue link: https://github.com/joke2k/faker/issues/1060)** 

Affects any version after zipcodes were added (commit from 01/18/2019 47f15768aecda50962f2d52dea1af7b510d04b56)

### How this fixes it
Now `zipcode_in_state("NY")` should return a random zipcode within the tuple range, as with other states.


fcurella:
Thank you! ✨

crd:
Closes #1060

PR #1069, opened 2019-12-05
Add provider methods for generating popular file types 

### LINKED ISSUE #603: Feature Request: Fake Binary Content
Let's say I want to write unit tests for my web service that has a file upload feature.
It would be great if I can fake random binary content of popular binary file types, especially images and documents, like:

- *.jpg
- *.pdf
- *.doc/docx

I think image uploading is probably the most common type of uploaded binary content, so it may be a good place to start.

**Why would someone need this?**
This could be great mainly for testing file upload APIs for several reasons.
It's way easier if the unit tests can generate fake binary content, so they will not depend on actual fixed binary files (and syncing them between teammates, YUCK).

**Some use case examples:**
- A user uploads any format of an image (jpg, png, bmp, etc.) and the server needs to recognize the image type (by its binary structure, headers etc.) and store/convert/compress/whatever according to it.
- A user uploads a document in a popular format (.pdf, .docx, .pages) and the server needs to read the content for parsing/analysis/whatever.

malefice:
Aims to fix #603 

Note:
Currently, I only wrote provider methods for zip and tar files. I am also in the process of finishing provider methods for delimiter-separated values (DSV's) and popular variants like CSV (comma), TSV (tab), and PSV (pipe), and yes, the DSV's support calling other provider methods for data columns.

The aforementioned are things I can write using only the standard library, but for images, MS office files, and pdfs, I am still waiting for feedback whether the libraries used should be optional dependencies or not.

### DIFF
--- a/faker/providers/misc/__init__.py
+++ b/faker/providers/misc/__init__.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 import hashlib
 import io
 import string
+import tarfile
 import uuid
 import sys
 import zipfile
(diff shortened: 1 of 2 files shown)


fcurella:
LGTM! Thank you so much! ✨

PR #1072, opened 2019-12-06
Add NIP generator

IlfirinPL:
### What does this changes

Add NIP generator

Number of tax identification.
Polish: Numer identyfikacji podatkowej (NIP)


faker/providers/person/pl_PL/__init__.py
```diff
@@ -774,3 +774,26 @@ def pwz_nurse(self, kind='nurse'):
...
+        """
+        Returns 10 digit of Number of tax identification.
+        Polish: Numer identyfikacji podatkowej (NIP).
+
+        https://pl.wikipedia.org/wiki/NIP
+
+        """
+        nip = []
+        for _ in range(9):
+            nip.append(self.random_digit())
```
malefice: The last three lines can be expressed more succinctly with a very slight performance improvement. Use a list comprehension like so:

```
nip = [self.random_digit() for _ in range(9)]
```

faker/providers/person/pl_PL/__init__.py
```diff
@@ -774,3 +774,26 @@ def pwz_nurse(self, kind='nurse'):
...
+
+        weights = (6, 5, 7, 2, 3, 4, 5, 6, 7)
+        check_sum = sum(d * w for d, w in zip(nip, weights)) % 11
+
+        if check_sum % 11 == 10:
+            nip.append(0)
```
malefice: I used [google translate](https://translate.google.com/#view=home&op=translate&sl=auto&tl=en&text=Dziesi%C4%85ta%20cyfra%20NIP%20jest%20cyfr%C4%85%20kontroln%C4%85%2C%20obliczan%C4%85%20wed%C5%82ug%20poni%C5%BCszego%20algorytmu%3A%0A%0APomno%C5%BCy%C4%87%20ka%C5%BCd%C4%85%20z%20pierwszych%20dziewi%C4%99ciu%20cyfr%20odpowiednio%20przez%20wagi%3A%206%2C%205%2C%207%2C%202%2C%203%2C%204%2C%205%2C%206%2C%207%2C%0AZsumowa%C4%87%20wyniki%20mno%C5%BCenia%2C%0AObliczy%C4%87%20reszt%C4%99%20z%20dzielenia%20przez%2011%20(operacja%20modulo%2011).%0ANIP%20jest%20tak%20generowany%2C%20aby%20nigdy%20w%20wyniku%20tego%20dzielenia%2C%20jako%20reszta%2C%20nie%20uzyska%C4%87%20liczby%2010.%20Zgodnie%20z%20tym%20algorytmem%20numer%20000-000-00-00%20jest%20prawid%C5%82owy%2C%20ale%20nie%20ma%20sensu.%20Dla%20ci%C4%85gu%20cyfr%20123-456-78-90%20nie%20mo%C5%BCna%20dobra%C4%87%20cyfry%20kontrolnej%2C%20by%20wygenerowa%C4%87%20prawid%C5%82owy%20NIP.), and it said that the check sum value cannot be 10. If that is truly the case, then the current implementation will generate invalid numbers.

Instead of calling this function recursively till the check sum does not equal to 10 (~9.1% chance of happening per attempt), I suggest using this algorithm instead when the check sum ends up being 10:

1. Increment/decrement a random digit by 1 (whichever is possible) 
2. Add/subtract the corresponding weight for that digit to the current check sum (which will be 10)
3. Apply modulo 11 to the new check sum, and the result will be the new check digit

For example, the number `123456789` will have a check sum of 10. If you want to alter the first digit from the left, say we increment it by 1, our number will become 223456789. The first digit's weight is 6, and since we incremented the first digit, we add it such that new check digit will be `(10 + 6) % 11 = 5`, so the new NIP will be `2234567895`.

faker/providers/person/pl_PL/__init__.py
IlfirinPL: I agree with your statement, change done

faker/providers/person/pl_PL/__init__.py
IlfirinPL: Do you think it would be beneficial do double check result NIP with code, for validation 
https://pl.wikibooks.org/wiki/Kody_%C5%BAr%C3%B3d%C5%82owe/Implementacja_NIP
?

malefice: I had this nagging feeling that I was forgetting something, and thank you for reminding me. You will need to write some tests. This is where the validation will be performed to ensure that `nip()` is generating valid output.

In your tests, you can probably generate 100 samples and validate each of them. I chose 100, because if any 9 digit combination has ~9.1% chance of returning a check sum of 10, then there is less than a 0.01% chance that all 100 samples will pass validation should the algorithm be flawed.

IlfirinPL: I have added tests, and notice that actual there still is bug with generation, I will have to look deeper for this issue

faker/providers/person/pl_PL/__init__.py
```diff
@@ -774,3 +788,26 @@ def pwz_nurse(self, kind='nurse'):
...
+
+        """
+        nip = [self.random_digit() for _ in range(9)]
+
+        weights = (6, 5, 7, 2, 3, 4, 5, 6, 7)
+        check_sum = sum(d * w for d, w in zip(nip, weights)) % 11
+
+        if check_sum % 11 == 10:
+            position = self.random_int(0, 8)
+            nip[position] = (nip[position]+1) % 10
```
malefice: > Increment/decrement a random digit by 1 (whichever is possible)

If you look at one of my comments, this is why I said "whichever is possible". Your code is failing, because you took a shortcut here. What this line does is essentially loop around from 9 back to 0. If you do that, you effectively performed a decrement of 9, and the check sum actually decreases by 9 times the weight for the digit.

faker/providers/person/pl_PL/__init__.py
```diff
@@ -21,6 +21,20 @@ def checksum_identity_card_number(characters):
     return check_digit
 
 
+def validate_nip(nip_str):
```
malefice: It should be OK to put this in the `TestCase`.

tests/providers/test_person.py
```diff
@@ -251,6 +251,10 @@ def test_pwz_nurse(self, mock_random_digit, mock_random_int):
         assert self.factory.pwz_nurse(kind='nurse') == '4534567P'
         assert self.factory.pwz_nurse(kind='midwife') == '0317512A'
 
+    def test_nip(self):
+        for _ in range(100):
+            assert pl_validate_nip(self.factory.nip())
```
malefice: Make sure to rebase your work against the current master. My PR #1079 was just merged, and we made some changes with variable names and stuff. Basically, use `self.fake` instead of `self.factory`.

IlfirinPL: fix that issue, also added valid tax code offices, its long list but better to improve it

faker/providers/person/pl_PL/__init__.py
IlfirinPL: I think now everything should be OK, corner case handled. Maybe so code optimization is required, waiting for comments

IlfirinPL:
I think now everything should be OK, corner case handled. Maybe so code optimization is required, waiting for comments

faker/providers/person/pl_PL/__init__.py
```diff
@@ -774,3 +774,798 @@ def pwz_nurse(self, kind='nurse'):
         kind_char = 'A' if kind == 'midwife' else 'P'
 
         return '{:02d}{}{}'.format(region, ''.join(map(str, core)), kind_char)
+
+    tax_office_codes = (
```
malefice: I think we can do away with the comments for each tax office code, and then just add the source of these tax office codes in the `nip()` method. That way the list can be compressed to span just a few lines.

faker/providers/person/pl_PL/__init__.py
```diff
@@ -774,3 +774,798 @@ def pwz_nurse(self, kind='nurse'):
...
+
+    def nip(self):
+        """
+        Returns 10 digit of Number of tax identification.
+        Polish: Numer identyfikacji podatkowej (NIP).
+
+        https://pl.wikipedia.org/wiki/NIP
+        """
+
+        nip = [int(i) for i in self.random_choices(self.tax_office_codes, 1)[0]]
```
malefice: The call to `self.random_choices(self.tax_office_codes, 1)[0]` can be replaced by something simpler: `self.random_element(self.tax_office_codes)`.

IlfirinPL: change done

faker/providers/person/pl_PL/__init__.py
IlfirinPL: comments removed, still long list :)

malefice: It beats having to scroll through an additional 800 lines though.

malefice:
@fcurella do you have any feedback on this PR?

fcurella:
LGTM! Thanks @IlfirinPL  and thank you @malefice for reviewing! ✨

PR #1073, opened 2019-12-07
🍁 Add en-CA postcode_in_province() method

oeuftete:
### What does this change

* Add `postcode_in_province` for en-CA addresses
* Remove extra `New Brunswick` from en-CA provinces
* Use tighter regex for en-CA postcode tests
* Add en-CA address source link

### What was wrong

* `postcode_in_province` seemed like a nice thing to have for parity with the US.
* The tests were too lax for the allowable en-CA postal codes.
* `New Brunswick` was listed twice.  Once is more than enough.


fcurella:
Thank you! ✨

PR #1075, opened 2019-12-09
Add provider methods for dsv files

malefice:
### What does this change

Allows generation of delimiter-separated values with a mostly same API to `csv.writer`

Related to #603


fcurella:
Nice work! Thank you! ✨

PR #1079, opened 2019-12-12
Test Suite Housekeeping (Part 1 maybe?)

### LINKED ISSUE #773: Pipeline: Coverage report does not reflect what actually changed in a PR

malefice:
### What does this change

First, this will remove all instances where the words `factory` or `generator` are used to refer to any operation involving a `Faker` instance. Since `Faker` is now an actual class, using those words is not only technically wrong, but we would also want to discourage users from directly using `Factory.create` and `Generator`.

As part of this process, all (local and instance) variables that are assigned a `Faker` instance have been renamed to either `fake` or `<something>_fake`. The latter is for cases where multiple `Faker` instances are required in a test. I used the word `fake`, since it is the one used in the `README`, and it has no potential conflict with any module name (like it would have been if faker was used instead).

Second, all `Faker` instances created in the `setUp` method of each `TestCase` will now be seeded. Specifically, `Faker.seed(0)` is called. This is to greatly minimize, if not to totally eliminate, any test coverage fluctuation caused by varying RNG seed values across test runs as mentioned in #773. Of course, any seeding performed in the individual test methods will still override the initial seeding.

Lastly, I got a little carried away and did some minor changes, and later on I realized it was outside of the scope of what I initially intended to do, but I already made the commits. Some of those changes are:

   - DRY-ifying some of the tests for credit card provider
   - Moving tests like `test_<provider_method>_<locale>` to a locale specific `TestCase` of its own
   - Removing unnecessary direct calls to provider methods

     ```
     from faker import Faker
     fake = Faker()

     # Removed these two lines
     provider = fake.provider('faker.providers.barcode')
     provider.ean13()

     # When this would have sufficed
     fake.ean13()
     
     # But not if the test involves something only
     # accessible from the provider itself
     upc_ae = provider._upc_ae()
     ```

If you agree with everything, then I can probably modify the existing coding style doc to include a section on writing tests.

Fixes #773, at least the part where RNG is a huge factor

### DIFF
--- a/tests/providers/__init__.py
+++ b/tests/providers/__init__.py
@@ -1,37 +1,35 @@
 import unittest
 from faker import Faker
-from faker.providers import BaseProvider
 
 
 class BaseProviderTestCase(unittest.TestCase):
     def setUp(self):
-        self.factory = Faker('en_US')
+        self.fake = Faker('en_US')
         Faker.seed(0)
-        self.provider = BaseProvider(generator=self.factory)
 
     def test_random_digit_or_empty(self):
-        ret = self.provider.random_digit_or_empty()
+        ret = self.fake.random_digit_or_empty()
         assert isinstance(ret, int)
-        assert ret >= 0 and ret <= 9
+        assert 0 <= ret <= 9
 
         Faker.seed(1)
-        assert self.provider.random_digit_or_empty() == ''
+        assert self.fake.random_digit_or_empty() == ''
 
     def test_random_digit_not_null_or_empty(self):
-        ret = self.provider.random_digit_not_null_or_empty()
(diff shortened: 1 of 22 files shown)


fcurella:
Thank you so much! Awesome work!

cookieMonsterDance

PR #1081, opened 2019-12-12
Remove parenthesis from German city name

jerr0328:
### What does this changes

There was a random parenthesis in one of the German city names that didn't belong.

### What was wrong

There was a city called "Main-Höchst)"

### How this fixes it

Removed paren, change it to be just "Main-Höchst"


fcurella:
Thank you! ✨

PR #1082, opened 2019-12-17
Add identity number for tr_TR

canadiyaman:
### What does this changes

Adds valid identity number generator for Turkish people. (ssn)

### What was wrong

When someone try to get turkish identity number, generator was create invalid identity number.

### How this fixes it

The government of Turkey use unique identity number for all citizen, these numbers creates with pattern. Pattern is: 

-  Identity number contains 11 numbers,
-  First number can't be zero
-  Eleventh number is result of division after sum first 10 number

### DIFF
--- a/tests/providers/test_ssn.py
+++ b/tests/providers/test_ssn.py
@@ -876,11 +876,12 @@ class TestTlPh(TestEnPh):
 
 
 class TestTrTr(unittest.TestCase):
-    num_sample_runs = 1000
+    num_sample_runs = 10
 
-    def setup_factory(self):
-        self.factory = Faker('tr_TR')
-        self.samples = [self.factory.ssn() for _ in range(self.num_sample_runs)]
+    def setUp(self):
+        self.fake = Faker('tr_TR')
+        self.samples = [self.fake.ssn() for _ in range(self.num_sample_runs)]
+        Faker.seed(0)
 
     def first_part_non_zero(self):
         for sample in self.samples:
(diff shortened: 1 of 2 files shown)


tests/providers/test_ssn.py
```diff
@@ -873,3 +873,21 @@ class TestTlPh(TestEnPh):
     def setup_faker(self):
         self.fake = Faker('tl_PH')
         Faker.seed(0)
+
+
+class TestTrTr(unittest.TestCase):
+    num_sample_runs = 1000
+
+    def setup_factory(self):
+        self.factory = Faker('tr_TR')
```
fcurella: could you please follow the patters that other tests are using? Sepcifically:
* use `setUp` instead of `setup_factory`
* use `self.fake` instead of `self.factory`
* seed the faker with `Faker.seed(0)`

```python
class TestTrTr(unittest.TestCase):
    num_sample_runs = 1000
    def setUp(self):
        self.fake = Faker('tr_TR')
        Faker.seed(0)
```

tests/providers/test_ssn.py
```diff
@@ -873,3 +873,21 @@ class TestTlPh(TestEnPh):
     def setup_faker(self):
         self.fake = Faker('tl_PH')
         Faker.seed(0)
+
+
+class TestTrTr(unittest.TestCase):
+    num_sample_runs = 1000
```
fcurella: Are `1000` iterations really necessary?

tests/providers/test_ssn.py
canadiyaman: Okey, I got it.

tests/providers/test_ssn.py
canadiyaman: Changed as 10

fcurella:
Thank you! ✨

Issue #1083, opened 2019-12-17
Revamping the Documentation


malefice:
With the current way the provider method documentation is generated, the result lacks details and examples that will showcase the flexibility and capability of each provider method. I mentioned in #1074 about the idea of using `sphinx.ext.autodoc` to generate docs, and I already have some ideas on how to implement this, but I want to hear some thoughts and get some feedback.

Since `build_docs` already traverse through each `Provider` class and its methods, we can change `build_docs` to write something like this for each provider class. This will autogenerate docs from the individual docstrings (class level and method level). We can also use `sphinx.ext.viewcode` to put links to the source code. 

```
.. autoclass:: {{provider_class}}
   :members: {{comma_separated_list_of_provider_methods}}
   :inherited-members:
   :undoc-members:
```

As for sample code, we can probably use the `autodoc-process-docstring` hook to parse through examples in the docstring, and basically do what `build_docs` currently already does when generating dummy samples. The huge difference is that in this case, we should be able to write as many examples as we want, and let the docs building process evaluate and prettify the output. Of course, it is easier said than done, but I think it is well worth the effort that will solve all documentation woes.

Of course if we go ahead with this style, we will need to write good docstrings for all existing provider methods, and we will also require contributors to write good docstrings moving forward. As for docstrings, we will also have to set a certain standard on how they should be written, e.g. reST vs Google vs numPy, and the format of the examples so we can write the parser.

Anyway thoughts and feedback?

fcurella:
Thank you so much for looking into this @malefice !

If we're going to require contributors to write docstrings, I'd like to have some kind of CI checking the PRs. Maybe we can add [`pydocstyle`](http://www.pydocstyle.org/en/5.0.1/index.html) to our linting check on travis.

malefice:
Yeah sure, I will look into that. Here is a screenshot of what each provider class can look like (without sample invocations for now).

![image](https://user-images.githubusercontent.com/7114869/71015502-ab0a1b80-212e-11ea-9977-403cb1d7de52.png)

malefice:
So I already have some dirty proof of concept code for this one. In this PoC, I used the `:sample:` keyword(?) followed by the arguments to be passed as the format. If the `random_int` docstring has something like this:
```
    :sample: default
    :sample: min=123
    :sample: max='152='
    :sample: step=1
    :sample: min=124, max=125,
    aaaaa=123456,
    b=12345,
    :sample: min=1, max=100
```

[...]

Issue #1078, opened 2019-12-11
How to set args of  provider `date_this_moth` using cli?


NamelessAshone:
* Faker version: faker 2.0.0
* OS: 4.19.85-1-MANJARO

I try to use provider `date_this_moth` with arg `after_today`.
### Steps to reproduce

In the document:
> [fake argument ...]: optional arguments to pass to the fake (e.g. the profile fake takes an optional list of comma separated field names as the first argument)

So, I try:
> faker date_this_month False,True
> faker date_this_month after_today=True
> faker date_this_month --after_today=True
> faker date_this_month --after_today
> faker date_this_month after_today

All these commands are not working.

malefice:
I did some digging, and none of the commands work as intended, because the current implementation only parses the optional arguments as a list of strings as shown [here](https://github.com/joke2k/faker/blob/94b166f1305cf4d9fb3611a92713403b5e64d439/faker/cli.py#L228), and this list is not processed any further and is just passed to the underlying provider method on a FIFO basis as shown [here](https://github.com/joke2k/faker/blob/94b166f1305cf4d9fb3611a92713403b5e64d439/faker/cli.py#L83).

For example, if the terminal command is:

```
faker date_this_month after_today=True before_today=False
```

Under the hood in Python, this will call:

```
date_this_month(before_today='after_today=True', after_today='before_today=False')
```

And not like so as one would expect:

```
date_this_month(before_today=False, after_today=True)
```

To fix this, the provider method argument list must be converted into some format, say `(args, kwargs)`, with the appropriate data type conversions before being passed to the underlying method. It is easier said than done though, because provider methods vary a lot with respect to the expected arguments (value and type). Your thoughts on this @fcurella?

fcurella:
I think we should be using [`Click`](https://click.palletsprojects.com/en/7.x/) to save us a lot of troubles when parsing the CLI options.

malefice:
Interesting, I will take a look at this.

W1M0R:
Any progress on this issue?

malefice:
@W1M0R, give me some time.

![image](https://user-images.githubusercontent.com/7114869/92889932-3fe53f80-f449-11ea-8fa5-e34a22b001fe.png)

malefice:
@fcurella so I pretty much implemented almost everything using `Click` (including multiple locale support) except for the `--include` option (I am still sorting out some issues), and there are some "roadblocks" I encountered. Before all that here are some screenshots!

[...]

Issue #1063, opened 2019-11-25
Python 2.7 EOL plans (Leaving feedback is encouraged)


malefice:
In a little over a month, the PSF will end support for Python 2. @fcurella, what do you intend to do with this project concerning Python 2 EOL? Do you plan to retain support for a certain duration after the EOL date?

In any case, I can help with making everything Python 3 only, but I think `Faker` should have at least one last "semi-significant", if not significant, release that is Python 2 compatible for the poor souls who are stuck with Python 2. Your thoughts?

fcurella:
Once python 2 is EOL, I would like to tag one last release supporting it
and release a new major one dropping support altogether. People stuck on
2.x can just pin to the last supported version (ie `pip install Faker<4").

I don't see the point into staying py2 compatible any longer than that.
People already had plenty of time to upgrade. If they didn't, it's because
something critical is preventing them from doing it, and prolonging Faker's
upgrade is unlikely to change that.

I appreciate any feedback, I'd like for this to be conversation with
Faker's users.

malefice:
I suggest pinning this issue to maximize visibility. I will change the title as well to solicit feedback from users.

maxnordlund:
A last point release would be nice. In addition I would like to see some sort of deprecation/unsupported warning in that versions README/docs.

I don't think a runtime warning would help though. It would only add insult to injury to those poor souls stuck on 2.x

fcurella:
So far, this is the plan:

1. Once python 2 is EOL, release one last version supporting py2.7
2. Clean up the codebase of py2.7-supporting code (`six` etc.)
3. Release new major version supporting only Python 3.

malefice:
The `faker.utils.datetime_safe` module looks like something that can be totally removed as part of the cleanup. Based on what I have read, this module was created to work around the limitations of `strftime()` when handling dates before 1900. This is no longer the case for the versions of Python 3 `faker` supports. At first glance, other code also does not seem to be using the `datetime_safe` module for that purpose either, at least maybe not anymore.

PR #1084, opened 2019-12-20
Test Suite Housekeeping (Part 2)

malefice:
Just another wave of housekeeping, but this time focuses on improving tests and test coverage. I did not tackle some files like `build_docs` and `cli` because of implied plans of deprecating them as discussed in #1078 and #1083. Part 3 will probably be the clean up for Python 3 only release as discussed in #1063.

### DIFF
--- a/faker/proxy.py
+++ b/faker/proxy.py
@@ -108,9 +108,6 @@ class Faker(object):
         :return: A factory that supports the provider method
         """
 
-        if len(self._factories) == 1:
-            return self._factories[0]
-
         factories, weights = self._map_provider_method(method_name)
         if len(factories) == 0:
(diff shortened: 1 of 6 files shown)


malefice:
I noticed the coverage report was not being displayed for this PR, and apparently it has been that way for the past week according to the coveralls [build history](https://coveralls.io/repos/9103/builds). I checked the travis builds around the time this started happening and saw that the `coveralls` command was failing.

After some digging, I found coveralls-clients/coveralls-python#203. The command was failing because it does not support the very recently released coverage 5.0, so for now, I pinned the version at 4.5.4.

fcurella:
Thank you! ✨

PR #1088, opened 2019-12-29
Add port number gen

pishchalnikov:
### What does this changes

Add network port number generator and test for it


fcurella:
Thank you! ✨

Issue #1074, opened 2019-12-08
Faker.random_number raises exception randomly


bontaurean:
* Faker version: 3.0.0

I advise you to document methods parameters and add some kind of validations.

### Actual behavior
```
>>> fake.random_number(fix_len=True)
44805
>>> fake.random_number(fix_len=True)
9543
>>> fake.random_number(fix_len=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 1, in <listcomp>
  File "...site-packages/faker/providers/__init__.py", line 157, in random_number
    pow(10, digits - 1), pow(10, digits) - 1)
  File "/usr/lib/python3.7/random.py", line 222, in randint
    return self.randrange(a, b+1)
  File "/usr/lib/python3.7/random.py", line 186, in randrange
    raise ValueError("non-integer arg 1 for randrange()")
ValueError: non-integer arg 1 for randrange()
```

malefice:
@bontaurean if the docs are insufficient, the next thing you should try is to check the source. It is open source after all, but admittedly, the docs can use some work, and not all methods have docstrings, but this one does.

Anyway, this happens when `random_digit()`returns a 0 in this [line](https://github.com/joke2k/faker/blob/6894660b8f6cfdd73f99fda1d60274ad9e42eb20/faker/providers/__init__.py#L154). Basically, each call to this method has a 10% chance of failing. This should be easy to fix, so it is a good first issue for aspiring contributors. Feel free to submit a PR with a fix and better tests if you wish to take the initiative. I will only tackle this if there are no takers for quite some time.

@fcurella regarding docs, should we take advantage of sphinx's `autodoc`, write docstrings for existing provider methods, and require good docstrings for all provider methods moving forward?

fcurella:
@malefice I"m not sure how hard would it be to integrate `autodoc`, since the providers' docs are already autogenerated by our custom `documentor.py`. But we should look into that.

PR #1089, opened 2019-12-29
Fix Faker.random_number intermittent exceptions #1074

Jengah:
### What does this change

Uses `self.random_digit_not_null()` instead of `self.random_digit()` to generate a random max length of digits when none are passed to `self.random_number()` method.

Also, raise more aptly worded / obvious ValueError when digits param is less than 1 and fix_len param is set to True.

### What was wrong

The use of `self.random_digit()` to generate a random max length of digits in a random number gives it a 10% chance of equaling 0 and failing with a ValueError when fix_len is set to True.

### How this fixes it

Setting `digits = self.random_digit_not_null()` will always generate a number between 1 and 9 instead of 0 and 9, resolving the case where a user specifies `fix_len=True` and doesn't specify a digits param.

Fixes https://github.com/joke2k/faker/issues/1074

### DIFF
--- a/faker/providers/__init__.py
+++ b/faker/providers/__init__.py
@@ -151,10 +151,15 @@ class BaseProvider(object):
             fixed length number
         """
         if digits is None:
-            digits = self.random_digit()
+            digits = self.random_digit_not_null()
+        if digits < 0:
+            raise ValueError("The digit parameter must be greater than or equal to 0.")
         if fix_len:
-            return self.generator.random.randint(
-                pow(10, digits - 1), pow(10, digits) - 1)
+            if digits > 0:
+                return self.generator.random.randint(
+                    pow(10, digits - 1), pow(10, digits) - 1)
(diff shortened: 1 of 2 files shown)


faker/providers/__init__.py
```diff
@@ -151,10 +151,15 @@ def random_number(self, digits=None, fix_len=False):
             fixed length number
         """
         if digits is None:
-            digits = self.random_digit()
+            digits = self.random_digit_not_null()
+        if digits < 0:
```
Jengah: Should also note that although it seems logical that this should be set to `if digits <= 0:`, this breaks test functionality for the python provider method for `pyfloat`.

Specifically this line: https://github.com/joke2k/faker/blob/16d2e6c343c476add06056d1a1929176f0c6acf2/tests/test_factory.py#L467

fcurella:
Thank you!

PR #1091, opened 2020-01-03
Fixing changelog links to show diffs

### LINKED ISSUE #1090: Changelog links are backwards
* Faker version: >=2.0
* OS:Ubuntu

In the changelog, you aren't able to see commits between tags due to the tags being in the opposite order in the URL. It should be https://github.com/joke2k/faker/compare/{FROM}...{TO}, but is instead https://github.com/joke2k/faker/compare/{TO}...{FROM}

For example: https://github.com/joke2k/faker/compare/v3.0.0...v2.0.5

### Steps to reproduce

1. Go to https://github.com/joke2k/faker/blob/master/CHANGELOG.rst
2. Click any version's link after the affected version, and you'll be greeted with an empty changelog (e.g., https://github.com/joke2k/faker/compare/v2.0.0...v1.0.8)

### Expected behavior

You should be able to see all commits between each tag.

### Actual behavior

You see nothing.

### Additional

If you manually swap the two tags in the URL or in the UI, you can see the commits.

I'd be happy to submit a fix for this if this is, as I expect, unintended behavior.

I don't expect this to be more than manually updating all affected changelog entries and (probably?) updating the bumpversion config.

Ianleeclark:
### What does this changes

All pertinent details can be found in https://github.com/joke2k/faker/issues/1090

In short, the links in the changelog are broken and can't show anything in the diffs since v2.0.0

### What was wrong

In short, the links in the changelog are broken and can't show anything in the diffs since v2.0.0

### How this fixes it
This puts the version tags in the URLs in the correct order to allow showing diffs in github browser

Fixes #1090

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,13 +1,13 @@
 Changelog
 =========
 
-`3.0.0 - 04-December-2019 <https://github.com/joke2k/faker/compare/v3.0.0...v2.0.5>`__
+`3.0.0 - 04-December-2019 <https://github.com/joke2k/faker/compare/v2.0.5...v3.0.0>`__
 --------------------------------------------------------------------------------------
 
 * Breaking change: Add support for multiple locale data generation.
   Thanks @malefice.
 
-`2.0.5 - 03-December-2019 <https://github.com/joke2k/faker/compare/v2.0.5...v2.0.4>`__
+`2.0.5 - 03-December-2019 <https://github.com/joke2k/faker/compare/v2.0.4...v2.0.5>`__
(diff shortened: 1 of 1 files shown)


Ianleeclark:
Something of note, I mentioned updating the changelog in the .bumpversion.cfg file, but it seems this is handled manually. If not, please point me to anything that automates this, otherwise I think this fix staying in place relies on following convention.

fcurella:
@Ianleeclark that's correct. Bumpversion is _not_ managing the changelog. It's all done manually.


Thank you much for fixing this! ✨

PR #1093, opened 2020-01-05
Run pyupgrade across codebase to cleanup old syntax

jdufresne:
Helps move towards idiomatic Python and reduces the number changes when
dropping Python 2 support.

For details, on pyupgrade, see:

https://github.com/asottile/pyupgrade

> A tool (and pre-commit hook) to automatically upgrade syntax for newer
> versions of the language.

Refs #1063


fcurella:
Thank you so much!

PR #1094, opened 2020-01-05
Replace deprecated 'setup.py test' with tox/pytest

jdufresne:
Since setuptools v41.5.0 (27 Oct 2019), the 'test' command is formally
deprecated and should not be used. Now, run the pytest command from tox
as recommended by setuptools.

https://setuptools.readthedocs.io/en/latest/history.html#v41-5-0


fcurella:
Thank you!

PR #1096, opened 2020-01-05
Remove unnecessary "exclude" from flake8 configuration

jdufresne:
These directories either don't exist, aren't linted, or are
automatically excluded. There is no need to specify these.


fcurella:
Thank you so much for all your work!

PR #1097, opened 2020-01-05
Run isort as part of CI

jdufresne:
isort was already configured in commit
93c937505a4009fcd95aa5e84ee44831330d59ff. Now include isort in the tox
matrix and run it during CI.

Fix faker/factory.py to avoid the circular dependency.

Fix isort early to reduce changes when dropping Python 2 support.

Refs #1063


fcurella:
Thank you! ✨

PR #1101, opened 2020-01-16
add pytest plugin

fcurella:
### What does this changes

Adds a pytest fixture for Faker.

### What was wrong

The `pytest-dev` team already made a [pytest-faker](https://github.com/pytest-dev/pytest-faker) plugin.

I think it would make more sense to include this plugin right into Faker. Users will have one less package to install, and the `pytest-dev` team will have one less package to maintain.

### How this fixes it

We are basically vendoring-in [pytest-faker](https://github.com/pytest-dev/pytest-faker) .

Fixes https://github.com/pytest-dev/pytest-faker/issues/11

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -98,13 +98,28 @@ If you need to override ``Faker``'s default locale, you must redefine the
 
     import pytest
 
-    @pytest.fixture(scope="session")
+    @pytest.fixture()
     def faker_locale():
         return "it_IT"
 
 
     def test_faker(faker):
         assert faker.providers[0].__lang__ == faker_locale
+
+If you need to seed the faker instance, override the ``faker_seed`` fixture:
+
+.. code:: python
+
+    import pytest
+
+    @pytest.fixture()
+    def faker_seed():
+        return 0
+
+
+    def test_faker(faker):
+        assert faker.name() == 'Norma Fisher'
+
 
 Providers
 ---------
@@ -402,6 +417,21 @@ the same way. For example:
 Please note that as we keep updating datasets, results are not guaranteed to be
 consistent across patch versions. If you hardcode results in your test, make sure
 you pinned the version of ``Faker`` down to the patch number.
+
+If you are using ``pytest``, you can seed the ``faker`` fixture by overriding
+the ``faker_seed`` fixture:
+
+.. code:: python
+
+    import pytest
+
+    @pytest.fixture()
+    def faker_seed():
+        return 0
+
+
+    def test_faker(faker):
+        assert faker.name() == 'Norma Fisher'
 
 Tests
 -----
--- a/README.rst
+++ b/README.rst
@@ -80,6 +80,32 @@ to ``faker.Generator.format(method_name)``.
     # 'Wellington Koelpin II'
     # 'Ms. Karley Kiehn V'
 
+pytest fixtures
+---------------
+
+``Faker`` includes a ``faker`` fixture for ``pytest``.
+
+
+.. code:: python
+
+    def test_faker(faker):
+        assert isinstance(faker.name(), str)
+
+If you need to override ``Faker``'s default locale, you must redefine the
+``faker_locale`` fixture:
+
+.. code:: python
+
+    import pytest
+
+    @pytest.fixture(scope="session")
+    def faker_locale():
+        return "it_IT"
+
+
+    def test_faker(faker):
+        assert faker.providers[0].__lang__ == faker_locale
+
 Providers
 ---------
 
--- a/faker/contrib/pytest/plugin.py
+++ b/faker/contrib/pytest/plugin.py
@@ -1,6 +1,6 @@
 import pytest
 
-from faker import Factory
+from faker import Faker
 
 
 @pytest.fixture(scope="session")
@@ -14,4 +14,5 @@ def faker_locale():
 @pytest.fixture(scope="session")
 def faker(faker_locale):
     """Faker factory object."""
-    return Factory.create(faker_locale)
+    faker = Faker(faker_locale)
+    return faker
--- a/faker/contrib/pytest/plugin.py
+++ b/faker/contrib/pytest/plugin.py
@@ -3,7 +3,15 @@ import pytest
 from faker import Faker
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture()
+def faker_seed():
+    """Faker seed.
+    None by default which means faker's default random seed.
+    """
+    return None
+
+
+@pytest.fixture()
 def faker_locale():
     """Faker locale.
     None by default which means faker's default locale.
@@ -11,8 +19,10 @@ def faker_locale():
     return None
 
 
-@pytest.fixture(scope="session")
-def faker(faker_locale):
+@pytest.fixture()
+def faker(faker_locale, faker_seed):
     """Faker factory object."""
     faker = Faker(faker_locale)
+    if faker_seed is not None:
+        faker.seed_instance(faker_seed)
     return faker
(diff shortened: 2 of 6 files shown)


malefice:
This does not use the new `Faker` class though, so there will be no multiple locale support, and if the new `Faker` class will be used, it will be awkward to call `Faker.seed()`.

```
from faker import Faker

def test_yuck(faker):
    Faker.seed(0)

def test_eww(faker):
    faker.__class__.seed(0)
```

fcurella:
@malefice good points.

I've pushed some changes. Do you mind taking another look?

malefice:
If I am not mistaken, the `Faker` class instantiation will happen every time the `faker` fixture is used in a function. This can potentially be expensive, more so if coupled with multiple locales. Granted the use case is for an automated test suite, and I highly doubt most users would even use 2 locales at a time, but it is still worth noting for potential optimizations in the future. 

```
$ time python -c "from faker import Faker; f = [Faker() for _ in range(1000)]"

real    0m9.891s
user    0m6.862s
sys     0m2.993s

$ time python -c "from faker import Faker; f = [Faker(['en_US', 'en_PH', 'en_GB']) for _ in range(1000)]"

real    0m29.354s
user    0m20.200s
sys     0m9.055s

```

Other than this minor concern, everything else seems good!

fcurella:
@malefice I'm not sure what do here.

I agree that it would be optimal to instantiate only once. But if we cache the instance per session, how could we provide an easy way to seed the faker on a per-test basis?

cognifloyd:
Can't we just reseed an existing Faker instance?

fcurella:
@cognifloyd The problem with just calling `faker.seed_instance` is that it will re-seed across tests.

Imagine this scenario:

```python
def test_one(faker):
    faker.seed_instance(0)
    faker.name()


def test_two(faker):
    faker.seed_instance(2)
    faker.name()


def test_three(faker):
    faker.name()
```

What's the seed in `test_three`?

Ideally, you'd seed for every test, so I'd like  to 'temporarily' seed the instance just within the scope of a single test.

malefice:
> @malefice I'm not sure what do here.

The concerns I raised in no way critical. They are just ideas for future improvements, and probably something we might want to let users know. As for actually tackling my concerns, maybe we can do something like this in the future (after much more research and extensive testing)?

```
from faker import Faker

@pytest.fixture(scope='session')
def _sessionfaker(faker_locale):
    return Faker(locale=faker_locale)

@pytest.fixture
def faker(request):
    if 'faker_locale' in request.fixturenames:
        locale = request.getfixturevalue('faker_locale') 
        fake = Faker(locale=locale)
    else:
        request.getfixturevalue('_sessionfaker')
    if 'faker_seed' in request.fixturenames:
        seed = request.getfixturevalue('faker_seed')
        fake.seed_instance(seed)
    else:
        fake.seed_instance(0)
    return fake

```

The only time we really need to return a new `Faker` instance is if there is a locale change, so we check if `faker_locale` fixture is requested or not, and the fixture should behave like so:

```
# in test_file1.py

pytest.fixture(autouse=True)
def faker_locale():
    return 'it_IT'

def test_something(faker):
   # faker here will be a new instance with it_IT locale
    assert True
```

```
# in test_file2.py

def test_something_else(faker):
   # faker here will be the session instance
    assert True
```

malefice:
@fcurella , I had some time to look into this, and we can enable a session level faker instance by default with the ability to still seed the instance per test. I think this covers most test use cases. In the oddball situations where one will need to change locales in certain tests, we can provide ways to drop down to function level faker instances.

To change the behavior of the session level instance, one way is to use custom command line options. For example, `pytest --faker-session-locale=ja_JP` will create a session level faker instance that uses `ja_JP` as its locale. On the other hand, `pytest --faker-session-locale=ja_JP,en_US,zh_TW` will use `['ja_JP', 'en_US', 'zh_TW']` as the session locale. Omitting the command line argument will just default to the `DEFAULT_LOCALE` which is `en_US`. We can also specify the default seed to be used in every test function using another command line option.

If command line options are unacceptable, then an alternative is to define session-scoped fixtures that control these values, but these fixtures have to be named differently from the function level ones (e.g. `faker_session_locale`). Users can override the default settings by redefining those fixtures in their top level `conftest.py` and supplying `scope='session', autouse=True`.

Using a `faker_locale` fixture anywhere will automatically cause a function level faker instance to be created for each test function within the scope of the `faker_locale` fixture. Using a `faker_seed` fixture will just seed whatever faker instance will be used regardless if it is session level or function level.

Admittedly, it is more complicated than desired, and while your current implementation may not be "optimal", it is very simple to use. If you prefer the simple route, then I think this PR should be merged. If you think the complexity is worth whatever performance improvement my proposal may bring, then I can submit a PR if you want.

fcurella:
We could keep it simple and just direct users with more complicated scenarios (eg multiple locales) to create their own fixture, eg tell them in the docs to do something like:

```python
import pytest


@pytest.fixture()
def faker_jp():
  return Faker('ja_JP')


@pytest.fixture()
def faker_tw():
  return Faker('zh_TW')
```

I think it's ok if more complicated scenarios are a little harder to achieve. We make the common case easy, and the edge case possible.

malefice:
Alright, so here is what I propose. We provide session level faker instances with sane (but overridable) defaults. This is what our `plugin.py` will look like.

```python
# plugin.py

import pytest

from faker import Faker


@pytest.fixture(scope='session', autouse=True)
def _session_faker(faker_session_locale):
    return Faker(locale=faker_session_locale)


@pytest.fixture(scope='session', autouse=True)
def faker_session_locale():
    return None


@pytest.fixture(scope='session', autouse=True)
def faker_default_seed():
    return 0


@pytest.fixture()
def faker(request, faker_default_seed):
    if 'faker_locale' in request.fixturenames:
        locale = request.getfixturevalue('faker_locale')
        fake = Faker(locale=locale)
    else:
        fake = request.getfixturevalue('_session_faker')

    seed = faker_default_seed
    if 'faker_seed' in request.fixturenames:
        seed = request.getfixturevalue('faker_seed')
    fake.seed_instance(seed=seed)

    return fake
```

The `faker` fixture is function-scoped, but in the absence of a `faker_locale` fixture, this fixture will always return a "hidden" session level faker instance that defaults to `en_US` locale. This session level default locale can be changed by doing this:

```python
# user's top level conftest.py

import pytest

@pytest.fixture(scope='session', autouse=True)
def faker_session_locale():
    # Returning a list of locales here will enable multiple locale support
    return 'it_IT'
```

If however, a user only wants to change locale for some test functions, then declaring this in a submodule's `conftest.py` or within a test file itself will only affect test functions within that submodule or test file.

```python
import pytest

@pytest.fixture(scope=not_a_session_scope, autouse=True)
def faker_locale():
    # Returning a list of locales here will enable multiple locale support
    return 'it_IT'
```

A user may choose to omit `autouse=True`, and only pass the fixture to specific test methods for the same result.

```python
import pytest

@pytest.fixture()
def faker_locale():
    # Returning a list of locales here will enable multiple locale support
    return 'it_IT'


def test_something(faker, faker_locale):
    # faker here will be a new instance using faker_locale
    assert faker.locales == faker_locale

def test_something_else(faker):
    # faker here is still the session level faker
    pass
```

Before I continue any further, this is already the most complex part of this proposal. The seeding stuff is incredibly simple compared to this. Does this look good and simple enough yet still flexible to you?

fcurella:
Thank you so much for taking the time to look into this @malefice !

Your proposal seems like a good balance to me! 👍

malefice:
@fcurella this can be closed now since #1155 was built on top of this PR.

PR #1103, opened 2020-01-22
add Birth Number to czech locale #1099

### LINKED ISSUE #1099: Implementation of person id number for cs_CZ - Czech (rodné číslo)
Can you implement randomizer which will generate a proper person ID number (rodné číslo) for Czech local?

Jengah:
### What does this changes

Adds `birth_number` functionality to Czech locale, which is National Identification Number used by Czech Republic and Slovakia as requested in #1099 

Source: https://en.wikipedia.org/wiki/National_identification_number#Czech_Republic_and_Slovakia

### What was wrong

Added functionality

### How this fixes it

Generates a birth number (who's whole number is divisible by 11) that adheres to the following formats
```
######/###
######/####
```
depending on if you were born before or after Jan 1, 1954

Fixes #1099

### DIFF
--- a/faker/providers/ssn/cs_CZ/__init__.py
+++ b/faker/providers/ssn/cs_CZ/__init__.py
@@ -1,3 +1,5 @@
+from math import ceil
+
 from .. import Provider as BaseProvider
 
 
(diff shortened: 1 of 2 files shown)


faker/providers/ssn/cs_CZ/__init__.py
```diff
@@ -8,10 +8,36 @@ class Provider(BaseProvider):
...
+        year = '%.2d' % (birthdate.year % 100)
+        month = self.random_element(self.national_id_months)
+        day = '%.2d' % birthdate.day
+        if birthdate.year > 1953:
+            sn = self.random_number(4, True)
+        else:
+            sn = self.random_number(3, True)
+        number = int('{}{}{}{}'.format(year, month, day, sn))
+        while number % 11 != 0:
+            number += 1
```
fcurella: instead of iterating, wouldn't be more efficient to calculate this as `number = 11 * (number % 11 + 1)`?

Jengah: Good point, there isn't a need to iterate here. How's `math.ceil(number / 11) * 11`? I've pushed up a commit with that in place of the iteration.

fcurella:
Looks good! Thanks! ✨

PR #1104, opened 2020-01-27
Provider updates for Russian locale

valestel:
### What does this changes

Credit Card, Currency and Date-Time Providers were improved and expanded in Russian locale

### What was wrong

There were errors when generating a name of a credit card's owner: e.g. female first name and male last name, also the whole name was in Cyrillic characters when it's supposed to be romanised.
There were no Russian equivalents for currency names and time zone names.

### How this fixes it
 `translit()` method was added to Person provider to generate a romanised name of a credit card's owner
Most common types of credit cards were added to Credit Card Provider. Now there's also card-issuing bank in the credit card information, as issuers are often mentioned in many Russian texts regarding credit cards.
Currency names in Russian were added. Now Currency Provider has `currency_symbol()` method that generates a graphic symbol of some currency, these symbols are used too.
Russian names for time zones in Date-Time Provider were added to `countries` list. `DAY_NAMES` and `MONTH_NAMES` dictionaries were taken out of `day_of_week()` and `month_name()` methods so they could be used for testing.

All tests provided as well


tests/providers/test_date_time.py
```diff
@@ -716,3 +718,29 @@ def test_month(self):
...
+
+class TestRuRu(unittest.TestCase):
+    """ Tests date_time in the ru_RU locale """
+
+    def setUp(self):
+        self.fake = Faker('ru_RU')
+        Faker.seed(0)
+
+    def test_day(self):
+        for _ in range(100):
```
fcurella: I think looping 100 times might be excessive, as there are only 7 possible values. Let's do `range(50)`

valestel: Okay, I reduced it to 50

fcurella:
Thank you! ✨

PR #1105, opened 2020-02-13
Add persons provider for fr_QC

Lionesk:
### What does this changes

Adds a provider for Québec French names.

### What was wrong

For French name providers, there are only France (FR) and Swiss (CH). French Canadians (called Québecois) have their own set of family names due mostly to [their colonial history](https://en.wikipedia.org/wiki/History_of_Quebec#French_Canada_(1608%E2%80%931759)), and while they share some first names with European French speakers, there are some European names that are seldom used in Québec, and vise-versa.

I am mocking data for an internal project, and the names generated with the `fr_FR` locale are distinctly different from the kinds of names my colleagues and clients would expect.

### How this fixes it

This adds a list of male first names, female first names, and family names that are more in line with the Québecois population.

Fixes #...


fcurella:
Thank you! ✨

PR #1109, opened 2020-02-18
Fixed: region_codes for bic must have 9 digits

MikeVL:
### What does this changes

Brief summary of the changes

### What was wrong

### How this fixes it

Fixes #...


fcurella:
Thank you!

PR #1110, opened 2020-02-18
[Add] pt_PT locale

mustakarhu:
### What does this changes

Add street prefix diversity which are missing for pt_PT.
Add names of streets generated from real street names from the two largest cities of the country (Lisbon and Porto). 

Add new file for pt_PT banking details 

### What was wrong

Previously, street names were generated only based on combinations of given names + surnames. 
Street names can be also dates, historical figures (real or allegorical) or allusive to certain landmarks. 

Bank information was not available for pt_PT as well.


fcurella:
Thank you! Could add tests for `place_name ` and the new bank provider?

mustakarhu:
Done. Should be ok now.

fcurella:
Thank you! ✨

PR #1111, opened 2020-02-19
Add more data for ko_KR address

alstn2468:
### Add more data at` faker/provider/address/ko_KR/__init__.py`

Add more town_suffixes, cities, data and concat uppercase alphabet tuple to bulding_dongs tuple.
Korea building dongs can contain upper case alphabet such as `###아파트 A동`, `###아파트 Z동`


alstn2468:
Repoen pull request after apply flake8.

PR #1112, opened 2020-02-19
Add more data for ko_KR address

alstn2468:
### Add more data at ko_KR address provider

Add more town_suffixes, cities, data and concat uppercase alphabet tuple to bulding_dongs tuple.
Korea building dongs can contain upper case alphabet such as `###아파트 A동`, `###아파트 Z동`

310f2c4 -> Add data
31bfc4f -> Apply flake8


fcurella:
Thank you! ✨

PR #1116, opened 2020-02-26
Add support for generating phone country codes

### LINKED ISSUE #1085: Generation of Country Phone Code
I would like to generate random country phone prefix from here: https://en.wikipedia.org/wiki/List_of_country_calling_codes 
Result should be: +213, +350, +56 etc. 

Is there a way to do this? 
Thanks

perssonsimon1:
### What does this changes

Adds the ability to get a random country code (phone) in the phone provider

### What was wrong

Used to be no support for this

### How this fixes it

Added a function for getting a random code

Fixes #1085


fcurella:
Thank you! ✨

PR #1117, opened 2020-02-26
Formatting option for Brazilian postcodes

### LINKED ISSUE #1086: Can we have a "non formatted" postcode for pt_BR localization?
* Faker version: 3.0.0
* OS: Windows 10 pro (x64)

I'm using Faker to generate postcode using the pt_BR localization. The example on [`address` section](https://faker.readthedocs.io/en/latest/locales/pt_BR.html#faker-providers-address) shows a numeric sequence containing 8 numbers and that was what I expect.

```python
fake.postcode()
# '35285968'
```

Anyway, I'm getting some problems in my tests because the output of this function sometimes returns a formatted postcode. Example:

```python
fake.postcode()
# '35285968'    -- cool :D

fake.postcode()
# '35285-968'    -- sad :(
```

### Expected behavior

I would like to generate postcode exactly hows the documentation says, just numbers, without any formatting. As a temporary solution, I'm using a `replace` to remove any `-` present, but I would like to have a cleaner solution using just Faker. Maybe this "formatted" version can be specified by an argument in the `postcode` function. 

### Actual behavior

Getting a non deterministic postcode, I don't know which type of postalcode format I will get.

perssonsimon1:
### What does this changes

Adds the ability to get raw Brazilian postcodes, i.e without the '-' by setting the new option `formatted` to `false`

### What was wrong

Previously the provided returned formatted and non-formatted postcodes at random.

### How this fixes it

Added an option to the provider to only use the raw template format

Fixes #1086

### DIFF
--- a/faker/providers/address/pt_BR/__init__.py
+++ b/faker/providers/address/pt_BR/__init__.py
@@ -295,6 +295,12 @@ class Provider(AddressProvider):
         :example 'Serra'
         """
         return self.random_element(self.bairros)
+    
+    def postcode(self, formatted=True):
+        postcode_raw_formats = ('########', )
+        postcode_all_formats = postcode_raw_formats + ( '#####-###', )
+        template = postcode_all_formats if formatted else postcode_raw_formats
+        return self.bothify(self.random_element(template))
 
     # aliases
     def neighborhood(self):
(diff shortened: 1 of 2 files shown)


fcurella:
Thank you! ✨

PR #1120, opened 2020-02-27
Added company_vat faker

alitaker:
### What does this changes

Added company_vat faker for locale it_IT, company provider

### What was wrong

company_vat was missing

### How this fixes it

Added company_vat function and luhn checksum calculator


faker/providers/company/it_IT/__init__.py
```diff
@@ -330,6 +330,24 @@ class Provider(CompanyProvider):
...
+        checksum = 0
+        checksum += sum(odd_digits)
+        for d in even_digits:
+            checksum += sum(digits_of(d * 2))
+        return checksum % 10
+
+    @staticmethod
+    def _calculate_luhn(partial_number):
+        check_digit = Provider._luhn_checksum(int(partial_number) * 10)
+        return check_digit if check_digit == 0 else 10 - check_digit
```
fcurella: This line has some trailing spaces. could you delete them?

faker/providers/company/it_IT/__init__.py
```diff
@@ -349,3 +367,12 @@ def bs(self):
             result.append(self.random_element(word_list))
 
         return " ".join(result)
+
+
```
fcurella: There are too many lines between those two methods. Can you remove one?

alitaker: Is it now ok with my latest commit, or do I need to do something else?

faker/providers/company/it_IT/__init__.py
```diff
@@ -330,6 +330,24 @@ class Provider(CompanyProvider):
...
+        odd_digits = digits[-1::-2]
+        even_digits = digits[-2::-2]
+        checksum = 0
+        checksum += sum(odd_digits)
+        for d in even_digits:
+            checksum += sum(digits_of(d * 2))
+        return checksum % 10
+
+    @staticmethod
+    def _calculate_luhn(partial_number):
```
fcurella: Since this logic gets reused a lot, we just added it in https://github.com/joke2k/faker/blob/master/faker/utils/checksums.py.

Could you refactor to use the functions from `utils`?

alitaker: Should be fine now

faker/providers/company/it_IT/__init__.py
```diff
@@ -1,4 +1,5 @@
 from .. import Provider as CompanyProvider
+from faker.utils.checksums import calculate_luhn
```
fcurella: Could you move this import to the top of the file, and put a newline between absolute and relative imports: https://travis-ci.org/github/joke2k/faker/jobs/667152889#L234-L241

fcurella:
Thank you! ✨

PR #1121, opened 2020-02-27
locale pt-BR fixed official cellphone_formats and PhoneNumberProvider without DDD

rodrigondec:
https://en.wikipedia.org/wiki/Telephone_numbers_in_Brazil

All phone numbers must have the DDD

### What does this changes
The variation of phone_number generation on locale pt-BR

#### PhoneNumber formats
the format '####-####' is a bad practice (without DDD)
#### msisdn_formats
the format '55AA#########' is wrong because in the example '55AA**4**12345678' the char at position 4 is only acceptable as a '9' digit else if must not exist this additional char.  
it must be '55AA12345678' or '55AA**9**12345678'
#### cellphone_formats
the format '+55 9#### ####' is wrong because is missing DDD, the correct would be '+55 ## 9#### ####'.

### How this fixes it
It fixes these issues by correcting the formats for random_element generation


tests/providers/test_phone_number.py
```diff
@@ -68,20 +68,27 @@ def setUp(self):
...
+        formats = ('5511', '5521', '5531', '5541', '5551', '5561', '5571', '5581', '5584')
 
         assert msisdn is not None
         assert isinstance(msisdn, str)
-        assert len(msisdn) == 13
+        assert len(msisdn) == 13 or len(msisdn) == 12
         assert msisdn.isdigit()
         assert msisdn[0:4] in formats
+        re.match(r"^[+]?([0-9]{2})?[ ]?[(]?[0]?\d{2}[)]?[ ]?[09]?[ ]?\d{4}[ -]?\d{4}$", msisdn)
+
```
fcurella: per pep8, we only keep one line between methods. Could you remove this empty line?

rodrigondec: Sure, didn't notice it.

It would be a good thing to set up pep8 verification on CI, with flake8 or some other lib

fcurella: We already have it in Travis

fcurella:
Thank you! ✨

PR #1124, opened 2020-03-01
ISSUE-1123 handle leap day year shifts

### LINKED ISSUE #1123: Faker is failing on 29th of February on the leap year
* Faker version: 4.0.1
* OS: ubuntu 18.04

Currently, Faker fails to generate a date_of_birth on 29th of February on the leap year, due to the code in here:
https://github.com/joke2k/faker/blob/master/faker/providers/date_time/__init__.py#L1999

        start_date = now.replace(year=now.year - (maximum_age+1))
        end_date = now.replace(year=now.year - minimum_age)

### Steps to reproduce

Execute a faker for date of birth on 29th of February

### Expected behavior

Works as usual

### Actual behavior

Raises exception: ValueError: day is out of range for month

mmcmahon:
### What does this changes

This PR changes the date_of_birth to not fail when trying to generate a dafe DOB on a leap day.

### What was wrong

When Faker attempts to generate a date_of_birth on 29th of February on the leap year it fails  if it tries to return a year that is not also a leap year.

### How this fixes it

This patch fixes the issue by extracting the year adjust to its own function that catches the ValueError and if the starting date is YYYY-02-29 it will adjust the date to YYYY-03-01.

Fixes #1123

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -29,6 +29,29 @@ def timestamp_to_datetime(timestamp, tzinfo):
     return pick
 
 
+def change_year(current_date, year_diff):
+    """
+    Unless the current_date is February 29th, it is fine to just subtract years.
+    If it is a leap day, and we are rolling back to a non-leap year, it will
+    cause a ValueError.
+    Since this is relatively uncommon, just catch the error and roll forward to
+    March 1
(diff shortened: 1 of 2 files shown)


fcurella:
Nice catch! Thank you! ✨

Issue #1133, opened 2020-03-19
Add support for Indian Aadhaar numbers


curiousElf:
* Faker version: 4.0.2
* OS: Ubuntu 18.04

Proposal to add support for Indian Aadhaar numbers. Aadhaar numbers are a 12 digit identifier for Indian residents which can be used for proving identity and for availing Government welfare benefits.

Official website: https://uidai.gov.in/my-aadhaar/about-your-aadhaar.html

curiousElf:
I have already made the changes for this. Will be pushing the PR soon

curiousElf:
Closing this - as I am opening another issue with the PR

PR #1134, opened 2020-03-19
Adds support for Indian Aadhaar numbers (#1133)

curiousElf:
### What does this changes
Adds support for Indian [aadhaar](https://en.wikipedia.org/wiki/Aadhaar) numbers . 

Brief summary of the changes:

Aadhaar number generation has been added within the en_IN locale which is Indian English. This locale is the most commonly used one in India. The locale did not exist previously in Faker.

As we needed to use a Luhn  check - moved an existing implementation from the sv_SE ssn provider to utils along with additional tests


curiousElf:
Need to clean up the github issue for this

PR #1136, opened 2020-03-19
Add dga by date

pishchalnikov:
### What does this changes

Add Domain Generator Algorithm by date method and test for it


fcurella:
Good idea! Thank you!

PR #1137, opened 2020-03-19
#1131 Fixes issue where pydict provider was not returing the correct …

### LINKED ISSUE #1131: pydict(variable_nb_elements=False) returns dicts with varying number of elements
* Faker version: 4.0.1
* OS: OS X 10.15.3

pydict with `variable_nb_elements=False` still has varying number of elements.

### Steps to reproduce

```
from faker import Faker

fake = Faker()

nb = 80
for _ in range(30):  # nb is large and the faker function is called 30 times to increase the probability of reproducing the bug.
    fake_dict = fake.pydict(nb_elements=nb, variable_nb_elements=False)
    nb_reality = len(fake_dict)
    if nb_reality != nb:
        print(nb_reality)
```

### Expected behavior

The length of the dictionary should always be `nb`. 

### Actual behavior

It is often shorter as the keys are generated with [`self.generator.words(nb_elements)`](https://github.com/joke2k/faker/blob/master/faker/providers/python/__init__.py#L166). If the keys are not unique, the element is dropped.

mstellon:
…number of elements.

### What does this changes

Changes python provider pydict word generator call

### What was wrong

Word generator was not returning unique words causing the returned dict to not equal the number of elements defined by nb_elements

### How this fixes it

Add unique=True to the words call to ensure no duplicate words are returned

Fixes #1131

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -163,7 +163,7 @@ class Provider(BaseProvider):
             nb_elements = self.randomize_nb_elements(nb_elements, min=1)
 
         return dict(zip(
-            self.generator.words(nb_elements),
+            self.generator.words(nb_elements, unique=True),
             self._pyiterable(nb_elements, False, *value_types),
         ))
 


fcurella:
Thank you! ✨

PR #1138, opened 2020-03-20
Adds support for Indian Aadhaar numbers

curiousElf:
### What does this changes
Adds support for Indian Aadhaar numbers
Wiki: https://en.wikipedia.org/wiki/Aadhaar

[Official Aadhaar site](https://uidai.gov.in/my-aadhaar/about-your-aadhaar.html)

## Brief summary of the changes
This change also introduces the en_IN locale - the one in which the aadhaar provider has been made available

As we needed to use a Luhn  check - moved an existing implementation
to utils along with additional tests


fcurella:
Thank you! ✨

PR #1143, opened 2020-03-30
Close #525. Make `value_types` a keyword argument

### LINKED ISSUE #525: python providers should use value_types as a keyword argument

slafs:
### What does this changes

In the `python` provider, `value_types` arguments are changed to be non-variadic.

An additional `allowed_types` variadic argument has been added to achieve backward compatibility. The deprecation plan is:

    1. on the next major version, promote the `PendingDeprecationWarning` to a `DeprecationWarning`

    2. on the second next major version, remove `allowed_types` completely.

### What was wrong

    * The method signatures are ambiguous and non intuitive. #366

    * Tools like `factoryboy` can't easily integrate with Faker. #525, [FactoryBoy/factory_boy#456](https://github.com/FactoryBoy/factory_boy/pull/456)

### How this fixes it

It is now possible to pass the value types either as multiple arguments, or as a single argument which could be a list or tuple.

Fixes #525

This is a rebased version of https://github.com/joke2k/faker/pull/916 (with removed redundant `test_pyint` method).

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -1,5 +1,6 @@
 import string
 import sys
+import warnings
 
 from decimal import Decimal
 
@@ -7,6 +8,23 @@ from .. import BaseProvider
 
 
 class Provider(BaseProvider):
+    default_value_types = (
+        'str', 'str', 'str', 'str', 'float', 'int', 'int', 'decimal',
+        'date_time', 'uri', 'email',
+    )
+
+    def _check_signature(self, value_types, allowed_types):
+        if value_types is not None and not isinstance(value_types, (list, tuple)):
+            value_types = [value_types]
+            warnings.warn(
+                'Passing value types as positional arguments is going to be '
+                'deprecated.  Pass them as a list or tuple instead.',
+                PendingDeprecationWarning,
+            )
+        if value_types is None:
+            value_types = ()
+        return tuple(value_types) + allowed_types
+
     def pybool(self):
         return self.random_int(0, 1) == 1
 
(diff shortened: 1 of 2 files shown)


fcurella:
ref https://github.com/FactoryBoy/factory_boy/pull/456

fcurella:
I'd appreciate if the FactoryBoy could take a look and let me know if it's acceptable for them. My plan is to merge it next week unless there's any objection.

francoisfreitag:
Thank you for pinging!

This patch makes sense to me and addresses https://github.com/FactoryBoy/factory_boy/issues/387.
It removes the use case that prompted https://github.com/FactoryBoy/factory_boy/pull/456, allows to always use keyword arguments when delegating to Faker. It is more explicit in the factory definition. Looks like all providers now accept parameters as keyword arguments. :tada:

fcurella:
Thanks for checking, @francoisfreitag !

jeffwidman:
Thanks @fcurella and @slafs!!

PR #1145, opened 2020-04-02
Implements __dir__ method to Faker proxy

douglasfarinelli:
### What does this changes

Implements __dir__ magic method to Faker proxy. This allows when trying to list/show all methods of all providers in IPython, for example:

image


fcurella:
Looks great! Thank you! ✨

PR #1149, opened 2020-04-04
Uppercased HEX colors for all localized color providers.

mondeja:
### What does this changes

Uppercased HEX colors on `all_colors` class properties for all localized color providers, to maintain consistency.

### What was wrong

I've found four HEX colors in some color providers that have been defined lowercased.


fcurella:
Nice catch! Thank you! ✨

PR #1150, opened 2020-04-05
Support for UK counties.

neilav:
### What does this changes

Adds support for UK counties to the GB address locale.


fcurella:
Thank you! ✨

PR #1155, opened 2020-04-17
Add pytest plugin

malefice:
### What does this changes

Adds a pytest fixture for Faker. This is based on #1101 with improvements discussed in that PR. 

### What was wrong

Same as initially stated in #1101 

### How this fixes it

On top of vendoring-in [pytest-faker](https://github.com/pytest-dev/pytest-faker), pytest version is being upgraded, and we are adding some improvements for common use cases while maintaining configurability for more uncommon usage. See #1101 for more details.

Fixes pytest-dev/pytest-faker#11, #1101

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -80,6 +80,12 @@ to ``faker.Generator.format(method_name)``.
     # 'Wellington Koelpin II'
     # 'Ms. Karley Kiehn V'
 
+Pytest fixtures
+---------------
+
+``Faker`` also has its own ``pytest`` plugin which provides a ``faker`` fixture you can use in your
+tests. Please check out the `pytest fixture docs` to learn more.
+
 Providers
 ---------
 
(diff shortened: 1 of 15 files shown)


malefice:
@fcurella before I write the docs, can you please review my work so far?

A somewhat significant change is the need to break pytest runs into two. Since we also need to test `faker_session_locale` and its impact on session level `Faker` instances, there needs to be two runs which is reflected in the updated `tox.ini`.

The first run is the usual pytest run of the whole suite excluding the `faker` fixture tests that explicitly need an overridden `faker_session_locale`. The tests excluded from the first run are then tested on the 2nd run exclusively.

fcurella:
Thank you @malefice ! Looks _really_ good! ✨

malefice:
@fcurella docs have been included. I also changed how tests were run, because I was not comfortable with the situation where running `pytest` (no command line flags, options, arguments) will always fail. I am not comfortable with the idea that others should look at `tox.ini` just to see how to run the test suite, so I changed it in such a way that running `pytest` will instead run the default test suite, and the tests that require a separate pytest session will be skipped atuomatically. Those skipped tests can be ran by explicitly specifying a path and using a command line flag.

fcurella:
Thank you so much @malefice , imporessive work!

I 100% agree with you, just running `pytest` should pass by default.

PR #1160, opened 2020-04-21
Minor fix and addendum to pytest fixture docs

malefice:
### What does this changes

This will fix a minor presentation issue in the pytest fixture docs as well as add some clarification/option regarding seeding.


fcurella:
Thank you! ✨

PR #1164, opened 2020-04-23
providers.misc.uuid with `cast_to=None` case to return `UUID` object

### LINKED ISSUE #1163: Why does faker.misc.uuid has `cast_to=str` by default?
# I'm willing to open PR
should this issue be accepted

* Faker version: latest
* OS: N/A

`faker.misc.uuid` has a arg `cast_to` with `str` as default.  So the returned value is always string.
https://github.com/joke2k/faker/blob/9f48a5c6a251b2b8d9f84f65472f946b85b9a923/faker/providers/misc/__init__.py#L95

I want to have the `uuid.UUID` object returned. 

Calling with `cast_to=UUID` again raises a Error on UUID lib (of course it would). Since it's returned `UUID(UUID([...]))`from faker with this

image

So I would have to cast it like this `UUID(str(UUID([...])))` to be funcional.

### Steps to reproduce

1. `faker.uuid(cast_to=str)`
1. `faker.uuid(cast_to=uuid.UUID)`

### Expected behavior
I should be able to call  `faker.uuid(cast_to=None)` and have the pure `UUID` object returned.

### Actual behavior
Faker forces object `UUID` to be cast.

rodrigondec:
### What does this changes

providers.misc.uuid now doesn't have to always cast to another type.

### What was wrong

providers.misc.uuid would always cast to another type.

### How this fixes it

Added None arg verification.

Fixes #1163

### DIFF
--- a/faker/providers/misc/__init__.py
+++ b/faker/providers/misc/__init__.py
@@ -92,6 +92,9 @@ class Provider(BaseProvider):
         By default, ``cast_to`` is set to ``str``.
 
         May be called with ``cast_to=None`` to return a full-fledged ``UUID``.
+
+        :sample:
+        :sample: cast_to=None
         """
         # Based on http://stackoverflow.com/q/41186818
         generated_uuid = uuid.UUID(int=self.generator.random.getrandbits(128), version=4)
--- a/faker/providers/misc/__init__.py
+++ b/faker/providers/misc/__init__.py
@@ -87,12 +87,17 @@ class Provider(BaseProvider):
         return res.hexdigest()
 
     def uuid4(self, cast_to=str):
-        """Generate a random UUID4 object and cast it to another type using a callable ``cast_to``.
+        """Generate a random UUID4 object and cast it to another type if specified using a callable ``cast_to``.
 
         By default, ``cast_to`` is set to ``str``.
+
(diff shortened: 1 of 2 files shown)


faker/providers/misc/__init__.py
```diff
@@ -87,12 +87,17 @@ def sha256(self, raw_output=False):
         return res.hexdigest()
 
     def uuid4(self, cast_to=str):
-        """Generate a random UUID4 object and cast it to another type using a callable ``cast_to``.
+        """Generate a random UUID4 object and cast it to another type if specified using a callable ``cast_to``.
 
         By default, ``cast_to`` is set to ``str``.
+
+        May be called with ``cast_to=None`` to return a full-fledged ``UUID``.
```
malefice, line 94: The autodoc system can now automatically create samples. If you want to learn more about how things are done under the hood, you may refer to this [page](https://faker.readthedocs.io/en/master/writing-docs.html). On that note, can you please add samples? Basically just change this line to this verbatim:

```
May be called with ``cast_to=None`` to return a full-fledged ``UUID``.

:sample:
:sample: cast_to=None
```

rodrigondec: That's brilliant! Didn't know it was even possible to do this

fcurella:
Thank you! ✨

PR #1166, opened 2020-04-27
Use repr for sample stringification

malefice:
Since #1164 was merged, the `uuid4` provider method now has an option to return a version 4 `UUID` object. The sample output in the  [docs](https://faker.readthedocs.io/en/master/providers/faker.providers.misc.html#faker.providers.misc.Provider.uuid4) does not look good/accurate though.

I do not know what had gotten into me as I attempted to reinvent the wheel to mimic Python interactive shell output, but I should have just used the builtin `repr` function for sample stringification. This PR is to correct that mistake and to properly mimic console output of non-primitives. This also fixes a very minor typo.

### DIFF
--- a/docs/writing-docs.rst
+++ b/docs/writing-docs.rst
@@ -2,7 +2,7 @@ Writing Documentation
 =====================
 
 Everything under :doc:`Standard Providers <providers>` and :doc:`Localized Providers <locales>`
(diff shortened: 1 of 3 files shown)


fcurella:
No worries! Thank you so much for fixing it!

Issue #1162, opened 2020-04-22
Shall we pytestify our tests?


malefice:
Since `Faker` now has its own `pytest` plugin,  it is kind of weird if our test suite itself does not take advantage of pytest-style tests and our own plugin. The test suite currently seems like it has an identity crisis because of all the `unittest` and `pytest` stuff mixed together.  One of the things I eventually like to see is the complete switch to a pure pytest-style test suite.

I did not push for this back then because as good as `pytest` is, there were just not enough good reasons to fight the inertia. The test suite works fine after all, but now, at least we have additional reasons, and I have an idea to simplify our test suite handling provider methods and to make it easier for contributors to write those tests.

We can define a ``faker`` fixture under `tests/providers/conftest.py` that extends the plugin's default behavior. This fixture will only be available for tests under `tests/providers`, and the additional behavior will involve introducing the availability of a class-scoped `Faker` instance for what i call "locale test classes". The details are best discussed via sample code:

```python
class TestEnUs:
   # Notice we did not declare anything here

    def test_something_1(self, faker):
        # but faker will be a class-scoped instance with locale='en_US'
        pass

    def test_something_2(self, faker):
        # and is shared here
        pass

    def test_something_3(self, faker):
        # and here as well
        pass

    def test_something_4(self, faker):
        # and seeded for each test inside this class
        pass


class TestEnPh:

    def test_something_else(self, faker):
        # but a new class-scoped instance will be created here with locale='en_PH'
        pass

class TestFilPh(TestEnPh):
    # and there will be a test similar to the EnPh test here
    # but with a fixture returning a 'fil_PH' class-scoped instance 
    pass


class TestNotALocale:

    def test_yet_another_thing(self, faker):
        # and since this is not a locale class, it defaults to the plugin's default behavior
        pass


def test_outside_of_class(faker):
    # this is not inside a locale class, so it also defaults to the plugin's default behavior
    pass
``` 

I already have some dirty test code that works as described, but I would like to hear your thoughts on this issue first.

fcurella:
Thank you @malefice !

The structure you outline makes a lot of sense to me :)

I would love to have the tests use `pytest`, but it seems a pretty tedious job to convert them all. 

I personally don't have the bandwidth to do it, but I would gladly review and merge a PR if someone else is willing to write it.

malefice:
[...]

PR #1167, opened 2020-04-28
Test Suite Housekeeping (Part 3) 

malefice:
**Note: I estimate that only 1/3 of the transition is complete.**

### What does this change

Since `Faker` also now ships its own `pytest` plugin, this PR aims to (fully or partially) transition the test suite from using `unittest` style test to `pytest` style tests as well as to start using `Faker`'s own fixtures in the test suite. This can also serve as a good opportunity to review the quality of individual tests, and improvements will be made whenever possible.


malefice:
@fcurella here is a first draft of the pytest transition we discussed in #1162. I would like to hear your thoughts and any suggestion you may have before I proceed with the rest.

tests/providers/conftest.py
```diff
@@ -0,0 +1,41 @@
...
+        return faker
+    _class_locale_faker.seed_instance(DEFAULT_SEED)
+    return _class_locale_faker
+
+
+@pytest.fixture(scope='class', autouse=True)
+def num_samples(request):
+    try:
+        num = int(request.cls.num_samples)
+    except:
```
fcurella, line 39: I would use `except AttributeError`, and let it raise `ValueError` if we pass something that's not castable to an `int`

malefice:
@fcurella I think travis hanged since the last merge to master. Anyway, looks like I might be rebasing my PR a fair amount prior to getting this merged if I do not split this now, so I will probably wrap this part up soon and continue with a new PR.

tests/providers/conftest.py
malefice: Fixed as part of this [commit](https://github.com/joke2k/faker/pull/1167/commits/44f9a3a7975b29247e54e0210b585535d9189866)

malefice:
@fcurella I am done for this portion which (I estimate) accounts for 1/3 of the transition. The tests for other providers are a bit tedious/repetitive/lacking, so I am going to think about how to tackle them, and I also do not want to introduce too much `pytest` magic into the test suite. Anyway, for now, can you please review my work?

fcurella:
@malefice It looks great! Impressive work!

PR #1168, opened 2020-04-28
More Spanish names added for person es_ES provider.

mondeja:
### What does this changes
Add more names to person es_ES provider.


fcurella:
Thank you @mondeja !

Do you have a statistical source for the new dataset? Where did you get the names from?

mondeja:
> Do you have a statistical source for the new dataset?
### Male names
- Number of previous male names: 100
- Number of new male names: 490
- Number of male names added: 390

### Female names
- Number of previous female names: 99
- Number of new female names: 502
- Number of female names added: 403

____________________________________

> Where did you get the names from?

From [here](https://www.behindthename.com/names/usage/spanish) filtering by `Spanish` found in `.listusage` spans texts.

mondeja:
I know that in your docs says:
> We only accept new data if it's coming from statistical sources, such as census or government institutions.

But in Spain there are so many languages, and the official INE dataset show names that are in `ES` but are not `es` (people that lives in Spain but doesn't have Spanish names). You can consult the official data [here](https://www.ine.es/tnombres/formGeneralresult.do?vista=1) searching by `Total`. Keep in mind that they even don't provide accents in names, is a crazy dataset.
- Are `Mohamed` (67th more common name in Spain) or `Gheorge` (325th) a `es_ES` name?
- `IVAN` name even exists? Not, the correct name is `Iván`, with accent.

I can submit the official dataset, but it is completely useless.

fcurella:
Thanks @mondeja 

> Are Mohamed (67th more common name in Spain) or Gheorge (325th) a es_ES name? We can include them?

This is exactly the reason why we want data from a statistical source. It's so that _we_ don't have to make the decisions about what to include or not in a provider.

I understand that the official dataset from the government is close to unusable. Is there any other source publicly available?

mondeja:
No, there are no more official sources.

fcurella:
The dataset doesn't have to be 'official', but at least collected in some systematic way.

Really all we care is that we can delegate responsibility to whoever collected the data. In some rare occasion we did had to combine data from multiple datasets and we have included all the sources that they came from.

> Is the localization of names intended by faker a geographic problem, a historic-idiomatic problem or both?

IMHO, that depends on the locale. It's whatever people using (ie: living in) `es_ES` think it's more appropriate, as they are the ones who have all the context and know best what's appropriate. The only requirement is that those names come from somewhere else than inside someone's head :)

mondeja:
I understand. As Spanish I can guarantee that the new names are valid `es_ES` names, not are inventions. Perhaps some names like `Mohamed` or `Gheorge` are actually more common than `Desiderio` or `Demetrio`, but I think that the dataset included in this pull is more accurate (for example `Joaquin` is a invented name, the correct name is `Joaquín`) and offers more valuability in terms of randomization because are more names.

So what are the problems? Any requested changes?

fcurella:
You can be the source for dataset, just use your email (or a URL) and name as source.

fcurella:
Thank you! ✨

PR #1169, opened 2020-05-01
BarcodeProvider prefix feature

yu-ichiro:
### What does this changes

Add more flexibility to BarcodeProvider so that it can handle localized barcodes.
 & Added US, Canada, Japan local EAN prefixes

### What was wrong

It only had the option to control the first digit to be 0 or not.

### How this fixes it

introduced a new parameter ``prefixes`` and _ean now chooses one of it to be the the first digits of the EAN code

Actually I wrote the code before I submitted an issue.. if its mandatory, I can create one.
First I wanted a Japanese-localized EAN barcode (or JAN code as its called here), but the BarcodeProvider was only capable of generating random or north-american oriented code only (as 
 I think upc codes are north american regional?)

I tried to follow the code patterns and syntax but if there's any suggestion, please inform me.


yu-ichiro:
ping @fcurella (sorry if it bothers you)
I really appreciate if you could take a look at my code

faker/providers/barcode/ja_JP/__init__.py
```diff
@@ -0,0 +1,15 @@
+from .. import Provider as BarCodeProvider
+
+
+class Provider(BarCodeProvider):
```
malefice, line 4: Please rename all instances of `BarCodeProvider` to  `BarcodeProvider`. Personally, I have always encountered the word "barcode" as one word, never as "bar code". It also means fewer presses of the SHIFT key for future contributors which sounds marginal, but I personally welcome it.

faker/providers/barcode/en_CA/__init__.py
```diff
@@ -0,0 +1,15 @@
+from itertools import product
+
+# Canada uses UPC too
+from ..en_US import Provider as BarCodeProvider
+
+
+class Provider(BarCodeProvider):
+    # Source of GS1 country codes: https://gs1.org/standards/id-keys/company-prefix
+    local_prefixes = (
```
malefice, line 9: I am not so sure about storing/specifying prefixes using tuples/lists of tuples/lists. For storing `local_prefixes`, it might be okay to avoid extra conversion steps, but at least for the user API, I think you will agree with this.

```python
# Typing this is quite cumbersome, and there can be too many parentheses
fake.ean13(prefixes=((4, 5), (4, 9)))

# This feels more intuitive and natural
fake.ean13(prefixes=('45', '49'))
```

 What do you think @fcurella?

tests/providers/test_barcode.py
```diff
@@ -185,3 +245,182 @@ def test_upc_e2a2e(self):
...
+            ean13_digits = [int(digit) for digit in ean13]
+            assert (sum(ean13_digits) + 2 * sum(ean13_digits[1::2])) % 10 == 0
+
+            assert any(
+                all(a == b for a, b in zip(ean13_digits, prefix))
+                for prefix in FrCAProvider.local_prefixes
+            )
+
+
+class TestJaCA(unittest.TestCase):
```
malefice, line 370: This should be `TestJaJP`.

faker/providers/barcode/ja_JP/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+    # Source of GS1 country codes: https://gs1.org/standards/id-keys/company-prefix
+    local_prefixes = (4, 5), (4, 9)
+
+    def jan(self, length=13):
```
malefice, line 8: Can you please write docstrings with samples for the methods specific to the `ja_JP` provider? They are essentially just aliases and call `localized_ean*` under the hood, so something noting that fact will suffice.

yu-ichiro:
Thanks for the review, @malefice!
I'll look into it.

faker/providers/barcode/ja_JP/__init__.py
yu-ichiro: got it. I agree that being more natural. I guess I weren't thinking when I wrote it 🤦

faker/providers/barcode/en_CA/__init__.py
yu-ichiro: I was wondering what will be the best way to pass it too. I just weren't pretty sure if string splitting and mapping int was the best way to get the prefix. But definitely strings are better on user API.

Also to make the API accept the last example, I can see try doing it in the current implementation  will be a pain.

faker/providers/barcode/ja_JP/__init__.py
yu-ichiro: Fixed this on the last commit! (Did I get it right?)

malefice: Please add summary lines to the docstrings. As for samples, only `jan()` requires an explicit sample section since it is the only one that accepts arguments. The methods `jan8()` and `jan13()` do not require one, and a sample section will automatically be generated for them. Just remove the `:return:` stuff in those two methods. You can read more about samples [here](https://faker.readthedocs.io/en/master/writing-docs.html).

Anyway, docstring should roughly follow this form.

```
def method(arg1, arg2):
    """Describe what method does in imperative mood.

   Notes and other info about method should be placed here.

    :sample: arg1='val1', arg2='val2'
    :sample: arg1=2, arg2=12
    """
   pass
```

faker/providers/barcode/en_CA/__init__.py
fcurella: I'm ok with just accepting lists/tuples, as long as the argument is plural (ie: `prefixes`)

yu-ichiro: Say, if I put map(int) around the chosen prefix, will that be better?

```
        if prefixes:
            prefix = self.random_element(prefixes)
            code[:len(prefix)] = map(int, prefix)
```

yu-ichiro: I was also thinking about the last example @malefice showed (`prefixes='45'`), but it might have the potential to make the meaning of the code ambiguous. (`prefixes=((4, 5),)` or `prefixes=((4,), (5,))`)
I mean, I'm sure 99% of the people will understand it as the prior one, but I feel a little uncomfortable.

By adopting the map(int) wrapping above, `prefixes=['45']` or `prefixes={'45'}` will work, so that might work as the more user friendly way of designating a single prefix?

faker/providers/barcode/ja_JP/__init__.py
yu-ichiro: I got to generate the actual sphinx doc and looked around.
I fixed some references and added summery lines and sample lines, and I just went and added `:sample: ` to jan8() and jan13() too  (I thought it looks nicer on the doc).

Thank you for guiding me out!

faker/providers/barcode/en_CA/__init__.py
fcurella: Would there be any reason to prevent someone from passing a prefix with a leading zero?

What would happen if a user calls `prefixes=('01', '02')`? Is that a use-case we want to prevent?

yu-ichiro: According to Honeywell(https://support.honeywellaidc.com/s/article/Can-EAN13-symbols-start-with-a-leading-zero), EAN-13 with leading zero is a legitimate code but identical to UPC-A, and there are some barcode readers that treat those codes as UPC-A and omit the first zero when outputting results.

So if someone is generating actual barcode from faker generated codes and is reading with that kind of reader, then it could be a problem. But I don't think it's faker's responsibility to exclude codes with leading zero from valid codes. I think noting the fact that there are those readers and if the user is going to generate the actual barcode, they should know the implications behind it should be enough.

I'll put a little more detail in the docstring and guide to upc code docstring if necessary.

yu-ichiro: @fcurella does this suffice?

fcurella:
Hi @yu-ichiro,

we refactored our test suite to use our new pytest plugin. Do you mind update your new tests? Sorry about the merge conflict!

yu-ichiro:
@fcurella No problem! I'll take a look into it.

yu-ichiro:
@fcurella fixed merge conflict :)

fcurella:
Looks great! Thank you!

PR #1171, opened 2020-05-09
Add provider for German day_of_week and month_name

eumiro:
### What does this changes

Add `day_of_week` and `month_name` to date_time/de_DE

### What was wrong

It was missing

### How this fixes it

Add the seven days of the week and twelve months of a year in standard German.


fcurella:
Thank you! Could you add some tests for the new providers?

eumiro:
Done. I copied the pl_PL test because it's actually the same code.

fcurella:
Thank you!

PR #1174, opened 2020-05-12
Synchronize cs_CZ and sk_SK address provider and split postcodes

eumiro:
Both address providers should now behave similarly.
The common Czechoslovak postcode system was split in 1993 and both countries CZ and SK kept their parts: CZ starts with 1 to 7, SK with 8 to 0. The standard `postcode()` method may be used to format it.
The `building_number` has been changed to a meaningful 1 to 3 digits number not starting with a zero.
Added more `street_suffixes`.

### DIFF
--- a/faker/providers/address/cs_CZ/__init__.py
+++ b/faker/providers/address/cs_CZ/__init__.py
@@ -11,8 +11,8 @@ class Provider(AddressProvider):
 
     building_number_formats = ('%', '%#', '%##')
 
-    street_suffixes_long = ('náměstí', )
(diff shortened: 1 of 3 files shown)


eumiro:
Here is the reference information on postal codes in CZ/SK:
https://en.wikipedia.org/wiki/Postal_codes_in_the_Czech_Republic
https://en.wikipedia.org/wiki/Postal_codes_in_Slovakia

eumiro:
@fcurella rebased and repushed, fails at coverall only

fcurella:
Thank you!

PR #1176, opened 2020-05-12
Add date_time providers for cs_CZ, de_AT, es_ES, it_IT, sk_SK, tr_TR

eumiro:
Similar to previous PR for `de_DE` https://github.com/joke2k/faker/pull/1171 along with tests, which are now inviting to switch to pytest structures with fixtures and parametrizing, just like proposed in https://github.com/joke2k/faker/issues/1162 and started in https://github.com/joke2k/faker/pull/1167


fcurella:
Thank you! ✨

PR #1178, opened 2020-05-13
[BREAKING] Fix CLI only giving 10 possible outputs

andrewdotn:
When running the cli command `faker name`, I’ve been surprised to see very
little variation in the outputs. For me, the output is always one of these
10 names:

    Brian Foster
    Chris Curtis
    James Taylor
    Joshua Wood
    Natalie Pope
    Norma Fisher
    Ryan Gallagher
    Samantha Washington
    Theresa Brown
    Victor Martinez

The line of code that causes this seems to be this:

    seeds = random.sample(range(arguments.repeat*10), arguments.repeat)

This is sampling `repeat` values from the list of integers
[0, 10 × repeat), which is [0, 9) when `repeat` is at its default value
of 1. When there are only 10 possible seeds, there are only 10 possible
outputs.

By seeding individual instances with `random.random()` output like
`0.49892926139211313`, running `faker <anything>` hundreds or thousands of
times will result in hundreds or thousands of different outputs, instead of
only 10.

This will break the reproducibility of `faker --seed` between different
releases of faker; if that’s a big problem, the default `seeds` could at
least be set to a list of `None` if no `--seed` argument is passed on the
command line.

### DIFF
--- a/faker/cli.py
+++ b/faker/cli.py
@@ -236,7 +236,7 @@ examples:
             logging.basicConfig(level=logging.CRITICAL)
 
         random.seed(arguments.seed)
-        seeds = random.sample(range(arguments.repeat*10), arguments.repeat)
+        seeds = [random.random() for _ in range(arguments.repeat)]
 
         for i in range(arguments.repeat):
 
@@ -264,3 +264,6 @@ def execute_from_command_line(argv=None):
 
     command = Command(argv)
     command.execute()
+
+if __name__ == '__main__':
+    execute_from_command_line()
--- a/faker/cli.py
+++ b/faker/cli.py
@@ -265,5 +265,6 @@ def execute_from_command_line(argv=None):
     command = Command(argv)
     command.execute()
 
+
 if __name__ == '__main__':
     execute_from_command_line()


faker/cli.py
```diff
@@ -264,3 +264,6 @@ def execute_from_command_line(argv=None):
 
     command = Command(argv)
     command.execute()
+
```
fcurella, line 267: could you add another newline here?

fcurella:
@andrewdotn I'm ok with breaking compatibility, as long as we upgrade the MAJOR version. Could you add `[BREAKING] ` to your commit subject, so I don't miss it?

andrewdotn:
@fcurella I’ve updated the PR title. If the next major version might not be for a while, would you accept some conditionally-backwards-compatible code?

```diff
diff --git a/faker/cli.py b/faker/cli.py
index 9b687737054b..93b433665860 100644
--- a/faker/cli.py
+++ b/faker/cli.py
@@ -236,7 +236,13 @@ examples:
             logging.basicConfig(level=logging.CRITICAL)
 
         random.seed(arguments.seed)
-        seeds = random.sample(range(arguments.repeat*10), arguments.repeat)
+        if int(VERSION.split('.')[0]) <= 4:
+            if arguments.seed is None:
+                seeds = [None] * arguments.repeat
+            else:
+                seeds = random.sample(range(arguments.repeat*10), arguments.repeat)
+        else:
+            seeds = [random.random() for _ in range(arguments.repeat)]
 
         for i in range(arguments.repeat):
 
```

andrewdotn:
@fcurella Could this be merged in some day?

A reminder of what bug this PR fixes: when running `faker name` or `faker email` or `faker postalcode` or `faker random_number` or many other commands, there are *only 10 possibilities for what `faker` will print.*

I applied my own patch to my local copy ages ago, so this doesn’t really affect me, but other users may be surprised and unhappy to learn that `faker random_number` can *only* print 1, 62, 582, 4969, 6068, 9709, 97227, 993908, 7056020, or 82304355; and nothing else.

fcurella:
Sorry, this must've fell between the cracks. Thank you for your patience!

andrewdotn:
Thanks!

PR #1181, opened 2020-05-15
bill provider for fa_IR locale

abtinmo:
### What does this changes

adds bill provider for fa_IR locale, it provides:
 bill_id, payment_id, provider, bill_name, and bill_full methods


fcurella:
Hi @abtinmo 

Thank you for taking the time to write this provider!

Unfortunately, I don't think the need this kind of data is common enough to be included in `Faker`, but if you could package it as an external provider, I'll be happy to include it among the [Community Providers](https://faker.readthedocs.io/en/master/communityproviders.html)

PR #1186, opened 2020-05-22
add some necessary items in gitignore

zarinpy:
### What does this changes

Brief summary of the changes
add some necessary items
### What was wrong
gitignore  was  not complete 

### How this fixes it

Fixes #...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,8 @@ cover/
 .idea
 .vscode/*
 .projectile
+.spyderproject
+.spyproject
 .ropeproject
 .DS_Store
 .venv
--- a/.gitignore
+++ b/.gitignore
@@ -31,10 +31,15 @@ pip-log.txt
 # Unit test / coverage reports
 .coverage
 .tox
+.nox
+.cache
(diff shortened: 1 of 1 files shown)


zarinpy:
@svenstaro @stephane

svenstaro:
Well what am I supposed to do with this. I'm not a maintainer.

.gitignore
```diff
@@ -1,7 +1,7 @@
 __pycache__/
 .mypy_cache/
 *.py[cod]
-
+MANIFEST
```
fcurella, line 4: I'm not sure about excluding `MANIFEST`. Where would this file be coming from?

zarinpy: A `MANIFEST.in` file can be added in a project to define the list of files to include in the distribution built by the sdist command.

When sdist is run, it will look for the `MANIFEST.in` file and interpret it to generate the **MANIFEST** file that contains the list of files that will be included in the package.
[some sugessions](https://perlmaven.com/dont-keep-generated-files-in-version-control) to exclude these files to be in repository

PR #1189, opened 2020-05-29
Adding .uk email providers and TLDs.

craiga:
Adds some British free email providers and TLDs.

Data taken from https://github.com/fzaninotto/Faker/blob/master/src/Faker/Provider/en_GB/Internet.php


fcurella:
> Data taken from ...

Would you mind add the source to the provider? Just copy and paste that line as a comment.

fcurella:
Thank you! ✨

PR #1193, opened 2020-06-02
Added generator for language names

### LINKED ISSUE #1179: New feature: language_name
Hi. I want to add a new method that would return random language name (e.g. English, Spanish, Dutch). I'm not sure where I should place this method. Initially I wanted to add this to BaseProvider near the language_code method, but this provider has no locale support.
Any thoughts?

ikhomutov:
Fixes #1179


faker/providers/person/__init__.py
```diff
@@ -10,6 +10,41 @@ class Provider(BaseProvider):
 
     last_names = ['Doe']
 
+    language_names = [
```
fcurella, line 13: Would add the source for this list?

ikhomutov: Done

fcurella:
Thank you! ✨

PR #1197, opened 2020-06-12
Add language names for person es_ES provider.

mondeja:
### What does this changes
Add Spanish language names to person `es_ES` provider.


fcurella:
Thank you! ✨

PR #1199, opened 2020-06-15
Add es_ES currency provider.

mondeja:
### What does this changes

Add Spanish currency names to `currency` provider.


tests/providers/test_currency.py
```diff
@@ -88,3 +88,25 @@ def test_currency_name(self, faker, num_samples):
...
+
+    @classmethod
+    def setup_class(cls):
+        from faker.providers.currency.es_ES import Provider as EsEsCurrencyProvider
+        cls.provider = EsEsCurrencyProvider
+        cls.currencies = cls.provider.currencies
+        cls.currency_codes, cls.currency_names = tuple(zip(*cls.currencies))
+
+    def test_currency(self, faker, num_samples):
+        for _ in range(num_samples):
```
fcurella, line 105: I don't think we need to check for `isinstance` 100 times.

fcurella:
Thank you! ✨

PR #1204, opened 2020-06-19
Add SWIFT code provider methods

malefice:
### What does this change

This will add support for generating SWIFT codes which are used in banking and financial institutions. Also, I took the liberty of creating the bank provider for the PH locales.


malefice:
@fcurella before I write the tests and fix the CI errors, do you have any suggestions and ideas for improvement?

fcurella:
I'm not sure I feel comfortable making up bank codes. My worry is that sometime in the future they might get used in the real world, and we'll have a conflict. I'd rather just fall back to the 'international' output (with the warning)

malefice:
Just so that we are in the same page, by bank codes, you are referring to the PH locales' `bban` and `iban` methods, correct? If yes, then would you prefer that to raise a `NotImplementedError` outright, or some combination of that plus an optional argument to force the fallback behavior like `forced=True`? How about the SWIFT codes?

fcurella:
>  by bank codes, you are referring to the PH locales' bban and iban methods, correct?

That's correct.

I thought about just raising en exception, but would prevent people from generating international ibans. They would have to instantiate a new provider with a different locale that implements the method. I'd rather fallback and let users know with a warning.

malefice:
What exactly is the "international" output to be expected in this case? Is it just the implementation of `en_GB` which is the default locale for the bank provider? I am asking because the default bank provider implemented the `bban` method such that it relies on `bban_format`, and the `iban` method depends on the value of `country_code` and on the output of `bban`.

I can leave the `bban_format` untouched for the PH locale bank providers, so it will default to the `bban_format` of `en_GB`, but the `country_code` will be an issue since it is required by the new SWIFT methods. Short of (1) declaring different `country_code` variables for bban/iban and swift or (2) reimplementing the bban/iban methods, I am not sure how to proceed, so I would like to hear your thoughts on this one.

fcurella:
After re-reading the code in PR, I realize I misunderstood how the country code in the SWIFT implementation would work, so please disregard my last comment. Sorry!

Back to your original question: If a local does not have a `bban` (or `swift`, or any other methods), we should fall back to a different locale. We always fall back to `en_US` (because of legacy reasons), but since `en_US` does not have `bban`, we'll have fall back to `en_GB`.

I'm a little unsure about the warning. We don't usually do that, but we did have quite a few issues in past from users confused about this behavior, so I think the fallback needs more visibility. I'm unsure about warning vs `logging.warning`. I tend to think about `logging` as more configurable for our users, so I'd favor that.

malefice:
OK I can definitely switch `warning` to `logging.warning`, but how do you want us to proceed in terms of implementing the fallback behavior?  Like I said, the swift methods and the iban method all depend on `country_code`. For locales that use both, there will be no problem, but for locales that use only swift or only bban/iban, "blatantly fake and invalid" bank codes will be generated because of the dependence on `country_code`. Using pseudocode to explain:

```
class DefaultBankProvider:
    bban_format = '????#############'
    country_code = 'GB'


# Uses SWIFT, but no BBAN and IBAN
class EnUsBankProvider(DefaultBankProvider):
    country_code = 'US'

    def bban(self):
        # Generates GB formatted bban which is desired fallback behavior

    def iban(self):
        # Uses GB formatted bban with a US country code, so blatantly fake and invalid

    def swift_methods(self):
        # Generates US formatted swift code which is correct


# Hypothetically uses BBAN/IBAN but not SWIFT
class HypotheticalBankProvider(DefaultBankProvider):
    bban_format = '??#######'    
    country_code = 'XX'

    def bban(self):
        # Generates XX formatted bban which is correct

    def iban(self):
        # Uses XX formatted bban with XX country code which is also correct

    def swift_methods(self):
        # Generates XX formatted swift code, so blatantly fake and invalid

```

malefice:
@fcurella have you made a decision regarding my concern?

fcurella:
@malefice I'm really undecided about what the best behavior would be, I can see both pros and cons of everything I can think of. Let's go with what you have here (+ the switch to `logging.warning`), and if necessary we can revisit later.

malefice:
@fcurella can you please review my work?

fcurella:
It's good to go! Thank you! ✨

PR #1209, opened 2020-07-08
Add safe domain names

### LINKED ISSUE #1182: Feature request: safe_domain_name()
There are methods for `safe_email()` but nothing that just returns the a  rfc6761 domain name for those cases where a user might want to provide a username generated off of other data (e.g. a faker-generated first/last name combo so it doesn't look weird that `George Smith` has an email address of `janedoe@example.com`).

I know I can get this with `fake.random_element(["example.com", "example.net", "example.org"])` but it would be nice to have this built in.

crd:
### What does this change

Refactors safe email domain generation and provides a new mechanism to generate random safe domains (one of 'example.com', 'example.org', 'example.net'). Refer to [RFC 2606](https://tools.ietf.org/html/rfc2606) and [RFC 6761](https://tools.ietf.org/html/rfc6761) for more information on these domains, intended for use in documentation.

Closes #1182

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -58,7 +58,7 @@ class _IPv4Constants:
 
 
 class Provider(BaseProvider):
-    safe_email_tlds = ('org', 'com', 'net')
+    safe_domain_names = ('example.org', 'example.com', 'example.net')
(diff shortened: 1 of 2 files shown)


fcurella:
Nice work! Thank you! ✨

PR #1218, opened 2020-07-21
💂 Fix max_value/positive pyfloat interaction

### LINKED ISSUE #1212: `pyfloat` with `positive=True` and `max_value` produces unexpected results.
* Faker version: v4.0.3
* OS: Mac OS

When using the `pyfloat` provider with both `positive=True` and `max_value` set, it sometimes produces negative numbers.

### Steps to reproduce

```python
from faker import Faker

Faker.seed(100)
faker = Faker()
print(faker.pyfloat(positive=True, max_value=100.0))
```

### Expected behavior

Should print some number in (0, 100).

### Actual behavior

Prints `-1039.2931776`.

coiax:
When pyfloat(max_value=100, positive=True) is called, it sometimes
returns negative numbers, which is unexpected.

To fix it, we just need to "cap" our internal minimum value to zero.

This does leave `pyfloat(max_value=100, min_value=-100, positive=True)`
as a call that secretly moves the min_value up to 0 though.

Fixes #1212.

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -77,7 +77,9 @@ class Provider(BaseProvider):
                 max_value += 1  # as the random_int will be generated up to max_value - 1
             if min_value is not None and min_value < 0:
                 min_value += 1  # as we then append digits after the left_number
-            left_number = self._safe_random_int(min_value, max_value)
+            left_number = self._safe_random_int(
+                min_value, max_value, positive,
+            )
         else:
             sign = '+' if positive else self.random_element(('+', '-'))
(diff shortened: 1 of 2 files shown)


fcurella:
Thank you!

Can we also raise an exception (maybe an AssertionError?) if `positive` is True and `min_value` is negative?

coiax:
A ValueError would make more sense, and would fit with the other exceptions that this function throws, if that's alright?

fcurella:
@coiax +1 to `ValueError`. Thanks!

fcurella:
Thank you! ✨

PR #1221, opened 2020-07-24
Fix person.ja_JP kana and roman characters

yameholo:
### What does this changes

- Added `first_kana_names` and `last_kana_names` and `first_romanized_names` and `last_romanized_names` elements based on [PHP Faker](https://github.com/fzaninotto/Faker/blob/eef05943cd5f9b00edf81daac1f1ef28b16942d6/src/Faker/Provider/ja_JP/Person.php).
- (IMO)Added the URL to get the element of the names, like `PHP Faker`.
- (IMO)The writing style in the tuple has been changed to be consistent.

### What was wrong

Japanese names are mainly written in kanji, and only one kana is used.
For example, if the name ( which uses kanji ) is "晃", then the kana will be "アキラ".
Likewise, kana are written in roman characters. (ex. if "アキラ" then "Akira".)
So we want to access elements with the same meaning when we access them in the same index.

ex
```
if first_names_male[i] == "晃":
    print(first_kana_names_male[i])  # => アキラ 
    print(first_romanized_names_male[i]) # => Akira
```

But it does not currently work as such.

### How this fixes it
The `PHP Faker` has a list of kanji and kana correctly mapped to it. I copied it.
Also, since the roman characters were not present in `PHP Faker`, I rewrote each list to correspond correctly.


fcurella:
Thank you! ✨

PR #1223, opened 2020-07-28
JSON and Fixed_Width Structures

johnbrandborg:
### What does this changes

Implements methods for building complex data structures with JSON, and flat data structures with Fixed Width, with generated data from the many provider methods.

Tuples are used for record declaration, so that parameters can be passed to the methods, and tune the values they generate.


johnbrandborg:
Wanting to get some feedback into if this is something that would be considered?

faker/builder.py
```diff
@@ -0,0 +1,107 @@
...
+            entry = []
+            for width, provider_name, *options in spec:
+                params = options[0] if options else {}
+                provider = getattr(self._factory, provider_name)
+                result = provider(**params) if callable(provider) else ""
+                field = "{0:<%s}" % width
+                entry.append(field.format(result)[:width])
+            yield ''.join(entry)
+
+    def json(self, spec: list, rows: int, indent=None) -> iter:
```
fcurella, line 65: I would move this to the [`misc` provider](https://github.com/joke2k/faker/blob/master/faker/providers/misc/__init__.py)

faker/builder.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __init__(self,
+                 locale: str = None,
+                 providers: str = None,
+                 generators: str = None,
+                 includes: str = None,
+                 **config):
+        self._factory = Factory.create(locale, providers, generators, includes, **config)
+
+    def csv(self, spec: list, rows: int, delimiter: str = "|", header: bool = False):
```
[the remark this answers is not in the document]
johnbrandborg: Oh fantastic!  I will remove this.

faker/builder.py
johnbrandborg: I have moved the code to the Misc provider, and refactored to integrate better.  If this works for you, I will look at the documentation and tests to ensure the tests pass.

johnbrandborg:
Looks like the coverage is failing due to a decrease, but I have 100% coverage on the code I introduced.

fcurella:
Thank you! Don't worry about a .007% drop :)

PR #1226, opened 2020-07-28
🦄 Add unique values support to Faker

### LINKED ISSUE #971: Return list of n unique companies (n_unique keyword for all providers?)

### LINKED ISSUE #1125: Enforce unique generated values
* Faker version: Latest
* OS: Any

I'm wondering if there is a way of forcing Faker to generate unique values always similar to UUID V4? I'd like to utilise Faker for a tokenisation service but this would required each token to be completely unique. i.e., if I generate a name, it has to be unique each and every time I generate it.

Do you think there's a way to accomplish this? Potentially in a similar way UUID V4 does?

### Steps to reproduce

N/A

### Expected behavior

N/A

### Actual behavior

N/A

coiax:
Adds support to Faker for guaranteed unique values.

A new `.unique` property has been added to the Proxy class, which
returns an instance of a `UniqueProxy`. Any attributes looked up on the
unique proxy goes to the common proxy. If the attribute is a function,
then it's wrapped with a wrapper that only allows unique values.

Obviously, it's not always possible to get a unique value, especially
with a small set of values. If the wrapper can't find a unique value
after a certain number of attempts (currently pegged at an arbitrary
1000), it throws a UniquenessSanityException.

Each "remembered" set of values is tied to the name of the function, and
the arguments passed to it, so, for example, different `.random_int()`
calls with different arguments will not share a "bucket".

Since it uses a set internally for remembering previous values, and a
dictionary for determining what bucket to use, it can't be used with any
arguments or return values that are unhashable.

---

Resolves #1125 
Resolves #971

- ~~Still needs full documentation added to sphinx, didn't want to spent lots of time documenting something that I didn't know if people were gonna accept or not. 100% happy to write some up if reaction is positive.~~ Done!
- Does the sanity value need to be changeable?
- Is this generally the right solution?

Had a lot of fun puzzling this one out, lemmie know if I'm on the right track!


faker/exceptions.py
```diff
@@ -0,0 +1,8 @@
+class BaseFakerException(Exception):
+    """The base exception for all Faker exceptions."""
```
coiax, line 2: Surprised this wasn't already a thing, some of the state address providers throw naked Exceptions already...

faker/proxy.py
```diff
@@ -235,3 +243,44 @@ def factories(self):
...
+
+    def clear(self):
+        self._seen = {}
+
+    def __getattr__(self, name: str):
+        obj = getattr(self._proxy, name)
+        if callable(obj):
+            return self._wrap(name, obj)
+        else:
+            return obj
```
coiax, line 262: I am not sure whether this should just throw an exception instead, given you shouldn't be looking up any attribute under `faker.unique` other than a function. Technically, currently, you could recurse. `faker.unique.unique.unique`...

faker/proxy.py
```diff
@@ -235,3 +243,44 @@ def factories(self):
 
     def items(self):
         return self._factory_map.items()
+
+
+class UniqueProxy:
+    def __init__(self, proxy):
+        self._proxy = proxy
+        self._seen = {}
+        self._sentinel = object()
```
coiax, line 252: I wanted a slightly more "fun" sentinel than a standard object, if only for a better `repr()`, but I didn't want to drag in another library just for that.

faker/proxy.py
```diff
@@ -235,3 +243,44 @@ def factories(self):
...
+        obj = getattr(self._proxy, name)
+        if callable(obj):
+            return self._wrap(name, obj)
+        else:
+            return obj
+
+    def _wrap(self, name, function):
+        @functools.wraps(function)
+        def wrapper(*args, **kwargs):
+            key = (name, *args, tuple(sorted(kwargs.items())))
```
coiax, line 267: Sort the keyword arguments, because `random_int(min=50, max=60)` should be equivalent to `random_int(max=60, min=50)`, although no clue why the signature would be different.

tests/test_unique.py
```diff
@@ -0,0 +1,53 @@
...
+from faker.exceptions import UniquenessSanityException
+
+class TestUniquenessClass:
+    def test_uniqueness(self):
+        fake = Faker("en_US")
+
+        names = set()
+        # There are (at time of writing 690) first names in the
+        # US identity provider. Birthday paradox puts the chances of
+        # no duplicates in 250 selections as low enough to be impossible
```
coiax, line 14: Wolfram Alpha [puts the probability](https://www.wolframalpha.com/input/?i=birthday+problem+calculator&assumption=%7B%22F%22%2C+%22BirthdayProblem%22%2C+%22pbds%22%7D+-%3E%22690%22&assumption=%7B%22F%22%2C+%22BirthdayProblem%22%2C+%22n%22%7D+-%3E%22250%22&assumption=%22FSelect%22+-%3E+%7B%7B%22BirthdayProblem%22%7D%7D) at 1 - 3.151 * 10^-23.

faker/exceptions.py
```diff
@@ -0,0 +1,8 @@
...
+
+
+class UniquenessSanityException(BaseFakerException):
```
fcurella, line 5: I think we can just call this `UniquenessException`

faker/proxy.py
fcurella: Let's throw an exception

faker/proxy.py
```diff
@@ -235,3 +244,44 @@ def factories(self):
...
+            # With use of a sentinel value rather than None, we leave
+            # None open as a valid return value.
+            retval = self._sentinel
+
+            for i in range(_UNIQUE_ATTEMPTS):
+                if retval not in generated:
+                    break
+                retval = function(*args, **kwargs)
+            else:
+                raise UniquenessSanityException
```
fcurella, line 281: I would add a message to the exception, to let the user know what happened (eg: "Got duplicated values after X iterations.")

coiax:
@fcurella Should I get working on the sphinx documentation for this PR, are you happy with it in its current form?

fcurella:
Thanks @coiax , looks good to me! You can start working on the docs :)

README.rst
```diff
@@ -358,6 +358,37 @@ By default all generators share the same instance of ``random.Random``, which
...
+
+Calling ``fake.unique.clear()`` clears the already seen values.
+Note, to avoid infinite loops, after a number of attempts to find a unique
+value, Faker will throw a ``UniquenessException``. Beware of the `birthday
+paradox <https://en.wikipedia.org/wiki/Birthday_problem>`_, collisions
+are more likely than you'd think.
+
+
+.. code:: python
+   from faker import Faker
```
fcurella, line 382:
```suggestion
.. code:: python

   from faker import Faker
```

docs/fakerclass.rst
```diff
@@ -287,3 +287,66 @@ a cheatsheet for new ``Faker`` in multiple locale mode.
...
+``Faker`` proxy.
+
+Accessing provider methods through this attribute guarantees that
+the returned values are unique for the lifetime of the ``Faker`` instance.
+
+
+.. code:: python
+
+   import faker
+   fake = faker.Faker()
```
fcurella, line 305:
```suggestion
   import faker
   
   fake = faker.Faker()
```

docs/fakerclass.rst
```diff
@@ -287,3 +287,66 @@ a cheatsheet for new ``Faker`` in multiple locale mode.
...
+allow previous values generated to be returned again.
+
+Different argument signatures for provider methods do not share
+a uniqueness pool.
+
+
+.. code:: python
+
+   import faker
+   fake = faker.Faker()
```
fcurella, line 320:
```suggestion
   import faker
   
   fake = faker.Faker()
```

docs/fakerclass.rst
```diff
@@ -287,3 +287,66 @@ a cheatsheet for new ``Faker`` in multiple locale mode.
...
+and the ``.unique`` attribute has been used, it's possible that no
+suitable unique value will be found after a certain number of iterations.
+
+To avoid infinite loops, at that point, a ``UniquenessException`` will
+be raised.
+
+
+.. code:: python
+   import faker
+   fake = faker.Faker()
```
fcurella, line 337:
```suggestion
.. code:: python

   import faker
   
   fake = faker.Faker()
```

docs/fakerclass.rst
```diff
@@ -287,3 +287,66 @@ a cheatsheet for new ``Faker`` in multiple locale mode.
...
+
+
+As a final caveat, only hashable arguments and return values can be used
+with the ``.unique`` attribute, as it is backed internally by a set for
+fast membership testing.
+
+
+.. code:: python
+   import faker
+   fake = faker.Faker()
```
fcurella, line 350:
```suggestion
.. code:: python

   import faker
   
   fake = faker.Faker()
```

faker/contrib/pytest/plugin.py
```diff
@@ -33,5 +33,6 @@ def faker(request):
     if 'faker_seed' in request.fixturenames:
         seed = request.getfixturevalue('faker_seed')
     fake.seed_instance(seed=seed)
+    fake.unique.clear()
```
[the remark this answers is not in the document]
fcurella: Take a look a the tests for the fixtures in https://github.com/joke2k/faker/blob/master/tests/pytest.

@malefice Do you have time to help out with this?

fcurella:
@coiax I've made some light changes re the formatting of the code examples

faker/contrib/pytest/plugin.py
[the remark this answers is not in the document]
coiax: I used https://docs.pytest.org/en/latest/writing_plugins.html#testing-plugins to test this addition, just pushed the commit for it.

coiax:
@fcurella Is there anything else that needs to be done for this PR now?

README.rst
```diff
@@ -358,6 +358,36 @@ By default all generators share the same instance of ``random.Random``, which
...
+
+   from faker import Faker
+   fake = Faker()
+   names = [fake.unique.first_name() for i in range(500)]
+   assert len(set(names)) == len(names)
+
+Calling ``fake.unique.clear()`` clears the already seen values.
+Note, to avoid infinite loops, after a number of attempts to find a unique
+value, Faker will throw a ``UniquenessException``. Beware of the [birthday
+paradox](https://en.wikipedia.org/wiki/Birthday_problem), collisions
```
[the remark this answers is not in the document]
malefice: I can relate.

tests/pytest/test_unique_clear.py
```diff
@@ -0,0 +1,18 @@
+def test_unique_clears(testdir):
+    """Successive uses of the `faker` pytest fixture have the
+    generated unique values cleared between functions."""
+
+    testdir.makepyfile(
```
malefice, line 5: I think it is better if you use a similar test you used in `TestUniquenessClass`, and I am thinking something like this:


```python
def test_fully_exhaust_unique_booleans(faker, num_samples):
    _dummy = [faker.boolean() for _ in range(num_samples)]
    faker.unique.boolean()
    faker.unique.boolean()
    with pytest.raises(UniquenessException):
        faker.unique.boolean()
    _dummy = [faker.boolean() for _ in range(num_samples)]

def test_do_not_exhaust_booleans(faker):
    faker.unique.boolean()

def test_fully_exhaust_booleans_again(faker, num_samples):
    _dummy = [faker.boolean() for _ in range(num_samples)]
    faker.unique.boolean()
    faker.unique.boolean()
    with pytest.raises(UniquenessException):
        faker.unique.boolean()
    _dummy = [faker.boolean() for _ in range(num_samples)]
```

As far as I know, `pytest` should run the tests in that order, but if the order is inconsistent, then you can probably use `makeconftest` and declare `pytest_collection_modifyitems` within to enforce the proper test execution order.

tests/test_unique.py
```diff
@@ -0,0 +1,62 @@
+import pytest
+
+from faker import Faker
+from faker.exceptions import UniquenessException
+
+
+class TestUniquenessClass:
```
malefice, line 7: I would add some tests to show that using the `unique` proxy will not interfere with the default behavior like the `_dummy = [faker.boolean() for _ in range(num_samples)]` thing I mentioned in another comment.

tests/pytest/test_unique_clear.py
coiax: I'm confused though, pytest suggest in their docs to use the [testdir](https://docs.pytest.org/en/stable/reference.html#pytest.pytester.Testdir) fixture to test plugins, so I'm not sure I understand the difference between what I've wrriten, and what you're suggesting.

malefice: I am not saying not to use the `testdir` fixture. What I am saying is that the tests you made inside `testdir.makepyfile` can be made better, because you only have these tests inside:

```python
def test_generate_two_unique_booleans(faker):
    faker.unique.boolean()
    faker.unique.boolean()
def test_another_boolean(faker):
    faker.unique.boolean()
```

coiax: Right, I'm following, apologies. :sweat_smile:

coiax:
Apologies on taking so long on this last review step.

fcurella:
@malefice Could you give one last review if you have the chance?

tests/pytest/test_unique_clear.py
```diff
@@ -4,15 +4,34 @@ def test_unique_clears(testdir):
 
     testdir.makepyfile(
         """
-        def test_generate_two_unique_booleans(faker):
+        import pytest
+        from faker.exceptions import UniquenessException
+
+        NUM_SAMPLES = 100
```
malefice, line 10: We already have a `num_samples` fixture for this purpose. You could have just copy pasted the code snippet I wrote. Did you encounter issues?

tests/pytest/test_unique_clear.py
```diff
@@ -4,15 +4,34 @@ def test_unique_clears(testdir):
...
+
+        def test_fully_exhause_unique_booleans(faker):
```
malefice, line 12: Typo: exhause should be exhaust.

tests/pytest/test_unique_clear.py
```diff
@@ -4,15 +4,34 @@ def test_unique_clears(testdir):
...
+            _dummy = [faker.boolean() for _ in range(NUM_SAMPLES)]
+
             faker.unique.boolean()
             faker.unique.boolean()
+            with pytest.raises(UniquenessException):
+                faker.unique.boolean()
+            _dummy = [faker.boolean() for _ in range(NUM_SAMPLES)]
 
-        def test_another_boolean(faker):
+        def test_do_not_exhause_booleans(faker):
```
malefice, line 21: Typo: exhause should be exhaust.

tests/pytest/test_unique_clear.py
```diff
@@ -4,15 +4,34 @@ def test_unique_clears(testdir):
...
+            faker.unique.boolean()
+
+        def test_fully_exhause_unique_booleans_again(faker):
```
malefice, line 24: Typo: exhause should be exhaust.

tests/pytest/test_unique_clear.py
coiax: the `num_samples` fixture isn't part of the fixtures that the faker pytest plugin exposes, so I just defined it there to stay within the spirit of the code sample.

malefice: While the `num_samples` fixture is not part of the pytest plugin `Faker` ships with, there is one defined for the test suite. I however misremembered, and the fixture is only available under `tests.providers` as defined [here](https://github.com/joke2k/faker/blob/565d3aa016cb986031ad00746c4d3066e9743710/tests/providers/conftest.py#L36). This should be OK for now.

malefice:
@fcurella I have nothing to add, and this looks OK now.

fcurella (approved):

fcurella:
Thank you so much for working on this! We've been needing this feature for a very long time! ✨

PR #1207, opened 2020-06-26
Type error on calling Faker from within an internal pytest fixture

fmeyer:
For reasons I don't understand yet, pytest is calling this private fixture
during the setup phase of the tests. Regardless the reason, the code contains a
bug since it expects a `string` but the `_session_faker` constructs a `List`
before calling the constructor

* Faker version: 0.7.4
* OS: `Ubuntu 14.04.6 LTS. travis`
* Python: 3.6.5

```
request = <SubRequest '_session_faker' for <TestCaseFunction test_get_access_method_for_event>>
    @pytest.fixture(scope='session', autouse=True)
    def _session_faker(request):
        """Fixture that stores the session level ``Faker`` instance.

        This fixture is internal and is only meant for use within the project.
        Third parties should instead use the ``faker`` fixture for their tests.
        """
        if 'faker_session_locale' in request.fixturenames:
            locale = request.getfixturevalue('faker_session_locale')
        else:
            locale = [DEFAULT_LOCALE]
>       return Faker(locale=locale)
../../../virtualenv/python3.6.5/lib/python3.6/site-packages/faker/contrib/pytest/plugin.py:20:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
cls = <class 'faker.factory.Factory'>, locale = ['en_US'], providers = None
generator = None, includes = [], config = {}
    @classmethod
    def create(cls, locale=None, providers=None, generator=None, includes=None, **config):
        if includes is None:
            includes = []

        # fix locale to package name
>       locale = locale.replace('-', '_') if locale else DEFAULT_LOCALE
E       AttributeError: 'list' object has no attribute 'replace'
../../../virtualenv/python3.6.5/lib/python3.6/site-packages/faker/factory.py:22: AttributeError
```

The stacktrace that leads to this execution is the following

```
-> "__main__", mod_spec)
  /opt/python/3.6.5/lib/python3.6/runpy.py(85)_run_code()
-> exec(code, run_globals)
  /home/travis/virtualenv/python3.6.5/lib/python3.6/site-packages/pytest.py(102)<module>()
-> raise SystemExit(pytest.main())
  /home/travis/virtualenv/python3.6.5/lib/python3.6/site-packages/_pytest/config/__init__.py(82)main()
-> return config.hook.pytest_cmdline_main(config=config)
[...]

Issue #1177, opened 2020-05-12
Issue with latest faker version: AttributeError: 'list' object has no attribute 'replace'


mouradmourafiq:
* Faker version: 4.1.0
* OS: ubuntu

Tests that used to pass are now failing, the error:

```
>   lambda: ihook(item=item, **kwds),
    when=when,
            )

/usr/local/lib/python3.7/site-packages/flaky/flaky_pytest_plugin.py:273:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/usr/local/lib/python3.7/site-packages/faker/contrib/pytest/plugin.py:20: in _session_faker
    return Faker(locale=locale)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <class 'faker.factory.Factory'>, locale = ['en_US'], providers = None, generator = None, includes = [], config = {}

    @classmethod
    def create(cls, locale=None, providers=None, generator=None, includes=None, **config):
        if includes is None:
            includes = []

        # fix locale to package name
>       locale = locale.replace('-', '_') if locale else DEFAULT_LOCALE
E       AttributeError: 'list' object has no attribute 'replace'

/usr/local/lib/python3.7/site-packages/faker/factory.py:22: AttributeError
``` 

### Steps to reproduce
I just pinned faker to: `faker==4.0.3` and all tests are passing.

### Expected behavior

It should work, similarly to the previous release.

malefice:
A new `pytest` plugin has been introduced in `4.1.0`, and based on your error log, something in your test suite and/or `flaky` might be conflicting with our `_session_faker` fixture ([source](https://github.com/joke2k/faker/blob/f930d83a85e49321437799bf19f00251143dcf47/faker/contrib/pytest/plugin.py#L9)).

[...]

PR #1229, opened 2020-07-31
Add README compatibility notice

malefice:
### What does this change

Add a version compatibility notice (as well as suggestions) in the README

### What was wrong

Ever since we added the pytest plugin in `4.0.0`, there have been reports of test suites failing like #1177 and #1207. In the former, OP has confirmed that the issue was caused by old Python versions and packages, and in the latter, it also seems to be the case, since OP stated version `0.7.4`.

### How this fixes it

I made the notice more prominent in the README and the first page of the docs, and I added more details and recommended actions.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,8 +19,18 @@ Faker is heavily inspired by `PHP Faker`_, `Perl Faker`_, and by `Ruby Faker`_.
 
 ----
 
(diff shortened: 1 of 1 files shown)


fcurella:
Thank you! I"m sure this will help a lot of people 👍

PR #1230, opened 2020-07-31
Handle isort 5.X upgrade

malefice:
### What does this changes

This will fix the recent isort failures in travis CI as well as fix outstanding isort issues in other files. 

### What was wrong

The isort package recently had a [major update](https://timothycrosley.github.io/isort/CHANGELOG/#500-penny-july-4-2020). Our CI isort tasks started failing around the 5.0 release, and basically, the isort 5.X now [requires](https://github.com/timothycrosley/isort/blob/64ab238666298d5ff203082a6213087fa4a39858/isort/main.py#L619) files/paths to be specified. This did not use to be the case in the 4.3.21 where the current working directory will [implicitly be used](https://github.com/timothycrosley/isort/blob/7c29dd9d55161704cfc45998c6f5c2c43d39264b/isort/main.py#L334).

### DIFF
--- a/faker/providers/job/bs_BA/__init__.py
+++ b/faker/providers/job/bs_BA/__init__.py
@@ -1,6 +1,5 @@
 from .. import Provider as BaseProvider
 
-
 """
 Data is provided by the official list of professions from
 National institue for statistics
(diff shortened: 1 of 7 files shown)


fcurella:
Thank you!

PR #1236, opened 2020-08-10
Cleanup italian names for person it_IT provider.

rsalmaso:
Update:
* Alphons -> Alfonso
* Atenulf -> Atenolfo
* Sophia -> Sofia

Removed: Hugo, T


fcurella:
Thank you @rsalmaso :)

Unfortunately, those edits don't match the source. Since the source is wikipedia, you could edit the source to match what you think the names should be.

Or even better, you could replace the data with something coming from a statistically compiled source, like a dataset from [ISTAT](https://www.istat.it/)

fcurella:
Closing for inactivity. Feel free to reopen.

Issue #1233, opened 2020-08-06
Issues in fi_FI company_business_id method


malefice:
* Faker version: 4.1.1
* OS: Ubuntu 18.04.4 LTS

### Steps to reproduce

Run `python -c "from faker import Faker; f = Faker('fi_FI'); Faker.seed(0); assert all(len(f.company_business_id()) == 9 for _ in range(100))"` and you will get an `AssertionError`.

### Expected behavior

There should be no `AssertionError` since we expect the ids to be exactly 9 characters.

### Actual behavior

Unfortunately, an `AssertionError` will be raised.

### Analysis

As part of the test suite housekeeping I have been doing, I found out that the `fi_FI` company provider's `company_business_id` method produces invalid id's. There were other buggy provider methods (due to obvious typos) that I fixed in my branch, but this one is rather significant that I think should separately be tracked.

 The company ids should be 7 digits followed by a dash followed by a check digit, so there should be 9 characters, but the problem lies [here](https://github.com/joke2k/faker/blob/15c0f6e345f4c7b179232b92ecc699d89d611b5e/faker/providers/company/fi_FI/__init__.py#L33). When that line results in a 10 (remainder is 1), the id gets an extra digit. I checked [other](https://github.com/vkomulai/finnish-business-ids/blob/4486d71060cb31e21a553dcbd4293f35563aca8a/src/finnish-business-ids.ts#L58) [implementations](https://gist.github.com/lauriii/3cb3b32ad86c271ebc681012855ce529) that validate Finnish company id's, and they say that such id's are truly invalid.

To fix this, there has to be a check that will retry if the remainder ends up being 1. Also, the `fi_FI` provider method `company_vat` is also affected since it depends on `company_business_id`. It would be nice if someone who uses the `fi_FI` locale can weigh in to verify my findings as well.

fcurella:
Thank you for digging into this @malefice ! I've added the 'help wanted' label in the hope of bigger visibility.

@kivipe Since you wrote the original implementation, would you mind weighing in?

kivipe:
Nice find and excellent bug report. I found a description about business id calculation and it says "if remainder is 1, id is not allowed."

fcurella:
Closed by #1237

PR #1237, opened 2020-08-10
Don't return too long company ids. Fix for #1233

kivipe:
### What does this changes

This is a bugfix for #1233 . 

### What was wrong

Finnish company id calculation did not handle invalid remainder 1 in checksum calculation, resulting two digit checksums.

### How this fixes it

Company id is generated until a valid one is found


fcurella:
Thank you! ✨

PR #1241, opened 2020-08-11
Improve proxy initialization as list type error.

mondeja:
### What does this changes

When `Faker` class is initialized as list with values that are not strings, a better exception is raised. 

### Actual behaviour

```python
>>> from faker import Faker
>>> Faker([1, 2])
  File "/.../faker/faker/proxy.py", line 33, in __init__
    assert all(isinstance(local_code, str) for local_code in locale)
AssertionError
```

### Behaviour after this pull

```python
>>> from faker import Faker
>>> Faker([1, 2])
  File "/.../faker/faker/proxy.py", line 36, in __init__
    raise TypeError('The locale "%s" must be a string.' % str(code))
TypeError: The locale "1" must be a string.
```

### DIFF
--- a/faker/proxy.py
+++ b/faker/proxy.py
@@ -30,9 +30,10 @@ class Faker:
         # This guarantees a FIFO ordering of elements in `locales` based on the final
         # locale string while discarding duplicates after processing
         elif isinstance(locale, (list, tuple, set)):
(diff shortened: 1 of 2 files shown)


fcurella:
Thank you!

PR #1243, opened 2020-08-12
Parser argument support

johnbrandborg:
### What does this changes

Adds the ability to send arguments through the Generator Parser, to further refine the generated values.  Anything using the parser, like DSV would benefit hugely by this.   Methods have been added for Set, Get, Del, and the __config is used to store the arguments.

Using the format {{ formatter:group }}, the arguments can be used.

Example:

generator.set_arguments('red_rgb', {"hue":"red", "color_format":"rgb"})
generator.parse("{{ color:red_rgb }}")

I would like to align JSON and Fixed Width to DSV, and implement the use of the parser.  This would allow JSON to be created using a Dict, rather than tuples and overall be much simpler, with the benefits DSV has by using the parser.


faker/generator.py
```diff
@@ -99,10 +98,24 @@ def parse(self, text):
...
         return _re_token.sub(self.__format_token, text)
 
     def __format_token(self, matches):
         formatter = list(matches.groups())
-        formatter[1] = str(self.format(formatter[1]))
-        return ''.join(formatter)
+
+        if formatter[1]:
+            try:
+                params = eval(formatter[1].replace(':', 'dict(') + ')')
```
fcurella, line 114: I'd rather tokenize the input, parse it, and return a `dict` instead of  just wrap it in a `dict(..)` sting and pass it to eval.

We could use something like [`sly`](https://github.com/dabeaz/sly).

I've actually done something very similar recently for a different project, but in Faker's case I'm wary of the performance impact of a more complex parsing, especially when methods are called in loops. I'd definitely do some profiling.

johnbrandborg: JSON maybe simpler than Sly, consider we just need a dictionary and it would handle all the data structures need to pass to the providers.  We could pre-parse the JSON to make basic syntax allowed.

fcurella: I agree that JSON would be simpler to implement, but doing string replacement feels hacky, and I fear that a) it would be more fragile (ie: not handling all cases) and b) in order to handle all cases, it could end up be just as complex as sly, if not more.

Consider having to support the following cases:
```
{{ pyint:min_value=1, min_value=10 }}
{{pyint:min_value=1, min_value=10}}
{{ pyint:min_value=1,   min_value=10 }}
{{ pyint:min_value = 1, min_value = 10 }}
{{ pyint : min_value=1, min_value=10 }}

{{ color:hue=[100, 200] }}
{{ color:hue=(100, 200) }}
```

I think it wold helpful to see what the pre-parsing code you have in mind actually looks like

johnbrandborg: You are right about the performance.  It took a big hit so I've put in format caching if parameters are used, and the results were great.  The parser seems to cover all the edge cases listed.   It's quite bullet proof, and even mixed/bad formatting gets corrected, so much so I had to remove my old test for bad formatting.   More tests are needed to ensure __format_json_parser retains stability though and I haven't done it yet.

faker/generator.py
```diff
@@ -128,8 +128,12 @@ def __format_token(self, matches):
...
-        string = re.sub(r'\((.*)\)', r'[\g<1>]', string)           # Safe Tuple
-        string = re.sub(r'^([^{].*[^}])$', r'{\g<1>}', string)     # Wrapped
+        string = re.sub(r'([-\w]+)\s*[:=]', r'"\g<1>":', string)            # Safe Keys
+        string = re.sub(r':\s*([^0-9\[\{\("][-\w]+)', r':"\g<1>"', string)  # Safe Values
+
+        string = re.sub(r'\((.*)\)', r'[\g<1>]', string)                    # Safe Lists
+        string = re.sub(r'(\[|\,)\s*([\w-]+)\s*(\,|\])', r'\g<1>"\g<2>"\g<3>', string)
+        string = re.sub(r'(\,)\s*([\w-]+)\s*(\,)', r'\g<1>"\g<2>"\g<3>', string)
+
+        string = re.sub(r'^([^{].*[^}])$', r'{\g<1>}', string)              # Wrapped
```
johnbrandborg, line 138: So what is missing still is dict handling, but I wouldn't think many providers would take a dict.  Would like to spend a little more time refining this and the tests even more.

fcurella:
I agree, this approach looks much cleaner. Thank you so much!

PR #1244, opened 2020-08-13
Add es_ES autonomous communities (Spanish regions).

mondeja:
### What does this changes

Add Spanish regions (locally known as "autonomous communities") to `es_ES` address provider using government spreadsheet as source.


fcurella:
Thank you!

PR #1245, opened 2020-08-13
Upgrade pytest to version 6

mondeja:
### What does this changes

The unique change affected for 6 version is the mandatory use of keyword arguments in fixtures decorators, proposed in [this issue](https://github.com/pytest-dev/pytest/issues/1682).

### What was wrong

Running the test suite with pytest 6 raises next warning error:

```
Passing arguments to pytest.fixture() as positional arguments is deprecated - pass them as a keyword argument instead
```

### How this fixes it

Use `scope='class'` in `tests/providers/test_barcode.py` fixture.


fcurella:
Thank you!

PR #1247, opened 2020-08-19
JSON and Fixed Width argument group and parser support.

johnbrandborg:
### What does this changes

The new Argument Groups is a much cleaner way to handle parsing arguments to provider methods when building data structures like DSV, JSON and Fixed Width.  It also allows for the JSON provider to now accept a Dictionary that utilizes the Argument Groups, which is a much simpler way to build complex data structures.

```Python
fake.set_arguments('top_half', {'min_value': 50,'max_value': 100})

fake.json(data_columns={'Name': 'name', 'Score': 'pyint:top_half'})
fake.fixed_width(data_columns=[(20, 'name'), (3, 'pyint:top_half')])
```

If string tokens are used then complex Strings can be created by using pystr_format under the hood.

```Python
fake.set_arguments('red_rgb', {'hue': 'red', 'color_format': 'rgb'})
fake.json(data_columns={'Address': 'address', 'Feature': '### - {{ color:red_rgb }}'})
```

All previous support for lists with tuple records is still in tacked and will not be broken.  More tests have been added to ensure absolutely safety for anyone that begins to use JSON with list format or dict format, and Fixed Width generation.


fcurella:
Nice work! Thank you! ✨

PR #1248, opened 2020-08-25
Test Suite Housekeeping (Part 4)

malefice:
### What does this change

Just another round of pytestifying. There were quite a lot of changes to master since the last batch, and I think a force push was applied to the master branch since the 4.1.1 release, so I am submitting this now before things become difficult to rebase.

Related to #1162


fcurella:
Thank you so much! ✨

PR #1249, opened 2020-08-25
🥼 Add coverage to uncovered ru_RU and fr_CH tests

coiax:
To stop random coveralls failures, the a number of parametrized tests
have been added to explore more code paths for the Russian address
provider and the Swiss "social security number" provider.


faker/providers/address/ru_RU/__init__.py
```diff
@@ -384,9 +384,10 @@ def street_name(self):
                 if street in self.street_titles_irregular_masc.keys():
                     result = self.street_titles_irregular_masc[street]
                 else:
-                    result = stem + 'ый'
                     if stem.endswith('ск') or stem.endswith('цк'):
                         result = stem + 'ий'
+                    else:
+                        result = stem + 'ый'
```
coiax, line 390: This change in the file should not actually change the outcome, but I felt was slightly cleaner.

faker/providers/ssn/fr_CH/__init__.py
```diff
@@ -42,5 +42,5 @@ def _checksum(digits):
                 return 5
             return remainder
 
-        vat_id = self.bothify('########')
+        vat_id = self.numerify('########')
```
coiax, line 45: `.bothify()` just calls `.numerify()` under the hood anyway.

fcurella:
@coiax Merging https://github.com/joke2k/faker/pull/1248 created some conflicts. Could you rebase and fix them?

coiax:
I didn't include the test I wrote for "ru_RU" region, since the new test for "ru_RU" region is actually better than the one I wrote, given it checks for the suffix/prefix in the right location.

Chances of a given path out of 4 not being picked 100 times in a row is quite small. The street name function is however very complex, so I kept my work on it.

fcurella:
Thank you!

PR #1250, opened 2020-08-27
Update zh_CN ssn provider support gender

mapoor:
### What does this changes

Update zh_CN ssn provider, it will support gender option.

### What was wrong

The zh_CN ssn not support gender option when generate ssn code.

### How this fixes it

ssn function add gender parameter with a default value.


tests/providers/test_ssn.py
```diff
@@ -25,6 +25,7 @@
 from faker.providers.ssn.pl_PL import calculate_month as pl_calculate_mouth
 from faker.providers.ssn.pl_PL import checksum as pl_checksum
 from faker.providers.ssn.pt_BR import checksum as pt_checksum
+from faker.providers.ssn.zh_CN import Provider as cn_Provider
```
fcurella, line 28: This import is not needed. `Faker('zh_CN')` on [L973](https://github.com/joke2k/faker/pull/1250/files#diff-8cab176395fd82f6b9a24fb3a2fd38d7R973) will already import for you.

fcurella:
Thank you!

PR #1252, opened 2020-08-28
Fixed typo in German job provider

datadominik:
### What does this change

Fixed typo in line 27 of German job provider `providers/job/de_DE/__init__.py`

### What was wrong

Minor typo in job "Forstwirst" -> should be "Forstwirt"

### How this fixes it

Changed typo to correct spelling


fcurella:
Thank you!

PR #1254, opened 2020-08-28
Add a random image provider

n1ngu:
Python Image Library is used under the hood to
generate a background and drawing a polygon,
while forwarding the color generation to the
existing color provider.

So as not to force a dependency on Pillow for all
installs, this provider will not work in
absence of this library.

A polygon shape is chosen because its the easiest
drawing method to interface with random coordinates,
as opposed to ellipses or alike that constrain input
parameters, while still providing easily distinguishable
random images, as opposed to random noise bitmaps.


n1ngu:
This addresses #603

n1ngu:
I'm still uncertain about how the provider should behave in the absence on PIL. Right now it will simply raise an ImportError on call, but maybe we could wrap it with a NotImplementedError explaining the soft dependency and state the extra dependency in setup.py? Maybe we should get inspired by whatever Django does?

Also, I put the provider in the color package for convinience so as to abuse the color provider, but maybe you want to sort the code elsewhere.

At the end I haven't bothered about filesystem-backed fileobjects because Pillow will nevertheless build the raw image in-memory so there's little we can do here. The provider users can move the bytearray to disk as soon as they want and we avoid the burden of making up sensible extensions for the variety of image formats supported by Pillow. Should the commit message speak about this details? What will the merging strategy be?

Also, I think I need some pointers on how to test the provider. What do you test in a random generator besides "it runs"? Would that fix the test coverage decrease? (I must say I am unused to this coverall.io thing)

n1ngu:
For the record, here is a sample of the generated images
image
image
image
You get the idea.

n1ngu:
@malefice @fcurella whenever you feel like it please review this PR. I think it's rather straightforward but it's growing merge conflicts :cry: 

Cheers

malefice:
> I'm still uncertain about how the provider should behave in the absence on PIL. Right now it will simply raise an ImportError on call, but maybe we could wrap it with a NotImplementedError explaining the soft dependency and state the extra dependency in setup.py? Maybe we should get inspired by whatever Django does?

Personally I am a fan of raising custom exceptions if none of the standard exceptions are a good fit. If custom exceptions are not desirable, then maybe just lazily raise the `ImportError` with a message stating the missing dependency. The most important thing for me is that only the methods that depend on `Pillow` will fail when invoked and not the whole module/stack.

> Also, I put the provider in the color package for convinience so as to abuse the color provider, but maybe you want to sort the code elsewhere.

You should just put it inside the `misc` provider alongside the other methods that produce binary data. It might be worthwhile to create a new `binary` provider altogether, though that will be up to @fcurella, but I digress. If you need to access provider methods outside of the provider you are writing for, you can use use the `generator` object to get the reference, so change `self.color` to `self.generator.color`.

> Should the commit message speak about this details? What will the merging strategy be?

Write docstrings because there is some magic that converts those things into something like [this](https://faker.readthedocs.io/en/master/providers/faker.providers.misc.html#faker.providers.misc.Provider.tar). Please also read [this](https://faker.readthedocs.io/en/master/writing-docs.html
) for more details. As for merging strategy, @fcurella usually squashes the PR commits and performs a fast-forward merge.

> Also, I think I need some pointers on how to test the provider. What do you test in a random generator besides "it runs"? Would that fix the test coverage decrease? (I must say I am unused to this coverall.io thing)

Generate images, load them back in `Pillow`, and check if each image conforms to the parameters you specified like the `size` and `image_format`. As for testing the actual colors, the `image` method depends on the output of the `color` method, and since the latter already has its own tests, you can assume 100% reliability and safely replace it with a mock object and coerce the return value to some fixed color values. From there, test that the image only contains pixels that match those colors. You can take a look at some of the tests that use this technique like [this](https://github.com/joke2k/faker/blob/965da0fefc8567b732e5ce1920b247a6ba4b4a07/tests/providers/test_internet.py#L403).

n1ngu:
> You should just put it inside the `misc` provider alongside the other methods that produce binary data. It might be worthwhile to create a new `binary` provider altogether, though that will be up to @fcurella, but I digress. If you need to access provider methods outside of the provider you are writing for, you can use use the `generator` object to get the reference, so change `self.color` to `self.generator.color`.

While trying to do this I broke the documentation tests beacuse the hu_HU color provider is almost empty! Essentially it does not inherit the generic faker.providers.color.Provider class so it has no methods whatsoever. So far this hasn't been a problem for generating the docs because... it had no methods to be documented! But the misc.hu_HU provider could not be implemented relying on `self.generator.color`.

This is probably related to #490 and feels like a bug to me. It's a minor fix, but should this be fixed in a separate PR or commit?

n1ngu:
> The most important thing for me is that only the methods that depend on `Pillow` will fail when invoked and not the whole module/stack.

Obviously that is a MUST I always had in mind. Yet, would a separate tox testenv be desirable to ascertain this, @malefice ? Maybe testing it against all the python versions would be unnecessary then? Should this affect any CI scripts?

Regarding the unsuccessful check about the coverage decrease: it is caused by the fix of the hu_HU color provider, the image provider **is** tested. I hope you understand testing the hu_HU color provider is beyond my scope. It'd be sad this PR fell into oblivion because it does not "get the green light"

fcurella:
@n1ngu Some line of codes are literally covered randomly, and coverall.io is misconfigured to fail with a decrease as small as -0.001%.

So don't worry about the test coverage issue, as long as it doesn't decrease more than 1%.

malefice:
According to this [blame](https://github.com/joke2k/faker/blame/master/faker/providers/color/hu_HU/__init__.py), the `hu_HU` color provider has been written that way for years now. I apologize though for not noticing it when I wrote docstrings for that provider. Anyway, I think the fix to this should be a separate PR, partly to credit you (if you want to have at it) and partly for separation of concerns.

n1ngu:
> [...] the fix to this should be a separate PR, partly to credit you (if you want to have at it) and partly for separation of concerns.

I'm a big fan of the separation of concerns, I'll open a separate PR for the fix

n1ngu:
4 the record: I'm on vacations but I'll be able to finish this PR very soon. Regards.

n1ngu:
Maybe not that very soon, sorry

loscil06:
@n1ngu looking forward for this pr to be merged, looks really cool

mgaitan:
🙏🏽  merge this

fcurella:
Thank you!

Issue #1210, opened 2020-07-09
name_male() can not work properly when your locale is 'zh_CN'


vintager:
* Faker version:4.1.1
* OS:windows 10

Brief summary of the issue goes here.

### Steps to reproduce
```
f = Faker(locale = 'zh_CN')
f.name_male()
```
you will get both male name and female name.

after i checked the code in `/faker/providers/person/zh_CN/__init__.py`, i found should add this in the top of this file:
```
    formats_female = ('{{last_name}}{{first_name_female}}')
    formats_male = ('{{last_name}}{{first_name_male}}')
```

`/faker/providers/person/zh_TW/__init__.py` also need be fix

fcurella:
Thank you for the report, @vintager !

Would you mind submitting a PR with those changes?

datadominik:
> Thank you for the report, @vintager !
> 
> Would you mind submitting a PR with those changes?

@fcurella , I created the following PR for the zh_CN provider: https://github.com/joke2k/faker/pull/1255

PR #1255, opened 2020-08-29
Fixed zh_CN name provider to distinguish between male and female firs…

datadominik:
### What does this change

Fixes a problem mentioned in https://github.com/joke2k/faker/issues/1210, where the zh_CN provider is not distinguishing between male and female first names.

### What was wrong
No `formats` for male and female first names

### How this fixes it
Using `formats_male` and `formats_female`:

`formats_male = ["{{last_name}}{{first_name_male}}"]`
`formats_female = ["{{last_name}}{{first_name_female}}"]`
`formats = formats_male + formats_female`

### DIFF
--- a/faker/providers/person/zh_CN/__init__.py
+++ b/faker/providers/person/zh_CN/__init__.py
@@ -4,7 +4,9 @@ from .. import Provider as PersonProvider
 
 
 class Provider(PersonProvider):
-    formats = ["{{last_name}}{{first_name}}"]
+    formats_male = ["{{last_name}}{{first_name_male}}"]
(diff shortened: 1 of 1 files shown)


fcurella (changes requested):
Thank you @datadominik ! Could you add test cases for each sex?

datadominik:
> Thank you @datadominik ! Could you add test cases for each sex?

@fcurella , sure can try that. Just to be sure, you mean adding a testcase in `faker/tests/providers/test_person.py` correct?

fcurella:
@datadominik correct

PR #1256, opened 2020-09-01
Test Suite Housekeeping (Part 5)

malefice:
### What does this change

Just another round of pytestifying, and a minor fix for the `pt_BR` phone provider.

Related to #1162


fcurella:
Thank you!

Issue #1201, opened 2020-06-16
With python3.8 date_time_between_dates returns faker.utils.datetime_safe.datetime instead of datetime.datetime


Rafalkufel:
* Faker version: 4.1.0
* OS: Linux
* Python3.8

### Brief summary of the issue goes here.

`date_time_between_dates` return `faker.utils.datetime_safe.datetime` instead of `datetime.datetime`
Problem stem from solution of python bug: https://bugs.python.org/issue32417
Faker works fine with earlier version of python (eg 3.7)

### Steps to reproduce
```
from datetime import datetime, timedelta

from faker import Faker

fake = Faker()
Faker.seed(0)

fake_date = fake.date_time_between_dates(
    datetime_start=datetime.utcnow() - timedelta(days=360),
    datetime_end=datetime.utcnow(),
    tzinfo=None
)
    
print(type(fake_date))
```
### Expected behavior
output:
<class 'datetime.datetime'>

### Actual behavior
output:
<class 'faker.utils.datetime_safe.datetime'>

fcurella:
Hi! Thank you for reporting this.

My understanding from the Python bug is that returning a `datetime.datetime` was a bug in the first place, and returning a `datetime_safe.datetime` is what's considered expected behavior by Python.

Could be more specific about what's breaking for you?

Rafalkufel:
Sorry, maybe I didn't make myself clear.

Using `date_time_between_dates` provider I would expect that the fake object which returns is the same class as the real one (eg `datetime.datetime` when it comes to date). And that is the way it works with previous version of python. But surprinsingly, with python3.8 it returns object of different type than expected. In my case `pymysql` starts receiving `faker.utils.datetime_safe.datetime` object in place where `datetime.datetime` is expected.

I think that change in the way python3.8 datetime works should be reflected also in Faker to get real `datetime` from datetime provider.

samcrang:
We're also seeing a problem with this.

I'd expect the `faker.date*()` methods to return a regular Python `date`/`datetime` objects not the subtyped `date`/`datetime`s Faker uses internally. Since Python3.8 we're seeing the latter behaviour.

Regular `date`/`datetime` objects are returned when running on Python3.7 however it appears [the behaviour of arithmetic on subtyped `date`/`datetime` objects changed in Python3.8](https://github.com/python/cpython/commit/d9503c307a8b6a7b73f6344183602ffb014d3356).

On Python 3.7 we get this:

```
[...]

PR #1258, opened 2020-09-10
Remove datetime_safe

samcrang:
# What does this change

This removes the `datetime_safe` type. According to the comments this was only included because "Python's datetime strftime doesn't handle dates before 1900". This doesn't appear to be the case for the supported Pythons for Faker:

```
Python 2.7.17 (default, Dec 23 2019, 21:25:34)
[GCC 4.2.1 Compatible Apple LLVM 11.0.0 (clang-1100.0.33.16)] on darwin
>>> import datetime
>>> datetime.datetime(1800, 12, 11)
datetime.datetime(1800, 12, 11, 0, 0)
>>> datetime.datetime(1800, 12, 11).strftime('%Y-%m-%d')
ValueError: year=1800 is before 1900; the datetime strftime() methods require year >= 1900
```

```
Python 3.5.9 (default, Sep 10 2020, 22:25:21)
[GCC 4.2.1 Compatible Apple LLVM 11.0.0 (clang-1100.0.33.17)] on darwin
>>> import datetime
>>> datetime.datetime(1800, 12, 11)
datetime.datetime(1800, 12, 11, 0, 0)
>>> datetime.datetime(1800, 12, 11).strftime('%Y-%m-%d')
'1800-12-11'
```

### What was wrong

This causes issues on versions of Python >= 3.8. See issue #1201.

### How this fixes it

This removes the problematic `datetime` sub-typing entirely.


fcurella:
Thank you! ✨

bitdivision:
Shouldn't this have been a breaking change with a major version update? This will break any code importing that type.

PR #1259, opened 2020-09-11
Remove "sexy" from the list of bs words

Asimov4:
### What does this changes

Removes "sexy" from the list of bs words.

### What was wrong

The bs faker could generate borderline not safe for work associations like "enhance sexy experiences" or "enhance sexy relationships".

### How this fixes it

Removing "sexy" as a word prevents this from happening.


fcurella:
Thank you!

Issue #1262, opened 2020-09-14
Invalid Org Id in sv_SE Provider


StabbarN:
Faker version: latest master

The `ORG_ID_DIGIT_1` is incorrect at https://github.com/joke2k/faker/blob/deb9e01c8c2d47b33a5d409ea7dfdef221947890/faker/providers/ssn/sv_SE/__init__.py#L45

### Expected behavior

It should be as in django-localflavor https://github.com/django/django-localflavor/blob/5808835c49bb4330a75f3cf730f70a1d8e105bb8/localflavor/se/utils.py#L90
That is, `('2', '5', '7', '8', '9')`

PR #1263, opened 2020-09-14
Generate Valid Org Id in sv_SE provider

StabbarN:
Should pass test "valid_organisation" in local-flavor https://github.com/django/django-localflavor/blob/5808835c49bb4330a75f3cf730f70a1d8e105bb8/localflavor/se/utils.py#L90

### What does this changes
Generate a valid organization number.

### What was wrong
Org ID introduced in https://github.com/joke2k/faker/pull/1022 is invalid.

Fixes https://github.com/joke2k/faker/issues/1262.

### DIFF
--- a/faker/providers/ssn/sv_SE/__init__.py
+++ b/faker/providers/ssn/sv_SE/__init__.py
@@ -42,7 +42,7 @@ class Provider(SsnProvider):
 
         return pnr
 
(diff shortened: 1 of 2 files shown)


StabbarN:
Nevermind, the org number is too strict.

PR #1268, opened 2020-09-22
Use "Belarus" instead of "Vitryssland"

Majsvaffla:
### What does this changes

The name of the country Belarus in the Swedish address locale is changed from the older Swedish variant. The Swedish government (among other governments) has since some time been using "Belarus" to name the country.

- https://www.svt.se/nyheter/utrikes/regeringen-vitryssland-blir-belarus
- https://spraktidningen.se/artiklar/2020/04/vitryssland-eller-belarus

### What was wrong

Belarus was named as "Vitryssland" which is outdated.

### How this fixes it

Instead of "Vitryssland", "Belarus" is used as the Swedish name of the country Belarus.

### DIFF
--- a/faker/providers/address/sv_SE/__init__.py
+++ b/faker/providers/address/sv_SE/__init__.py
@@ -40,8 +40,8 @@ class Provider(AddressProvider):
         'Afghanistan', 'Albanien', 'Algeriet', 'Amerikanska Samoa', 'Andorra',
         'Angola', 'Anguilla', 'Antarktis', 'Antigua och Barbuda', 'Argentina',
         'Armenien', 'Aruba', 'Ascension', 'Australien', 'Azerbajdzjan',
-        'Bahamas', 'Bahrain', 'Bangladesh', 'Barbados', 'Belgien', 'Belize',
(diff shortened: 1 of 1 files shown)


fcurella:
Thank you! Could we keep the list in alphabetical order?

Majsvaffla:
> Thank you! Could we keep the list in alphabetical order?

I thought of that but chose to not reorder it because of the way it is formatted. I totally disagree with the impractical formatting of lists in this project. I'll obviously reorder it if you want me to but I thought I'd point that out anyway.

fcurella:
@Majsvaffla I agree it's inconvenient, but in our case it saves a LOT on the package size, especially on disk after installation. We really don't want to end up being 25% of your container!

If reordering is too cumbersome, I can do it upon merge. Just let me know :)

Majsvaffla:
> @Majsvaffla I agree it's inconvenient, but in our case it saves a LOT on the package size, especially on disk after installation. We really don't want to end up being 25% of your container!
> 
> If reordering is too cumbersome, I can do it upon merge. Just let me know :)

I get that. I pushed the changes just now. Turns out it didn't take that much effort ;)

fcurella:
Thank you so much! ✨

PR #1273, opened 2020-09-23
Improve VAT generation for IT provider

### LINKED ISSUE #1272: VAT for IT provider generates invalid values
* Faker version: `4.1.3`
* OS: Linux Debian

`company_vat` for `it_IT` providers randomly returns invalid VAT values.

### Expected behavior

As example, `company_vat()` generated the invalid value `IT00472451194`. This value is not valid because the last part of the code, `IT#######119#` is not a valid number for the algorithm that validates the Italian VAT (the [stdnum library](https://github.com/arthurdejong/python-stdnum), for example, correctly recognize this as an invalid VAT). As described in the ["Partita IVA" (VAT) wikipedia page](https://www.wikiwand.com/it/Partita_IVA), the valid values must be in the range `001-100` plus the values `120,121,888,999`.

The current code is:

```python
def company_vat(self):
        """
        Returns Italian VAT identification number (Partita IVA).
        """
        code = "0" + self.bothify('######') + str(self.generator.random.randrange(1, 121)).zfill(3)
        luhn_checksum = str(calculate_luhn(code))
        return 'IT{}{}'.format(code, luhn_checksum)
```

this code has some issues:

- The `randrange` is generating invalid values.
- The method is not generating all the possible VAT due to the missing `888` and `999` values. Also, the first digit of the VAT is allowed to be values other than 0 (currently it is fixed to 0).

LordGordonQ:
### What does this changes

Closes #1272 and improves VAT generation for IT provider allowing to generate all possible formally valid italian VAT.

This also adds a test case scenario for VAT generation with the IT provider.

### DIFF
--- a/tests/providers/test_company.py
+++ b/tests/providers/test_company.py
@@ -1,6 +1,9 @@
 import re
+import pytest
 
 from datetime import datetime
+
+from unittest.mock import patch
 
 from faker.providers.company.en_PH import Provider as EnPhCompanyProvider
 from faker.providers.company.fil_PH import Provider as FilPhCompanyProvider
@@ -332,3 +335,18 @@ class TestItIt:
         for _ in range(num_samples):
             company_vat = faker.company_vat()
             assert self.vat_regex.match(company_vat)
+
+    @pytest.mark.parametrize("value", (
+        101,
+        102,
+        103,
+        104,
+    ))
+    def test_company_vat_mock(self, faker, value):
+        # this test allows to get full code coverage for company_vat
+        title_patch = patch(
+            "faker.generator.random.randint",
+            autospec=True,
+            return_value=value,
+        )
+        assert self.vat_regex.match(faker.company_vat())
--- a/tests/providers/test_company.py
+++ b/tests/providers/test_company.py
@@ -350,8 +350,7 @@ class TestItIt:
         # this test allows to get full code coverage for company_vat fixing the internal state of the random generator
         fake = ItItCompanyProvider(generator=Generator())
 
-        with patch.object(fake, "random_int", return_value=value, autospec=True) as mock_randint:
+        with patch.object(fake, "random_int", return_value=value, autospec=True):
             company_vat = fake.company_vat()
-            mock_randint.assert_called()
             assert self.vat_regex.match(company_vat)
             assert company_vat[9:12] == expected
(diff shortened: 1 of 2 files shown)


LordGordonQ:
Going to fix the linting issue and improve code coverage for my new code.

tests/providers/test_company.py
```diff
@@ -322,3 +327,30 @@ def test_bs(self, faker, num_samples):
...
+    @pytest.mark.parametrize("value, expected", (
+        (100, "100"),
+        (101, "120"),
+        (102, "121"),
+        (103, "888"),
+        (104, "999"),
+    ))
+    def test_company_vat_mock(self, value, expected):
+        # this test allows to get full code coverage for company_vat fixing the internal state of the random generator
+        fake = ItItCompanyProvider(generator=Generator())
```
malefice, line 351: `Generator` and `Factory` can be considered legacy and for internal use only. Personally, I want to see them and providers redesigned to reduce the amount of magic (and abuse) going on under the hood, but it will be difficult for us to proceed if the test suite uses them everywhere.

On that note, there is no need to instantiate a `Generator` object. Remove the import, and use the `faker` pytest fixture instead like so:

```python
@pytest.mark.parametrize("value, expected", (
    # ...
))
def test_company_vat_special_cases(self, faker, value, expected):
    # ...
    provider = ItItCompanyProvider(generator=faker)
    # ...
```

tests/providers/test_company.py
```diff
@@ -322,3 +327,30 @@ def test_bs(self, faker, num_samples):
...
+            assert self.vat_regex.match(company_vat)
+
+    @pytest.mark.parametrize("value, expected", (
+        (100, "100"),
+        (101, "120"),
+        (102, "121"),
+        (103, "888"),
+        (104, "999"),
+    ))
+    def test_company_vat_mock(self, value, expected):
```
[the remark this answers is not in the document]
LordGordonQ: sure

tests/providers/test_company.py
LordGordonQ: In fact I was trying to using directly a `Faker` instance, but I had trouble to properly mock the internal random state.
I'll update the PR with your suggestions.

fcurella:
Thank you both!

PR #1274, opened 2020-09-25
Avoid sv_SE postal codes starting with zero

nip3o:
### What was wrong

Swedish postal codes are five digits and start with a non-zero digit (https://sv.wikipedia.org/wiki/Postnummer_i_Sverige). Currently invalid postal codes are sometimes generated since they start with a zero.

### How this fixes it

Changes postal code starting digit from any digit to 1..9.


nip3o:
Duplicate of #1267

PR #1279, opened 2020-10-01
Fix hu_HU color provider inheritance

n1ngu:
### What does this changes

Fix hu_HU color provider inheritance

### What was wrong

It was broken since it was first introduced
in d5ad2d04dd512852bd2903f12d991db6de2d21dc

### How this fixes it

The common ColorProvider must be inherited, no the abstract BaseProvider

Addresses #490 
Unblocks #1254


malefice:
@fcurella this looks OK.

fcurella (approved):

PR #1281, opened 2020-10-01
Added new style pt_PT automotive plates

gplgps:
### What does this changes

Adds the newest style of Portugal's automotive plates, which is being used since March 2020.

### What was wrong

Missing newest style of Portugal's automotive plates.

### How this fixes it

The new style can now be generated.


malefice:
@fcurella this looks OK.

PR #1282, opened 2020-10-01
Removed duplicate jobs from the pt_PT provider

gplgps:
### What does this changes

Removes duplicate jobs from the pt_PT jobs provider

### What was wrong

There was duplicate jobs within the jobs list.

### How this fixes it

Removes the duplicate jobs present within this list.


malefice:
The list of jobs looks like it has been generated by running something like this:

``` python
>>> from faker.providers.job.pt_PT import Provider
>>> jobs = []
>>> [jobs.append(job) for job in Provider.jobs if job not in jobs]
>>> jobs
```

And after some string manipulation and formatting, it matches the list of jobs in the PR's branch, so I guess this looks OK.

PR #1283, opened 2020-10-01
Implement color provider for ``sk_SK`` locale

pipozzz:
### What does this changes

This PR implements color provider for ``sk_SK`` locale


faker/providers/color/sk_SK/__init__.py
```diff
@@ -0,0 +1,11 @@
+from faker.providers import BaseProvider
+
+
+class Provider(BaseProvider):
```
malefice, line 4: You must inherit from the standard color provider, not the base provider; otherwise, there will be some unexpected behavior. Please refer to #1279.

pipozzz: Now, it should be OK.

malefice:
@fcurella this looks OK.

fcurella:
Can we add some tests?

pipozzz:
I added test there, btw something is wrong with appveyor, it's still in pending state.

fcurella:
@pipozzz unfortunately appveyor is VERY slow, so we'll have to patient. I'd love to port the windows tests to GitHub Actions, but I can't find a way to run the same matrix (specifically, python 3.6 on x32).

fcurella (approved):

PR #1292, opened 2020-10-02
Add province list and 2 new district for ne_NP address provider

iamsauravsharma:
### What does this changes

Add 2 new district which was formed by splitting 2 old district to 4 district and add province list of ne_NP. As of 20 September 2015 there are 77 districts.


fcurella (approved):

fcurella:
Thank you for keeping Faker up to date! :)

PR #1297, opened 2020-10-04
Remove abandoned Community Provider CloudProvider

crd:
CloudProvider, a Community Provider added in 2018, consisted of only a single release and the Home Page link from pypi no longer worked. Poking around the owner's github account shows that the repo has been archived.


fcurella:
Thank you!

PR #1298, opened 2020-10-07
Remove warnings when building documentation

Tenzer:
### What does this changes

This fixes any warnings emitted by Sphinx when the documentation site is built

### What was wrong

Lots of different things, mostly incorrect rST syntax. See the individual commits for details.
This resulted in 322 warnings as counted by Sphinx.

### How this fixes it

The invalid syntax has been corrected.

Note: This still leaves behind some messages about sample generation failing for a couple of methods plus a stream of "Numbers generated by this method are purely hypothetical." messages originating from https://github.com/joke2k/faker/blob/master/faker/providers/bank/en_PH/__init__.py

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,6 +28,7 @@ extensions = [
     'sphinx.ext.todo',
     'sphinx.ext.intersphinx',
     'faker.sphinx.autodoc',
+    'sphinx_rtd_theme',
 ]
 
 # Add any paths that contain templates here, relative to this directory.
@@ -98,7 +99,7 @@ pygments_style = 'sphinx'
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'default'
+html_theme = 'sphinx_rtd_theme'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
--- a/docs/fakerclass.rst
+++ b/docs/fakerclass.rst
@@ -21,7 +21,7 @@ new ``Faker.seed()`` points to ``Generator.seed()``, in new ``Faker``, invocatio
 from a ``Faker`` object instance has been disabled, and attempting to do so will raise a
 ``TypeError`` as shown below.
 
-.. code:: python
+.. code:: text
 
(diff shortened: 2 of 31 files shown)


Tenzer:
I was contemplating changing the theme configured in Sphinx from "default" to "sphinx_rtd_theme" which is the same used on the hosted docs site, but I couldn't find any obvious place to add the extra dependency on the "sphinx-rtd-theme" package which would be needed for that.
It would be helpful if the site generated locally would be as much alike what the end result is when it get deployed to the hosted site.

malefice:
> I was contemplating changing the theme configured in Sphinx from "default" to "sphinx_rtd_theme" which is the same used on the hosted docs site

@Tenzer I have been down this path before, and IIRC the local `sphinx_rtd_theme` and the one used in ReadTheDocs do have some minor presentation differences. I did not include this back then, because I preferred to see what the final output would be, so I opted to set up the RTD workflow instead. It is also easier to just include links to showcase what impact the proposed changes will have. I am not against your idea though, and I am just pointing out some of the pros and cons I encountered.

> but I couldn't find any obvious place to add the extra dependency on the "sphinx-rtd-theme" package which would be needed for that.

We can probably use `extra_requires` for this and define a set of `dev` dependencies, and then just have aspiring contributors do `pip install -e .[dev]` or something.

Tenzer:
> @Tenzer I have been down this path before, and IIRC the local `sphinx_rtd_theme` and the one used in ReadTheDocs do have some minor presentation differences.

The differences I see are very minor. This is what a page looks like on faker.readthedocs.io:

Screenshot 2020-10-09 at 09 12 28

compared to the Read the Docs theme locallly:

Screenshot 2020-10-09 at 09 16 53

and just for the comparison, this is what the default Sphinx theme looks like:

Screenshot 2020-10-09 at 09 15 31

I actually find that theme difficult to navigate, perhaps due to the site not being structured with that theme in mind.

> We can probably use `extra_requires` for this and define a set of `dev` dependencies, and then just have aspiring contributors do `pip install -e .[dev]` or something.

Sure, I've done that now. I have added a "docs" extra requirement which specifies Sphinx and the RtD theme.

PR #1303, opened 2020-10-11
Add Swiss bank provider locales

### LINKED ISSUE #1196: Fake IBAN for Switzerland should be of 21 characters in length
* Faker version: 4.1.0

Fake IBAN for Swiss locale should be 21 characters long, is 22 characters long now.

### Steps to reproduce

    >>> swiss_iban = Faker('de_CH').iban()
    >>> swiss_iban
    'GB80WQBB93456553336766'
    >>> len(swiss_iban)
    22

### Expected behavior

    >>> swiss_iban = Faker('de_CH').iban()
    >>> swiss_iban
    'CH4489144168326729379'
    >>> len(swiss_iban)
    21

### Actual behavior

    >>> len(swiss_iban)
    22

mondeja:
### What does this changes

Add Swiss `de_CH`, `fr_CH` and `it_CH` bank provider locales.

Closes #1196

### DIFF
--- a/faker/providers/bank/fr_CH/__init__.py
+++ b/faker/providers/bank/fr_CH/__init__.py
@@ -1,0 +1,1 @@
+from ..de_CH import Provider
--- a/faker/providers/bank/it_CH/__init__.py
+++ b/faker/providers/bank/it_CH/__init__.py
@@ -1,0 +1,1 @@
+from ..de_CH import Provider
--- a/tests/providers/test_bank.py
+++ b/tests/providers/test_bank.py
@@ -6,7 +6,10 @@ from faker.providers.bank import Provider as BankProvider
 from faker.providers.bank.en_GB import Provider as EnGbBankProvider
 from faker.providers.bank.en_PH import Provider as EnPhBankProvider
 from faker.providers.bank.es_ES import Provider as EsEsBankProvider
+from faker.providers.bank.de_CH import Provider as DeChBankProvider
 from faker.providers.bank.fi_FI import Provider as FiFiBankProvider
+from faker.providers.bank.fr_CH import Provider as FrChBankProvider
+from faker.providers.bank.it_CH import Provider as ItChBankProvider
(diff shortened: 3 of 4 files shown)


faker/providers/bank/fr_CH/__init__.py
```diff
@@ -0,0 +1 @@
+from ..de_CH import Provider  # noqa
```
malefice, line 1: By not creating a subclass, the docstrings will be the same, and that will be reflected in the docs. If there is no difference, then it is preferable to be explicit about it in the docstrings too so that the docs will also include that information. Please create a subclass similar to what has been done [here](https://github.com/joke2k/faker/blob/9e360fa90eb6bfe3fde4c320ac32f033cc803e00/faker/providers/bank/fil_PH/__init__.py#L4).

faker/providers/bank/it_CH/__init__.py
```diff
@@ -0,0 +1 @@
+from ..de_CH import Provider  # noqa
```
malefice, line 1: Same comment as the `fr_CH` localization.

tests/providers/test_bank.py
```diff
@@ -207,3 +210,48 @@ def test_iban(self, faker, num_samples):
...
+
+    def test_iban(self, faker, num_samples):
+        for _ in range(num_samples):
+            iban = faker.iban()
+            assert is_valid_iban(iban)
+            assert iban[:2] == DeChBankProvider.country_code
+            assert re.fullmatch(r"\d{19}", iban[2:])
+
+
+class TestFrCh:
```
malefice, line 230: Absolutely no need to change this, but I just want to let you know that you could have done something like [this](https://github.com/joke2k/faker/blob/9e360fa90eb6bfe3fde4c320ac32f033cc803e00/tests/providers/test_bank.py#L187) since the 3 locales have the same implementation anyway.

fcurella (approved):

fcurella:
Thank you!

PR #1304, opened 2020-10-11
Add more samples to automotive es_ES test to prevent random coverage decrease

mondeja:
### What does this changes

Seeds an automotive `es_ES` provider test to prevent a random coverage decrease.

### What was wrong

The method `license_plate` of `es_ES` automotive provider selects randomly between old and new Spanish license plates formats, and this is causing a random coverage decrease because one of the strategies are not been processed, like in [this job](https://coveralls.io/jobs/68951268/source_files/2836171701#L123).

### How this fixes it

Seeds the faker instance in test and checks each format separately.

### DIFF
--- a/tests/providers/test_automotive.py
+++ b/tests/providers/test_automotive.py
@@ -157,9 +157,15 @@ class TestEsEs:
             assert plate[:2] == "CA"
 
     def test_plate_format(self, faker):
+        faker.seed_instance(5600)
+
         plate = faker.license_plate()
         assert isinstance(plate, str)
-        assert self.new_format_pattern.match(plate) or self.old_format_pattern.match(plate)
+        assert self.new_format_pattern.match(plate)
+
+        plate = faker.license_plate()
+        assert isinstance(plate, str)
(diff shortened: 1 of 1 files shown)


tests/providers/test_automotive.py
```diff
@@ -157,9 +157,15 @@ def test_plate_old_format_explicit_province_prefix(self, faker, num_samples):
             assert plate[:2] == "CA"
 
     def test_plate_format(self, faker):
+        faker.seed_instance(5600)
```
malefice, line 160: Checked the git blame, and I wonder why I did not just generate more samples like I did for the other tests. Anyway, since the `license_plate` switches between two styles 50/50, it is better to just generate more samples instead of seeding. This has the advantage of ensuring (to a degree) that the method can be used multiple times without messing up. I reserve seeding within the test itself for generating rare outcomes or for testing specific conditions.

malefice (commented):
Do not forget to modify the title!

fcurella:
Thank you so much!

PR #1308, opened 2020-10-22
Add nationalities for en_US geo provider.py

andreypetrenko82:
Please add nationalities in English variant

### What does this changes

Adds a possibility to generate random nationalities in English for en_US locale

### What was wrong

Nationalities were absent in English variant


fcurella (changes requested):
Thank you! Could you add a test case?

PR #1310, opened 2020-11-02
add relative frequencies for japanese last names

TianyiShi2001:
### What does this changes

Add relative frequencies for Japanese last names to make its generator more realistic

### What was wrong

This is just an improvement. This technique has been applied to name generators of other locales.

### How this fixes it

Add the frequencies based on https://www2.nipponsoft.co.jp/bldoko/index.asp


faker/providers/person/ja_JP/__init__.py
```diff
@@ -31,16 +32,59 @@ class Provider(PersonProvider):
...
-        '大垣', '加藤', '加納', '喜嶋', '木村', '桐山', '工藤', '小泉',
-        '小林', '近藤', '斉藤', '坂本', '佐々木', '佐藤', '笹田', '鈴木',
-        '杉山', '高橋', '田中', '田辺', '津田', '中島', '中村', '渚',
-        '中津川', '西之園', '野村', '原田', '浜田', '廣川', '藤本', '松本',
-        '三宅', '宮沢', '村山', '山岸', '山口', '山田', '山本', '吉田',
-        '吉本', '若松', '渡辺',
-    )
+    # based on: https://www2.nipponsoft.co.jp/bldoko/index.asp
+    last_names = OrderedDict((
+        "佐藤", 366803,
```
fcurella, line 37: You need to wrap each name/probability (ie: each row) in parenthesis. For an example, see https://github.com/joke2k/faker/blob/master/faker/providers/person/en_NZ/__init__.py#L24

faker/providers/person/ja_JP/__init__.py
```diff
@@ -1,4 +1,5 @@
 from .. import Provider as PersonProvider
+from collections import OrderedDict
```
fcurella, line 2: can you move this import to the top of the file, and separate std-library imports from relative imports with a new line?
```suggestion
from collections import OrderedDict

from .. import Provider as PersonProvider
```

fcurella (changes requested):

fcurella (approved):

fcurella:
Thank you!

PR #1313, opened 2020-11-09
[Japanese/PersonProvider] add name pairs to get matched representation

yu-ichiro:
### What does this changes

add pairs of matching representation to Japanese person provider

### What was wrong

#1310 added frequency to Japanese last names, but it didn't keep the consistency with other representation of it (katakana version and the romanized version)
You could say #1310 broke the consistency, but the original consistency was implicit and hard to tell (you needed to zip the list in order to get a consistent representation) so I came up with a new way.

### How this fixes it

This patch introduces the idea of "pairs" to names and its representation, and you can get a consistent pair if you wanted, or you could get a part of it if you prefer the normal way.

example: (assuming the same seed was provided)
```
>>> fake = Faker(locale='ja_JP')
>>> fake.first_name_pair()
('太郎', 'タロウ', 'Taro')
>>> fake.first_name()
太郎
>>> fake.first_kana_name()
タロウ
>>> fake.first_romanized_name()
Taro
```

### DIFF
--- a/faker/providers/person/ja_JP/__init__.py
+++ b/faker/providers/person/ja_JP/__init__.py
@@ -1,225 +1,243 @@
 from collections import OrderedDict
+from operator import itemgetter
 
 from .. import Provider as PersonProvider
 
 
 class Provider(PersonProvider):
-    formats_female = (
-        '{{last_name}} {{first_name_female}}',
+    # link: http://dic.nicovideo.jp/a/日本人の名前一覧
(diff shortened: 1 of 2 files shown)


fcurella (changes requested):
Thank you! Could you add tests for the new methods?

yu-ichiro:
@fcurella sure! I'll come back to it right away

fcurella:
Thank you!

PR #1314, opened 2020-11-14
Update __init__.py

DeVecchis:
Add random choice of Italian regions.

references: https://en.wikipedia.org/wiki/Regions_of_Italy.


fcurella (changes requested):
Thank you! Could add the reference in the file, and add a new test case?

PR #1316, opened 2020-11-16
[th_TH] Add postcode format, country names, person prefix weights, and update phone number format

bact:
### What does this changes

- Add weights for person name prefixes
- Add more first names
- Add country names
- Add postcode format
- Update phone number format to be more precise and update test
- Add postcode and person name tests

### What was wrong

These are new additions. The old code does not has the functionality or work somehow but may be less precise.

### How this fixes it

- For person name prefixes, some prefixes are in reality occur more often. So add weights with OrderedDict.
- Add country names in Thai language
- Change postcode to limits them to start only with 1-9, and not zero.
- Change phone number formats to make it distinguished between landline (9 digits, starts with 02, 03, 04, 05 or 07) and mobile (10 digits, starts with 06, 08, or 09).
- Add/update tests for postcode, person name, and phone number


fcurella:
Thank you!

PR #1317, opened 2020-11-17
[th_TH] Add SSN, company name, address, and license plate providers

bact:
### What does this changes

- Add social security number (SSN) provider
- Add company name provider
- Add province, district, and sub-district to address provider
- Add license plate provider
- Update jobs

### What was wrong

- Previously there were no SSN and company name providers for th_TH locale.
- Previously available address and license plate providers do not provide a Thailand-style format.
- List of jobs contain a strange job like "spy"

### How this fixes it

- SSN provider provides 13-digit Thai national ID in the format of 1-2345-67890-12-3. The same ID is also used as personal VAT ID.
- Company prefixes and suffixes are added, for regular companies and limited companies. Also include nonprofits (club, foundation, and association).
- New address formats added. Also with list of provinces, amphoes (districts), and tambons (sub-districts).
- New license plate formats added, for cars, motorcycles, and trucks.
- Update jobs (add few more jobs and remove a strange job like "spy")


fcurella:
Thank you!

PR #1320, opened 2020-11-17
Replace image_url provider lorempixel by picsum.photos

### LINKED ISSUE #1319: image_url provider lorempixel is down
* Faker version: all
* OS: all

Provider image_url lorempixel is down

### Steps to reproduce

Enter in lorempixel.com

### Expected behavior

The provider should be online

### Actual behavior

lorempixel.com is down

### Proposal change

I propose changed by http://picsum.photos

Virako:
### What does this changes

Change lorempixel by picsum.photo

### What was wrong

lorempixel is down

### How this fixes it

Change the provider by other that is online

Fixes #1319

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -103,7 +103,7 @@ class Provider(BaseProvider):
         '{{url}}{{uri_path}}/{{uri_page}}{{uri_extension}}',
     )
     image_placeholder_services = (
-        'https://www.lorempixel.com/{width}/{height}',
+        'https://picsum.photos/{width}/{height}',
         'https://dummyimage.com/{width}x{height}',
         'https://placekitten.com/{width}/{height}',
         'https://placeimg.com/{width}/{height}/any',


fcurella (changes requested):
Replacing the services every time one goes temporarily down is not a sustainable strategy. 

I'd rather keep these as defaults, and make them overridable in the method signature.

acron0:
In the case of `lorempixel.com` it's widely reported that this site is unreliable and has frequent outages.

fcurella:
We can replace the defaults, but I still want to make it configurable.

I fear that if we change the defaults, this issue will just be forgotten until the next time a service goes down. And somebody will ask the fix the data again, and the underlying problem won't ever get fixed.

Roconda:
It would be highly appreciated if this PR could be merged. Our tests are constantly failing due to loading images from lorempixel.com while it's down. Thank you!

fcurella:
superseded by https://github.com/joke2k/faker/pull/1580

PR #1321, opened 2020-11-18
[th_TH] Add date_time and bank providers

bact:
### What does this changes

- Add date_time provider for th_TH locale, with Thai day of week and month names, also Buddhist Era.
- Add bank provider, with swift code dataset
- Update female, male, and nonbinary person names

Test-related (generic, not th_TH specific):
- Add tests for profile, md5, sha1, and sha256 providers
- Add additional tests for generic date_time providers (few combinations of parameters)
- Remove unnecessary hasattr check in hu_HU and ru_RU person providers (the class members they checked for are always presented in those locales)
- Mark `pragma: no cover` for empty en_US profile, python, and user_agent providers

### What was wrong

Previously, there's no date_time and bank providers for th_TH

### How this fixes it

- Implement `thai_strftime()`, a localized strftime()
- Override `date()` and `time()` to use thai_strftime()
  - The new date() and time() have `thai_digit` boolean flag to switch the output between Thai digit (True) and Arabic (False). Default is Arabic.
  - The new date() also has `buddhist_era` boolean flag to switch the output between Buddhist Era and standard AD. Default is Buddhist Era (AD+543).
- Also override `century()` to use regular numbers like 18, 19, 20 (instead of Roman numerals)
  - century() also has thai_digit and buddhist_era boolean flag.
  - When buddhist_era is True, the output will be between 1-25.
  - When buddhist_era is False, the output will be between 1-21.

```
>>> faker.date()
'21 พ.ย. 2537'
>>> faker.date(buddhist_era=False)
'18 พ.ย. 1979'
>>> faker.date(thai_digit=True)
'๒๘ ก.ย. ๒๕๑๕'

>>> faker.time()
'14:28:13'
>>> faker.time(thai_digit=True)
'๑๕:๐๗:๕๖'

>>> faker.century()
'23'
>>> faker.century(buddhist_era=False)
'19'
>>> faker.century(thai_digit=True, buddhist_era=False)
'๒๐'
```

- Add bban format (10 digits) and swift code dataset for bank provider

```
>>> faker.bban()
'6860886113'

>>> faker.iban()
'TH715587648248'

>>> faker.swift(use_dataset=True)
'GSBATHB2'
```

### License

- I'm the author of `thai_strftime()`, which is also available from https://gist.github.com/bact/b8afe49cb1ae62913e6c1e899dcddbdb  as public domain/CC0.
- If it is necessary for maintainability of the code base, I also grant the permission here for the faker project to change the license of this piece of code as the maintainer of faker sees fit (like, change it to MIT License or compatible).


bact:
I also change `num_sample_runs` value of filPh's test_date_time from 1,000 to 50, which should be enough for the list of days of week/months with no more than 12 members.

bact:
I also add some tests on some generic date_time() methods and few from other providers, since the coverall keeps put warning about them. This is a bit stretch from the th_TH locale itself.

I think I'm ready for a review. Thank you.

fcurella:
Thank you! ✨

PR #1328, opened 2020-11-27
fix Arabic mistakes and add Arabic currency

itsmohmans:
### What does this change
One change to [providers/persons/ar_AA/__init__.py](https://github.com/joke2k/faker/blob/594d34414b/faker/providers/person/ar_AA/__init__.py)
I also added a new currency for Arabic ar_AA

### What was wrong
There was a wrong name in [providers/persons/ar_AA/__init__.py](https://github.com/joke2k/faker/blob/594d34414b/faker/providers/person/ar_AA/__init__.py)

### Note
This is my very first contribution to a project.


fcurella:
Thank you! Could add some tests for the new currency provider?

itsmohmans:
Can you guide me more on where to add the tests? As I didn't really dig into this project I just localized it. Thanks.

fcurella:
You can copy and modify any of the existing tests for the other currency providers, eg the [Russian](https://github.com/joke2k/faker/blob/master/tests/providers/test_currency.py#L71-L90) one.

itsmohmans:
Done.

tests/providers/test_currency.py
```diff
@@ -68,6 +68,28 @@ def test_cryptocurrency_name(self, faker, num_samples):
...
+    def setup_class(cls):
+        from faker.providers.currency.ar_AA import Provider as ArCurrencyProvider
+        cls.provider = ArCurrencyProvider
+        cls.currencies = cls.provider.currencies
+        cls.currency_codes, cls.currency_names = tuple(zip(*cls.currencies))
+
+    def test_currency(self, faker, num_samples):
+        for _ in range(num_samples):
+            cur = faker.currency()
+            assert isinstance(cur, tuple) and cur in self.currencies
```
fcurella, line 85: can you split this assertion?

```suggestion
            assert isinstance(cur, tuple)
            assert cur in self.currencies
```

tests/providers/test_currency.py
```diff
@@ -68,6 +68,28 @@ def test_cryptocurrency_name(self, faker, num_samples):
...
+
+    def test_currency_name(self, faker, num_samples):
+        for _ in range(num_samples):
+            name = faker.currency_name()
+            assert isinstance(name, str) and name in self.currency_names
```
fcurella, line 90: Same as above:

```suggestion
            assert isinstance(name, str)
            assert name in self.currency_names
```

fcurella (changes requested):

PR #1329, opened 2020-11-28
Remove noqa from init imports, add exposed objects to __all__ variable

mondeja:
### What does this changes

Removes `noqa` statements in `faker/__init__.py` and add `__all__` variable that exposes faker public objects.

### What was wrong

Running `from faker import *` (although it is not a good practice), the `VERSION` variable becomes available to the current namespace.

### How this fixes it

Running `from <module> import *`, if the module has a `__all__` variable, only objects defined inside `__all__` are imported.


fcurella:
Thank you!

PR #1331, opened 2020-11-30
Added Assert statement used outside of tests

HarshCasper:
### What does this changes

I have added Assert statement used outside of tests. This usually follows the modern standards where the assert keyword should be used when failure to meet a condition violates the integrity of the program. 

### What was wrong

In the original Code, asserts were used outside of the Tests. 

### How this fixes it

I have added the Assert Statement used outside of tests by refactoring the Code. 

----------------
I ran a DeepSource Analysis on my fork of this repository. You can see all the issues raised by DeepSource [here](https://deepsource.io/gh/HarshCasper/faker/).

There are plenty of other issues in relation to Bug Discovery, Performance and Anti-Patterns which you would be interested to take a look at.

If you do not want to use DeepSource to continuously analyze this repo, I'll remove the `.deepsource.toml` from this PR and you can merge the rest of the fixes. If you want to setup DeepSource for Continuous Analysis, I can help you set that up.

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,10 @@
+version = 1
+
+test_patterns = ["tests/**"]
+
+[[analyzers]]
+name = "python"
+enabled = true
+
+  [analyzers.meta]
+  runtime_version = "3.x.x"
--- a/faker/utils/distribution.py
+++ b/faker/utils/distribution.py
@@ -22,8 +22,10 @@ def choices_distribution_unique(a, p, random=None, length=1):
     if random is None:
         random = mod_random
 
-    assert len(a) == len(p)
-    assert len(a) >= length, "You can't request more unique samples than elements in the dataset."
+    if len(a) != len(p):
+        raise AssertionError
+    if len(a) < length:
+        raise AssertionError("You can't request more unique samples than elements in the dataset.")
 
     choices = []
     items = list(a)
@@ -45,7 +47,8 @@ def choices_distribution(a, p, random=None, length=1):
     if random is None:
         random = mod_random
 
-    assert len(a) == len(p)
+    if len(a) != len(p):
+        raise AssertionError
 
     if hasattr(random, 'choices'):
         choices = random.choices(a, weights=p, k=length)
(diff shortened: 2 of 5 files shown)


fcurella:
Thank you @HarshCasper !

I'm not sure I understand the difference between using `assert` vs `if ...: raise AssertionError()`. I personally find `assert` more concise and readable, but maybe it's just me?

HarshCasper:
Hello @fcurella 

Basically, Python has an option to compile the optimized bytecode and create the respective .pyo files by using the options -O and -OO. During this process, the Assert statements are removed, and if getting the compiled bytecode, is not an intention, this issue can be ignored. But in the other case, it might cause problems if used in non-test files. That is why it is explicitly checked and a respective Error is raised using an if statement.

faker/utils/distribution.py
```diff
@@ -22,8 +22,10 @@ def choices_distribution_unique(a, p, random=None, length=1):
     if random is None:
         random = mod_random
 
-    assert len(a) == len(p)
-    assert len(a) >= length, "You can't request more unique samples than elements in the dataset."
+    if len(a) != len(p):
+        raise AssertionError
+    if len(a) < length:
+        raise AssertionError("You can't request more unique samples than elements in the dataset.")
```
bact, line 28: should this be ValueError instead?

fcurella:
@HarshCasper I'm ok with adding DeepSource to the CI, but I would add it in a separate PR.

HarshCasper:
> @HarshCasper I'm ok with adding DeepSource to the CI, but I would add it in a separate PR.

@fcurella Should I make a separate PR for the same?

HarshCasper:
Hi @fcurella 

Can you kindly have a look at this again! Thanks :)

.deepsource.toml
```diff
@@ -0,0 +1,10 @@
+version = 1
+
+test_patterns = ["tests/**"]
+
+[[analyzers]]
+name = "python"
+enabled = true
+
+  [analyzers.meta]
+  runtime_version = "3.x.x"
```
fcurella, line 10: Could you move the deepsource config out of this PR and into one of its own?

fcurella (changes requested):

Issue #1334, opened 2020-11-30
Add mypy to typecheck the codebase


nicarl:
* Faker version: master
* OS: all

How about adding mypy (http://mypy-lang.org/) to typecheck the codebase? 
One could start to add type annotations and once the codebase is covered activate mypy type checking in the CI.

UrielMaD:
Hello @nicarl, can I help on typechek the codebase?

nicarl:
Hey that would be great! At the moment I am using the following mypy.ini and mypy 0.790:

```
[mypy]
python_version = 3.8
disallow_subclassing_any = True
disallow_untyped_calls = True
disallow_untyped_defs = True
disallow_incomplete_defs = True
check_untyped_defs = True
warn_redundant_casts = True
warn_unused_ignores = True
warn_unreachable = True
allow_redefinition = True
pretty = True
follow_imports=silent 
```

I think best is to go file by file. The root folder (containing `config.py`, `cli.py` ...) is a good place to start I think.

Should we check in the mypy config and already add the dependency?

UrielMaD:
@nicarl sure! I'm kind of new contributing and I just started to using mypy, I saw that you just merged some files, so, could I start right into the root folder?

nicarl:
Yes you can start in the root folder. You can just ping me if you get stuck.

UrielMaD:
@nicarl Hello again Nicarl, I just did a little implementation of type checking in the cli.py file, but I realize that this topic is deeper than I thought, I'm very interested on dive into this and I was wondering that I could need a little guidence on this

mdantonio:
Hello,
just as information: mypy ignores type hints from packages that do not include a py.typed marker

https://www.python.org/dev/peps/pep-0561/#packaging-type-information

_Package maintainers who wish to support type checking of their code MUST add a marker file named py.typed to their package supporting typing._

If some type hints have been added (even if very partial), would be useful to include the py.typed file to the release to let users to start using type hints and encourage contributions

fcurella:
Thank you @mdantonio , I've added the `py.typed` file in https://github.com/joke2k/faker/commit/52e5018e3fcf5d2b176d2031672c56bc2140ebc9

mdantonio:
Great, I will test it as soon as a new version will be released

sirxbiff:
Now the `py.typed` file is in there mypy complains about `error: Call to untyped function "Faker" in typed context` for a simple line like:
```python
fake = Faker()
```
[...]

PR #1332, opened 2020-11-30
Travis CI: Add Python 3.9 to the testing

cclauss:
### What does this changes

Brief summary of the changes

### What was wrong

Description of what was the root cause of the issue.

### How this fixes it

Description of how the changes fix the issue.

Fixes #...

PR #1335, opened 2020-12-02
Add type annotations for files in /utils

nicarl:
### What does this changes

I added type annotations for the functions in utils.

### What was wrong

In order to introduce type checking I suggested running mypy on the codebase (#1334 ). This is a first attempt to cover a small part of the codebase. 
Benefit for developers should be instantaneous as most IDEs already give some hints.

### How this fixes it

Adding type annotations for function arguments and return values of functions.

### DIFF
--- a/faker/utils/loading.py
+++ b/faker/utils/loading.py
@@ -3,9 +3,11 @@ import pkgutil
 import sys
 
 from importlib import import_module
+from types import ModuleType
+from typing import List, Set
 
 
-def get_path(module):
+def get_path(module: ModuleType) -> str:
     if getattr(sys, 'frozen', False):
         # frozen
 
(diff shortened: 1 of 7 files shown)


nicarl:
It seems that variable annotation (in contrast to annotations of function arguments and return values) were introduced in python 3.6.
I guess you want to keep support for python 3.5? One approach could be to just annotate functions for now.

faker/utils/loading.py
```diff
@@ -37,25 +39,26 @@ def list_module(module):
         return [name for _, name, is_pkg in pkgutil.iter_modules([path]) if is_pkg]
 
 
-def find_available_locales(providers):
-    available_locales = set()
+def find_available_locales(providers: List[str]) -> List[str]:
+    available_locales: Set[str] = set()
```
nicarl, line 43: >= python 3.6 required for variable annotations :-(

fcurella:
We've just deprecated 3.5 with #1332

nicarl:
I took the liberty to also update the config file for appveyor.

fcurella:
Thank you! ✨

PR #1336, opened 2020-12-05
th_TH strftime: normalize output for unsupported directive

### LINKED ISSUE #1333: test_date_pattern fails on 4.18.0
* Faker version: 4.18.0
* OS: Alpine Linux

```
======================================================================
FAIL: test_date_pattern (tests.providers.test_date_time.TestThTh)
----------------------------------------------------------------------
  File "/home/sircmpwn/sources/aports/community/py3-faker/src/Faker-4.18.0/tests/providers/test_date_time.py", line 942, in test_date_pattern
    assert date == "Q"
AssertionError

----------------------------------------------------------------------
```

bact:
### What does this changes

Normalizes output from th_TH's localized version of `strftime()` to make the output of unsupported directives the same in every platforms.

### What was wrong

`strftime()` with unsupported directive returns empty string in `musl` (used in Linux distros like Alpine).
This is different from the output from `glibc`, where it will return a character part (non-%) of the directive.

For example, "%Q" directive will result "Q" on glibc, while it will be "" on musl.

### How this fixes it

Make th_TH strftime() returns the character part (non-%) of the directive, independently from the platform.

Fixes #1333 

Requires #1337 to test against.

### DIFF
--- a/faker/providers/date_time/th_TH/__init__.py
+++ b/faker/providers/date_time/th_TH/__init__.py
@@ -64,11 +64,10 @@ def _std_strftime(dt_obj: datetime, fmt_char: str) -> str:
     str_ = ""
     try:
         str_ = dt_obj.strftime("%{}".format(fmt_char))
-        if str_ == "%{}".format(fmt_char):
+        if not str_ or str_ == "%{}".format(fmt_char):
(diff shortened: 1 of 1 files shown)


fcurella:
Thank you!

PR #1337, opened 2020-12-05
Add Alpine test environment to tox

bact:
### What does this changes

- Add python3-alpine to test environment list
- Update Python requirement/description to 3.6+

### What was wrong

- Current test environments have Ubuntu Focal 20.04 amd64 (Travis), Windows (AppVeyor), and i386/ubuntu (Travis). It does not cover Alpine as part of the test environment. Alpine is a popular Linux distro for container environment.
- See #1336 for a use case of test under different Linux environments. (Faker 4.18.0 and 5.0.0 will fail on Alpine, because of th_TH test_date_pattern - see #1333. #1336 will fix that and this PR can help test a case like that.)

### How this fixes it

- Add build-alpine.sh script and alpine test environment to tox.ini and .travis.yml.
- Also remove metadata/description about 3.5, add 3.9.
- WARNING: This proposal will make the test/CI take longer time. The Alpine test alone can be longer than 10 mins.
  - If this is too long for Travis (I think 10 mins is too long), we can remove it from .travis.yml, which mean that it will be responsibility of the developer to run tox on their own machine.


bact:
This PR will definitely failed until #1336 got merged.

README.rst
```diff
@@ -22,8 +22,8 @@ Faker is heavily inspired by `PHP Faker`_, `Perl Faker`_, and by `Ruby Faker`_.
 Compatibility
 -------------
 
-Starting from version ``4.0.0``, ``Faker`` dropped support for Python 2 and only supports Python
-3.5 and above. If you still need Python 2 compatibility, please install version ``3.0.1`` in the
+Starting from version ``5.0.0``, ``Faker`` dropped support for Python 2 and only supports Python
```
fcurella, line 25: The statement still holds true, I don’t think we need to update this paragraph

fcurella (changes requested):

README.rst
bact: Agreed. Changed back. Thanks.

fcurella:
Thank you!

PR #1338, opened 2020-12-06
Add Python 3.9, drop Python 3.5

eumiro:
Add Python 3.9 and drop Python 3.5 (because after EOL) from `setup.py` and `tox.ini` since they are already in travis and appveyor.

When we are at 3.6+, we can finally move to `f'strings'`.


eumiro:
When we are at 3.6+, we can finally move to `f'strings'`.

fcurella:
Thank you! This change was already merged in #1337

PR #1339, opened 2020-12-06
Use pathlib for dir/file operations

eumiro:
Since the project is Python3 only, we can use `pathlib.Path` instead of all `os.path/glob/fnmatch/open` calls.

### DIFF
--- a/faker/sphinx/documentor.py
+++ b/faker/sphinx/documentor.py
@@ -1,13 +1,14 @@
 # coding=utf-8
 import importlib
 import inspect
-import os
+
+from pathlib import Path
 
 from faker.config import AVAILABLE_LOCALES
 from faker.config import PROVIDERS as STANDARD_PROVIDER_NAMES
 from faker.providers import BaseProvider
 
-DOCS_ROOT = os.path.abspath(os.path.join('..', 'docs'))
+DOCS_ROOT = Path(__file__).resolve().parents[1] / 'docs'
 
 SECTION_ADORNMENTS = '#*=-~'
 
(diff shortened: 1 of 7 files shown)


faker/sphinx/documentor.py
```diff
@@ -104,8 +103,7 @@ def _write_base_provider_docs():
 
 def _write_standard_provider_docs():
     for provider_name in STANDARD_PROVIDER_NAMES:
-        fname = os.path.join(DOCS_ROOT, 'providers', '%s.rst' % provider_name)
-        with open(fname, 'wb') as fh:
+        with (DOCS_ROOT / 'providers' / '{}.rst'.format(provider_name)) as fh:
```
fcurella, line 106: We could use an f-string here:

```suggestion
        with (DOCS_ROOT / 'providers' / f'{provider_name}.rst') as fh:
```

faker/sphinx/documentor.py
```diff
@@ -131,8 +128,7 @@ def _write_localized_provider_index():
 def _write_localized_provider_docs():
     for locale in AVAILABLE_LOCALES:
         info = _get_localized_provider_info(locale)
-        fname = os.path.join(DOCS_ROOT, 'locales', '{}.rst'.format(locale))
-        with open(fname, 'wb') as fh:
+        with (DOCS_ROOT / 'locales' / '{}.rst'.format(locale)).open('wb') as fh:
```
fcurella, line 131: we could use an f-string here too:
```suggestion
        with (DOCS_ROOT / 'locales' / f'{locale}.rst').open('wb') as fh:
```

fcurella (changes requested):

faker/sphinx/documentor.py
eumiro: There are plenty of f-strings there and I waited for the Python 3.6+ merge. I'll do all f-strings in the next step.

faker/sphinx/documentor.py
eumiro: ditto

fcurella (approved):

fcurella:
Thank you! ✨

PR #1341, opened 2020-12-07
use f-strings

eumiro:
We are at Python 3.6+, so replace (most) string formatting with f-strings.
This is based on the current master, so some file names built in #1339 will be added after their merge and this rebase.

### DIFF
--- a/faker/cli.py
+++ b/faker/cli.py
@@ -24,8 +24,7 @@ def print_provider(doc,
         excludes = []
 
     print(file=output)
-    print("### {}".format(
-          doc.get_provider_name(provider)), file=output)
+    print(f"### {doc.get_provider_name(provider)}", file=output)
     print(file=output)
 
     for signature, example in formatters.items():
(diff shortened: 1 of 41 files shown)


eumiro:
Rebased after #1339.

eumiro:
There are also quite many (but not 100%) cases of backticks used to highlight the value of a variable. Do they have any special purpose or would `…{value!r}…` with simple quotes and (what is more important) correctly escaped quotes within the string do the job as well?

fcurella:
@eumiro I think the backticks are just for aesthetic purposes. Can you show me a specific example?

eumiro:
> @eumiro I think the backticks are just for aesthetic purposes. Can you show me a specific example?

https://github.com/eumiro/faker/blob/f-strings/faker/cli.py#L41 contains the `!r` flags in an f-string. I find it better and safer than the explicit backticks elsewhere in the code and would change it accordingly.

fcurella:
@eumiro I thought by "backticks" you meant the <code>`</code> character.

I would just keep the quotes the way they are. I don't really see a compelling reason to change them now.

eumiro:
@fcurella rebased, repushed, fails at coveralls only.

fcurella:
Thank you so much ✨

PR #1342, opened 2020-12-08
Add type annotations for file cli.py

UrielMaD:
issue  #1334
### What does this changes

Add Typechecking for Mypy

### What was wrong

Nothing wrong, just an enhancement


fcurella:
Thank you!

PR #1344, opened 2020-12-08
v5: only python 3.6 and above

thijstriemstra:
### What does this changes

Only python 3.6 is supported now.


fcurella:
Thank you!

PR #1345, opened 2020-12-09
Add link to reference docs in README

Leo-Ryu:
### What does this changes

Hi there!

My name is Leo, a CS student at Brown building an automated platform for editable codebase documentation.

I generated documentation for this repository and placed a link in the README. Please feel free to try it out and let me know what you think!

### What was wrong

### How this fixes it

Fixes #...

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -393,6 +393,11 @@ Write documentation for providers:
 .. code:: bash
 
     $ python -m faker > docs.txt
+    
+
+Reference Docs
+--------------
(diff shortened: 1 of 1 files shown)


fcurella:
Thank you for looking into this.

I agree that we're missing reference docs, but I'd prefer to add them to the existing documentation rather than having docs split in 2 different locations. Could you look into integrating [Sphinx's apidoc](https://www.sphinx-doc.org/en/master/man/sphinx-apidoc.html)?

PR #1357, opened 2020-12-30
en_IN address + phone number provider

pulkitgupta2k:
What does this changes
With this PR, I add a new provider for the Indian address in the English script and the tests for it too. Also added the Indian phone number format.

What was wrong
There wasn't a way to generate a random Indian address and phone number. 

How this fixes it
Added Indian address with the source. Also added test case for the same.

Fixes #...


fcurella:
Thank you! ✨

PR #1360, opened 2021-01-08
Add `upper` argument to bothify provider

ohduran:
### What does this changes

Like `hexify`, it adds the upper keyword argument to capitalize the letters.

### What was wrong

Instead of doing fake.bothify(.....).upper(), include upper as a keyword argument.

### How this fixes it

By checking the upper keyword argument like displayed in the hexify method.


fcurella:
Thank you! Could you add a test case for the new argument?

fcurella (changes requested):
Thank you! Could you add a test case for the new argument?

PR #1368, opened 2021-01-11
Close #1367. Fix lorem provider `sentence` method

### LINKED ISSUE #1367: Random sentence of length 1 fails
* Faker version: 5.5.0
* OS: Ubuntu 20.04.1

Creating a random sentence of length 1 fails because it attempts to titleize the first word of the sentence, but `random_choices`  is returning a tuple which is immutable. 

This also occurs sometimes with 2 or 3 length sentences, although I presume that under the hood they are randomly selecting sentences of length 1.

This works as expected in Faker version 5.4.1.

### Steps to reproduce

```
>>> import faker
>>> thing = faker.Faker()
>>> print(thing.sentence(1))
```

### Expected behavior

A random titleized word is printed with a fullstop

### Actual behavior

```
  File "<input>", line 1, in <module>
  File "~/.../venv/lib/python3.6/site-packages/faker/providers/lorem/__init__.py", line 86, in sentence
    words[0] = words[0].title()
TypeError: 'tuple' object does not support item assignment
```

fcurella:
### What does this changes

Cast the result of `words` in the `sentence` provider to a list.

### What was wrong

The `sentence` provider was trying to modify a tuple.

### How this fixes it

By casting to a list before mutating, we still keep the performance of having `words` returning a tuple.

Fixes #1367

### DIFF
--- a/faker/providers/lorem/__init__.py
+++ b/faker/providers/lorem/__init__.py
@@ -22,7 +22,7 @@ class Provider(BaseProvider):
     sentence_punctuation = '.'
 
     def words(self, nb=3, ext_word_list=None, unique=False):
-        """Generate a list of words.
+        """Generate a tuple of words.
 
         The ``nb`` argument controls the number of words in the resulting list,
         and if ``ext_word_list`` is provided, words from that list will be used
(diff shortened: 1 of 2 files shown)


aimestereo (approved):
yep, that's solve the problem and simplify things,

regarding original issue, I run a few tests and found out the culprit:
```
		# faker/utils/distribution.py:64
        if length == 1 and p is None:
            return (random.choice(a),)
```

then why 2 words were failing?
```
subfaker = faker.Faker(locale=faker.config.DEFAULT_LOCALE)
subfaker.format('sentence', nb_words=2)
```

because it was actually 1 word sentence

```
		# faker/providers/lorem/__init__.py:82
        if variable_nb_words:
            nb_words = self.randomize_nb_elements(nb_words, min=1)

```
here nb_words was set to 1

PR #1372, opened 2021-01-13
ci: enable pip cache

thijstriemstra:
### What does this changes

Enable pip cache and improve job name in github actions

### DIFF
--- a/.github/workflows/test-ubuntu.yml
+++ b/.github/workflows/test-ubuntu.yml
@@ -8,9 +8,17 @@ jobs:
     strategy:
       matrix:
         python: [3.6, 3.7, 3.8, 3.9, "pypy3"]
-
+    name: Python ${{ matrix.python }}
     steps:
       - uses: actions/checkout@v2
+      - name: Setup pip cache
+        uses: actions/cache@v2
+        id: pipcache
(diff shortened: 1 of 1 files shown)


fcurella:
Thank you for looking into this, @thijstriemstra !

I'm not seeing the actions for unit tests running on this branch anymore. Could you have a second look?

thijstriemstra:
> I'm not seeing the actions for unit tests running on this branch anymore.

Me neither, WTF.

.github/workflows/test-ubuntu.yml
```diff
@@ -1,16 +1,28 @@
 name: Python Tests
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [ master ]
```
fcurella, line 5: I think this may be why it's not running when you push to your branch.

thijstriemstra: no it didn't run before either, i changed it to test if it made a difference.

fcurella:
fixed in 50ff1b35a09a940e1a4858c3b938473f116b06e9 and dec10d2b4834403033a5a798578961364405e5f8

PR #1373, opened 2021-01-15
Remove redundant Python 2 int/float conversions

eumiro:
### What does this changes

Remove some useless int/float conversions from the code.

### What was wrong

`a / b` for two integers always returns a float in Python 3, so there's no need to convert `a` or `b` to a `float` beforehand. 

### How this fixes it

There's less astonishment for someone who has never seen Python 2 code before.

### DIFF
--- a/faker/providers/color/color.py
+++ b/faker/providers/color/color.py
@@ -272,11 +272,7 @@ class RandomColor:
         h = 1 if h == 0 else h
         h = 359 if h == 360 else h
 
-        h = float(h)/360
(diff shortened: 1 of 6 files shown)


eumiro:
@fcurella  this also fails at coverall only.

fcurella:
@eumiro we just fixed coveralls. Could you rebase your branch or merge form master?

eumiro:
Finally green! :tada:

fcurella:
Thank you! ❤️

PR #1376, opened 2021-01-19
[th_TH] Update lorem text provider's punctuations, reduce size of word list

bact:
### What does this changes

th_TH LoremProvider:
- Change `word_connector` and `sentence_punctuation` to match usages in Thai language 
- Shorten `word_list` to 400 words (removing compounds and long words)  

### What was wrong

- `word_connector` was previously `" "` (space), while Thai writing system does not use space as a separator
- `sentence_punctuation` was previously `"."` (period), while Thai uses space to separate sentences and phrases
- `word_list` has lots of compounds which are repetition of existing smaller base words, the list also very long considered the purpose of lorem text (some other languages has only around 200 words)

### How this fixes it

- Change `word_connector` to `""` (empty)
- Change `sentence_punctuation` to `" "` (space)
- Reduce the size of `word_list` to 400


fcurella:
Thank you!

PR #1377, opened 2021-01-19
try fixing coveralls

fcurella:


mondeja:
[This](https://github.com/mondeja/mdpo/commit/25e387780f56727cc2af2dc8c972ea1cc6edcb4f#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR41) has worked for me in other repos.

fcurella:
Thank you @mondeja ! @joke2k already tried that approach in https://github.com/joke2k/faker/pull/1374, but it's still not working.

mondeja:
Have you tested the same approach as @joke2k but removing `COVERALLS_REPO_TOKEN`?

joke2k:
> Have you tested the same approach as @joke2k but removing `COVERALLS_REPO_TOKEN`?

not yet. i'm gonna try

fcurella:
welp, coveralls.io's GH action only work with coverage in `lconv.info` file format, and I can't find any way to output in that format (or convert to it) in Python. So this approach seems to be a no-go.

mondeja:
[This change](https://github.com/Zulko/moviepy/pull/1450/files#diff-bd07b6f2904b3965cc56060e7e8e29daa89d25bdf814911e3de55f5206eb2de6R157) worked for me in moviepy, removing `COVERALLS_REPO_TOKEN` and specifying explicitly the service.

joke2k:
seems that @mondeja is right,
https://github.com/joke2k/faker/pull/1374/checks?check_run_id=1728940194

fcurella:
@joke2k could you update #1374?

joke2k:
@fcurella how?
I was thinking that using `COVERALLS_SERVICE_NAME: github` is better then the `--service` argument.
btw it works

joke2k:
> @joke2k could you update #1374?

Do you mean, update with the "Coveralls Parallel" in this PR?

fcurella:
I mean remove `COVERALLS_REPO_TOKEN` in your PR,  #1374

bact:
This #1380 looks like work but I don't know why. Since it's exactly the same as #1374 - -"

PR #1383, opened 2021-01-20
Handle missing ukpostcodeparser gracefully in tests

mgorny:
### What does this changes

The single test is skipped if `ukpostcodeparser` is not installed.

### What was wrong

Nothing was wrong, it's a convenience feature.

### How this fixes it

Using `pytest.importorskip()` to skip the test if `ukpostcodeparser` is not installed.


fcurella:
Thank you!

PR #1391, opened 2021-02-02
Add current_country, current_country_code and administrative_unit

### LINKED ISSUE #1386: Synchronize states and countries?

People often use my Snowfakery tool to write programs that randomly choose a locale and generate an address for the locale. There are two challenges that I'd like to ask for advice on.

1. The name for the field bigger than city but smaller than country varies from place. State, province, prefecture, ... A lot of databases have a single field for this concept: does Faker have a standard way to handle it? Is there any name to ask it to find the middle-region-thing no matter what it is called in that locale.

2. If you set the locale you must also hard-code the country to go with it for the purposes of addresses. Does Faker have within it any easy way to do that mapping or should I to bundle a locale-code -> country mapping?

prescod:
### What does this changes

Add three new methods: current_country, current_country_code and administrative_unit

### What was wrong

1. There was no way to determine the name for the current country
2. There was no generic way to fake "states/provinces/adminastritive_unit" of various countries

Fixes #1386

### DIFF
--- a/faker/documentor.py
+++ b/faker/documentor.py
@@ -1,4 +1,5 @@
 import inspect
+import warnings
 
 
 class Documentor:
@@ -86,7 +87,8 @@ class Documentor:
             try:
                 # make a fake example
                 example = self.generator.format(name, *faker_args, **faker_kwargs)
-            except (AttributeError, ValueError):
+            except (AttributeError, ValueError) as e:
+                warnings.warn(str(e))
                 continue
             formatters[signature] = example
 
--- a/faker/documentor.py
+++ b/faker/documentor.py
@@ -83,9 +83,11 @@ class Documentor:
             # build fake method signature
             signature = f"{prefix}{name}({', '.join(arguments)})"
 
-            # make a fake example
-            example = self.generator.format(name, *faker_args, **faker_kwargs)
-
+            try:
+                # make a fake example
+                example = self.generator.format(name, *faker_args, **faker_kwargs)
+            except (AttributeError, ValueError):
+                continue
             formatters[signature] = example
 
             self.max_name_len = max(self.max_name_len, len(signature))
--- a/faker/providers/address/__init__.py
+++ b/faker/providers/address/__init__.py
@@ -82,3 +82,21 @@ class Provider(BaseProvider):
             return self.random_element(self.alpha_3_country_codes)
         else:
             raise ValueError("`representation` must be one of `alpha-2` or `alpha-3`.")
+
+    def current_country_code(self):
+        try:
+            return self.__lang__.split("_")[1]
+        except IndexError:
+            raise AttributeError("Country code cannot be determined from locale")
+
+    def current_country(self):
+        current_country_code = self.current_country_code()
+        current_country = [tz['name']
+                           for tz in date_time.Provider.countries
+                           if tz['alpha-2-code'] == current_country_code]
+        if len(current_country) == 1:
+            return current_country[0]
+        elif len(current_country) > 1:
+            raise ValueError(f"Ambiguous country for country code {current_country_code}: {current_country}")
+        else:
+            raise ValueError(f"No appropriate country for country code {current_country_code}")
--- a/faker/providers/address/fr_CH/__init__.py
+++ b/faker/providers/address/fr_CH/__init__.py
@@ -116,12 +116,14 @@ class Provider(AddressProvider):
         """
         return self.random_element(self.cantons)
 
-    def canton_name(self):
+    def administrative_unit(self):
         """
         Randomly returns a Swiss canton name.
         :example 'Vaud'
         """
         return self.canton()[1]
+
+    canton_name = administrative_unit
 
     def canton_code(self):
         """
--- a/faker/providers/address/fr_FR/__init__.py
+++ b/faker/providers/address/fr_FR/__init__.py
@@ -162,14 +162,12 @@ class Provider(AddressProvider):
         """
         return self.random_element(self.departments)
 
-    def administrative_unit(self):
+    def department_name(self):
         """
         Randomly returns a french department name.
         :example 'Ardèche'
         """
         return self.department()[1]
-
-    department_name = administrative_unit
 
     def department_number(self):
         """
--- a/faker/providers/address/fr_FR/__init__.py
+++ b/faker/providers/address/fr_FR/__init__.py
@@ -147,11 +147,13 @@ class Provider(AddressProvider):
         """
         return self.random_element(self.city_prefixes)
 
-    def region(self):
+    def administrative_unit(self):
         """
         :example 'Guadeloupe'
         """
         return self.random_element(self.regions)
+
+    region = administrative_unit
 
     def department(self):
         """
@@ -160,12 +162,14 @@ class Provider(AddressProvider):
         """
         return self.random_element(self.departments)
 
-    def department_name(self):
+    def administrative_unit(self):
         """
         Randomly returns a french department name.
         :example 'Ardèche'
         """
         return self.department()[1]
+
+    department_name = administrative_unit
 
     def department_number(self):
         """
--- a/faker/providers/address/cs_CZ/__init__.py
+++ b/faker/providers/address/cs_CZ/__init__.py
@@ -760,8 +760,10 @@ class Provider(AddressProvider):
     def street_name(self):
         return self.random_element(self.streets)
 
-    def state(self):
+    def administrative_unit(self):
         return self.random_element(self.states)
+
+    state = administrative_unit
 
     def city_with_postcode(self):
         return self.postcode() + " " + self.random_element(self.cities)
--- a/faker/providers/address/da_DK/__init__.py
+++ b/faker/providers/address/da_DK/__init__.py
@@ -103,5 +103,7 @@ class Provider(AddressProvider):
     def city_name(self):
         return self.random_element(self.cities)
 
-    def state(self):
+    def administrative_unit(self):
         return self.random_element(self.states)
+
+    state = administrative_unit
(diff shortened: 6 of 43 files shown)


prescod:
@fcurella Please review before I look into improving the coverage. If you don't like the overall direction then it isn't worth improving the coverage.

faker/providers/address/fr_CH/__init__.py
```diff
@@ -123,6 +123,8 @@ def canton_name(self):
         """
         return self.canton()[1]
 
+    administrative_unit = canton_name
```
fcurella, line 126: `administrative_unit` should be the method with logic in, and then we should have `canton_name = administrative_unit`

faker/providers/address/__init__.py
```diff
@@ -82,3 +82,40 @@ def country_code(self, representation=ALPHA_2):
             return self.random_element(self.alpha_3_country_codes)
         else:
             raise ValueError("`representation` must be one of `alpha-2` or `alpha-3`.")
+
+    def administrative_unit(self):
+        for name in ["province", "region", "state", "prefecture", "county"]:
```
fcurella, line 87: I don't like this 'guessing'. Feels t0o magical. I'd rather force the providers to have an `administrative_unit` method by using unit tests.

faker/providers/address/__init__.py
```diff
@@ -82,3 +82,40 @@ def country_code(self, representation=ALPHA_2):
...
+        try:
+            return self.__lang__.split("_")[1]
+        except IndexError:
+            raise AttributeError("Country code cannot be determmined from locale")
+
+    def current_country(self):
+        try:
+            current_country_code = self.current_country_code()
+        except AttributeError:
+            return "No Country"
```
fcurella, line 104: I'd rather just raise the exception than returning an invalid value.

faker/providers/address/__init__.py
```diff
@@ -82,3 +82,40 @@ def country_code(self, representation=ALPHA_2):
...
+        current_country = [tz['name']
+                           for tz in date_time.Provider.countries
+                           if tz['alpha-2-code'] == current_country_code]
+        if len(current_country) == 1:
+            return current_country[0]
+        elif len(current_country) > 1:
+            raise ValueError(f"Ambiguous country for country code {current_country_code}: {current_country}")
```
fcurella, line 111: are there actually any countries that share the same code? Would this ever happen?

faker/providers/address/__init__.py
```diff
@@ -82,3 +82,40 @@ def country_code(self, representation=ALPHA_2):
...
+        elif current_country_code == "AA":
```
fcurella, line 112: I think the [`ar_AA` provider](https://github.com/joke2k/faker/blob/master/faker/providers/date_time/ar_AA/__init__.py) has actually a mistake,the wrong name. It should be called just `ar`, like we have `en` and `es`.

faker/providers/address/__init__.py
```diff
@@ -82,3 +82,40 @@ def country_code(self, representation=ALPHA_2):
...
+            return "No Country"
+        elif current_country_code == "PS":
+            return "Palestine"
+        elif current_country_code == "TW":
+            return "Taiwan"
+        elif current_country_code == "QC":
```
fcurella, line 118: `fr_QC` is wrongly named too. The ISO code should be `fr_CA`. You'd think people would know their own locale code.

faker/providers/address/__init__.py
```diff
@@ -82,3 +82,40 @@ def country_code(self, representation=ALPHA_2):
...
+        if len(current_country) == 1:
+            return current_country[0]
+        elif len(current_country) > 1:
+            raise ValueError(f"Ambiguous country for country code {current_country_code}: {current_country}")
+        elif current_country_code == "AA":
+            return "No Country"
+        elif current_country_code == "PS":
+            return "Palestine"
+        elif current_country_code == "TW":
+            return "Taiwan"
```
fcurella, line 117: Same here. `TW` should work without special-case. Every provider should have it. Where is it failing?

Would be nice if we had a canonical list of countries, and we can enforce to providers to name them all.

faker/providers/address/__init__.py
```diff
@@ -82,3 +82,40 @@ def country_code(self, representation=ALPHA_2):
...
+        current_country = [tz['name']
+                           for tz in date_time.Provider.countries
+                           if tz['alpha-2-code'] == current_country_code]
+        if len(current_country) == 1:
+            return current_country[0]
+        elif len(current_country) > 1:
+            raise ValueError(f"Ambiguous country for country code {current_country_code}: {current_country}")
+        elif current_country_code == "AA":
+            return "No Country"
+        elif current_country_code == "PS":
```
fcurella, line 114: `PS` should work without special-case. Every provider should have it. Where is it failing?

fcurella (changes requested):

faker/providers/address/__init__.py
prescod: So just to be clear: you prefer me to go to each locale, find the attribute called "state", "province", etc.

Rename that to "administrative_unit"

Then alias it back to the original name?

faker/providers/address/__init__.py
prescod: It's just defensive programming. I don't control whoever edits that list later.

faker/providers/address/__init__.py
prescod: Taiwan does not have a time-zone entry in date_time.countries. Perhaps the time-zone people are not as open to Taiwanese independence as the 2-letter ISO people. Same for Palestine.

faker/providers/address/__init__.py
prescod: Palestine does not have a time-zone entry in date_time.countries. Perhaps the time-zone people are not as open to Palestinian independence as the 2-letter ISO people. Same for Taiwan.

faker/providers/address/__init__.py
fcurella: Correct. I do realize how labor-intensive (and boring) it is, but I think it would be more correct for `administrative_unit` to be "canonical" method, and localized terminology to just be a translation of the canonical name.

faker/providers/address/__init__.py
fcurella: 👍🏼

faker/providers/address/__init__.py
fcurella: Taiwan has a zone in the [IANA database](https://en.wikipedia.org/wiki/Time_in_Taiwan#IANA_time_zone_database) as `Asia/Taipei`. We should probably add it to `date_time.countries`.

faker/providers/address/__init__.py
fcurella: I see 2 timezone in the [TZ Database](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) for `PS`: `Asia/Gaza` and `Asia/Hebron`. We should probably add both?

faker/providers/address/__init__.py
prescod: Renaming it causes an exception. Faker asks Python to normalize it and Python returns ar_AA

```
>>> import locale
>>> locale.normalize("ar")
'ar_AA.ISO8859-6'
```

prescod: @fcurella ^

fcurella: thank you for looking into it. We'll special-case it if we have to. 👍🏼

prescod:
@fcurella : Looks like this is ready now that the other PRs are in and this is updated to reflect them.

fcurella (changes requested):

faker/documentor.py
```diff
@@ -83,9 +83,11 @@ def get_provider_formatters(self, provider, prefix='fake.',
...
             signature = f"{prefix}{name}({', '.join(arguments)})"
 
-            # make a fake example
-            example = self.generator.format(name, *faker_args, **faker_kwargs)
-
+            try:
+                # make a fake example
+                example = self.generator.format(name, *faker_args, **faker_kwargs)
+            except (AttributeError, ValueError):
+                continue
```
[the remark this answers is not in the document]
prescod: What convention do you use for logging?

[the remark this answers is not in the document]
prescod: Just warnings.warn?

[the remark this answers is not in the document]
fcurella: yes, `warnings.warn` will do :)

faker/providers/address/fr_FR/__init__.py
```diff
@@ -147,26 +147,30 @@ def city_prefix(self):
...
 
-    def department_name(self):
+    def administrative_unit(self):
         """
         Randomly returns a french department name.
         :example 'Ardèche'
         """
         return self.department()[1]
 
+    department_name = administrative_unit
```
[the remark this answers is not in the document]
prescod: Google (and the post office) seems to just use city names. The departments and regions are not postal units.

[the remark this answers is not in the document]
prescod: @fcurella Any thoughts on this?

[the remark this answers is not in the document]
fcurella: The Google Geocoding API uses `region` for `administrative_area_level_1` (level 1 is the level just below the country). Eg [Paris, Île-de-France](https://developers-dot-devsite-v2-prod.appspot.com/maps/documentation/utils/geocoder#place_id%3DChIJD7fiBh9u5kcRYJSMaMOCCwQ)

prescod:
@fcurella Okay it uses region = administrative_unit in fr_fr and warnings.warn for missing properties. Let me know if anything else has to change.

fcurella (approved):

fcurella:
Thank you! ❤️

PR #1393, opened 2021-02-03
Fix pyfloat returning zeros when positive=True

### LINKED ISSUE #1387: pyfloat provider with positive=True sometimes returns zero
* Faker version: 5.8.0
* OS: macOS 11.1

Call pyfloat with positive=True, and it will sometimes return a zero. 

### Steps to reproduce

Here's an example:

```
import faker
f=faker.Faker()
are_zero = [f.pyfloat(positive=True, max_value=10)==0 for _ in range(100)]
print(any(are_zero))
```

### Expected behavior

It should always print `False` always. 

### Actual behavior

Sometimes it's `True` with f.pyfloat() returning zero.

sciencectn:
### What does this change

When `pyfloat()` is about to return a zero, this adds the machine epsilon to make it barely positive. 

Fixes #1387

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -87,7 +87,10 @@ class Provider(BaseProvider):
             sign = '+' if positive else self.random_element(('+', '-'))
             left_number = self.random_number(left_digits)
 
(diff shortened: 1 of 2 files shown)


fcurella:
Looks good! Thank you! ✨

PR #1395, opened 2021-02-09
Rename fr_QC to fr_CA

prescod:
### What does this changes

Rename fr_QC to fr_CA

### What was wrong

fr_QC is not a recognized locality because QC is not a recognized country code.

### How this fixes it

Rename fr_QC to fr_CA and leave behind a deprecation note.

Does not output a deprecation message.

Fixes #...


faker/providers/person/fr_QC/__init__.py
```diff
@@ -1,481 +1,3 @@
-from .. import Provider as PersonProvider
+from ..fr_CA import *
```
fcurella, line 1: I would rather have a subclass of the `fr_CA` that emits a warning on `__init__`:

```python3
import warnings

from ..fr_CA import Provider as FRCAProvider


class Provider(FRCAProvider):
    def __init__(self,, *args, **kwargs):
        warnings.warn(...)
        super().__init__(*args, **kwargs)

```

fcurella (changes requested):

fcurella (approved):

fcurella:
Thank you!

PR #1396, opened 2021-02-09
Add Taiwanese and Palestinian time zones.

prescod:
### What does this changes

Adds Taiwanese and Palestinian time zones.

### What was wrong

Taiwanese and Palestinian time zones were missing.


prescod:
I did not reduce coverage. Coveralls can be annoying that way. Probably floating point bugs.

fcurella (approved):

fcurella:
> I did not reduce coverage. Coveralls can be annoying that way. Probably floating point bugs.

Yeah, it's fine. Don't sweat it :)

PR #1399, opened 2021-02-17
Add cs_CZ lorem - word list

jakubno:
### What does this changes

Add Czech lorem


fcurella:
Thank you! Do you mind adding a test case?

jakubno:
> Thank you! Do you mind adding a test case?

Sure, will do! Do you want to test all the methods of lorem (`paragraph`, `paragraphs`, `word`, `words`, `text`, `texts`)? Because all of them use the same word list, right?

fcurella:
Test for each methods will be better to prevent regressions

jakubno:
What do you think, should all names with first capital letter (countries, cities) be removed or lower cased?

fcurella:
@jakubno I would say keep them the way they were in the original dataset

jakubno:
I hope it's fine like that, but I am open to any suggestion how improve the tests :)

fcurella:
Thank you! ✨

PR #1403, opened 2021-02-24
Add addresses for de_CH

andre-dasilva:
### What does this changes

Add address formats for de_CH


fcurella:
Thank you! ✨

faker/providers/address/de_CH/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+               ('VD', 'Waadt'), ('VS', 'Wallis'), ('ZG', 'Zug'), ('ZH', 'Zürich'))
+
+    def canton(self):
+        """
+        Randomly returns a swiss canton ('Abbreviated', 'Name').
+        :example ('ZH', 'Zürich')
+        """
+        return self.random_element(self.cantons)
+
+    def administrative_unit(self):
```
m-aciek, line 30: Can we add an alias from state() to administrative_unit()? Not adding the state breaks backwards compatibility.

@fcurella @andre-dasilva

m-aciek:
```
  File "/usr/local/lib/python3.9/site-packages/faker/proxy.py", line 105, in __getattr__
    return getattr(self._factories[0], attr)
AttributeError: 'Generator' object has no attribute 'state'
```

andre-dasilva: @m-aciek thats good for me.

PR #1404, opened 2021-02-25
Just fixed a typo

mdantonio:
### What does this changes

It is a very simple typo

### What was wrong

wrong word: "minium"

### How this fixes it

Correct word: "minimum"


fcurella:
Thank you!

PR #1405, opened 2021-03-06
Replace rfc name initials when they are forbidden words

briferz:
### What does this changes

This PR modifies the RFC generation so it will never return one with forbidden words.

### What was wrong

Fake natural person RFCs were being generated with forbidden words in them.

### How this fixes it

If the name initials are among the forbidden words, they are replaced with their corresponding alternative.


fcurella:
Thank you!

Issue #1402, opened 2021-02-23
Calling pyfloat with `right_digits` and `positive` set can yield much smaller number (sys.float_info.epsilon)


tomage:
* Faker version: 6.4.1 (latest as of submitting this issue)
* OS: OSX 10.14.6

Calling pyfloat with `left_digits=0`, `right_digits=N` and `positive=True` can yield a value with way more right digits than `N`.

### Steps to reproduce

This _can_ happen (if you're "lucky"):
```python
>>> import faker
>>> print(faker.Faker().pyfloat(left_digits=0, right_digits=2, positive=True))
2.220446049250313e-16
```

A more surefire way to reproduce this is to run in a for-loop, i.e.:
```python
>>> import faker
>>> my_faker = faker.Faker()
>>> for _ in range(1000):
...     f = my_faker.pyfloat(left_digits=0, right_digits=2, positive=True)
...     if f < 0.001:
...         print('This is odd!', f)
```

Out of the 1000 calls, about 10 should print out `This is odd! 2.220446049250313e-16`

This can also be reproduced with `left_digits=1`, with less frequency (and probably for any value of `left_digits` - just less likely), and also with various values of `right_digits`. `right_digits=1` makes this happen roughly every 10th call to the function.

### Expected behavior

If `left_digits=0` `right_digits=2` and `positive=True` - I would expect all values to be in the interval `[.01 -> .99]`.

Actually it might even be debatable if `left_digits=0` has really any meaning. Perhaps it means that the number is always `0.<right-side>`? Whereas `left_digits=1` means it can be `\d.<right-side>`.

### Actual behavior

Occasonally (every 10th/100th/etc.. call, depending on input parameters) a value of `2.220446049250313e-16` pops out.

### Analysis/suggestions

Reading thru the source code, I see the culprit. `Provider.pyfloat` _can_ end up with a `0` close to the end of that function. But if that is the case, and `positive == True`, it adds `sys.float_info.epsilon`. This does make sense on the outset - but not if the caller has also stated they want maximum of `N` `right_digits`.

I monkey patched the lib in my code for now to call `pyfloat` again and again, until I get a "valid" value.

[...]

PR #1406, opened 2021-03-07
Fix a problem with pyfloat where it can return `sys.epsilon` when `right_digits=N` and `positive=True`

tomage:
### What does this changes

Fixes issue https://github.com/joke2k/faker/issues/1402, where a call to `pyfloat` (or other methods that call it, i.e. pydecimal) can result in the smallest possible float value the system can reprisent, even if the end user had set `right_digits` to some particular number.

I.e. could end up with `2.220446049250313e-16` (number with 15 leading zeroes on the right-hand side of the dot), even if `right_digits` had been, say, `2`.

### What was wrong

Calling pytest with `right_digits=2` should never result in any number where the right-hand side goes beyond `.01`.

### How this fixes it

If we hit that odd/rare situation where we're about to return the epsilon, the function does a quick check to see if we had limits on the # of right-hand-side digits, and if so, instead of using epsilon, we go for the next-best thing - the smallest possible right-hand-side number, i.e. `.01` for N=2. Or `0.0000001` for N=7.

It's not necessarily the perfect solution, but should do the trick.

# Notes

* I did implement this in 2 commits - but I'm more than happy to squash these into a single comment if you like. The first commit highlights the issue by adding a test that showcases the behavior. The second commit is the fix.
* I noticed that the tests in the module I was editing are written in the "old-school" `unittest.TestCase` form. I figured I could write it how I usually see pytest tests written, while leaving the original set of tests be. I can also definitely change this, and add the new test(s) to fit in the style that the test module is using.

Tagging @fcurella since we discussed the problem in the issue (https://github.com/joke2k/faker/issues/1402).

B.S.: I still am aiming to fix the problem of never actually getting a leading zero in the right-hand side, but I'll tackle that once this bug has been squashed.

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -89,7 +89,10 @@ class Provider(BaseProvider):
 
         result = float(f'{sign}{left_number}.{self.random_number(right_digits)}')
         if positive and result == 0:
-            result += sys.float_info.epsilon
+            if right_digits:
+                result = float('0.' + '0' * (right_digits - 1) + '1')
+            else:
+                result += sys.float_info.epsilon
         return result
 
     def _safe_random_int(self, min_value, max_value, positive):
--- a/tests/providers/test_python.py
+++ b/tests/providers/test_python.py
@@ -3,7 +3,31 @@ import warnings
 
 from unittest.mock import patch
 
+import pytest
+
 from faker import Faker
+
+
+@pytest.mark.parametrize(
+    'mock_random_number_source, right_digits, expected_decimal_part',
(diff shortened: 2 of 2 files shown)


tomage:
(Note! I was a bit premature in creating the PR - see some lints failing - I'll fix and amend my comments ASAP)

tomage:
FYI - fixed the lints, and force-pushed out (decided to force push since the changes were trivial, and the patch is very new and easy to read). Should be all good-to-go!

fcurella:
Thank you so much!

PR #1407, opened 2021-03-08
Avoid too many digits problem with `pyfloat()`

tomage:
### What does this change

It was possible for someone to call `pyfloat(left_digits=10, right_digits=10)`, and get back a value that did not have that many digits on left- and right-hand side of the decimal point.

On my system, a typical result of this kind of call would be something like `1234567890.1234567`, with 10 digits on the left hand side, but only 7 on the right hand.

### What was wrong

This is due to the limitations (on most (all, really) systems) w.r.t. implementation of floating point numbers. On my system, the maximum amount of numbers ensured to be properly kept is 15. The example above would ask for 20 numbers.

"But, in the example above are 17 digits total!" I hear you say. Well, yeah. but this is about "guaranteed" accurate number of digits. Highly recommend reading up on the IEEE 754 standard (https://ciechanow.ski/exposing-floating-point/ and https://en.wikipedia.org/wiki/IEEE_754 for those interested).

### How this fixes it

It makes `pyfloat()` raise an exception if the user explicitly calls for a float with more digits than the system can even represent.

### Notes

@fcurella : This isn't precisely the bug I had reported on the side here https://github.com/joke2k/faker/issues/1402 , but it seemed pertinent to just send this PR out rather than repeating this in an issue as well. I can 100% do so though if you'd prefer, for "keeping to the process"' sake.

On that side-issue reported (not being able to get values such as `0.01`, I'm planning to follow this up with a PR no that, but saw this opportunity to harden the function a bit first.

Added as 2 commits as before: First one to add a test showcasing the issue, second that brings in the fix. Can squash if preferred, tho it seems you do indeed do that upon merging in.

### DIFF
--- a/tests/providers/test_python.py
+++ b/tests/providers/test_python.py
@@ -46,14 +46,23 @@ def test_pyfloat_right_or_left_digit_overflow():
 
     with patch('faker.providers.BaseProvider.random_int', mock_random_int):
         with patch('faker.providers.BaseProvider.random_number', mock_random_number):
-            result = faker.pyfloat(left_digits=10, right_digits=10, positive=True)
-            assert str(result) == '1234567890.1234567'
 
-            result = faker.pyfloat(left_digits=17, positive=True)
-            assert str(result) == '1.2345678901234568e+16'
+            # A bit too much, but ~half on either side
+            with pytest.raises(ValueError, match='Asking for too many digits'):
+                faker.pyfloat(left_digits=max_float_digits // 2 + 1, right_digits=max_float_digits // 2 + 1)
 
-            result = faker.pyfloat(right_digits=17, positive=True)
-            assert str(result) == '123456789012345.12'
+            # Asking for max digits on either side also fails, because we need one digit on the other side, i.e.
+            # 0.123123123, or 123123123.0 (at least needs to lead with `0.` or trail with `.0`).
+            with pytest.raises(ValueError, match='Asking for too many digits'):
+                faker.pyfloat(left_digits=max_float_digits)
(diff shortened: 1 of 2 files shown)


tests/providers/test_python.py
```diff
@@ -30,6 +31,40 @@ def mock_random_number(self, digits=None, fix_len=False):
...
+
+    with patch('faker.providers.BaseProvider.random_int', mock_random_int):
+        with patch('faker.providers.BaseProvider.random_number', mock_random_number):
+
+            # A bit too much, but ~half on either side
+            with pytest.raises(ValueError, match='Asking for too many digits'):
+                faker.pyfloat(left_digits=max_float_digits // 2 + 1, right_digits=max_float_digits // 2 + 1)
+
+            # Asking for max digits on either side also fails, because we need one digit on the other side, i.e.
+            # 0.123123123, or 123123123.0 (at least needs to lead with `0.` or trail with `.0`).
```
tomage, line 55: TBH - I don't think this is strictly true.. It's possible we can squeeze out one more digit, so there's that.
I honestly didn't look super close into that aspect - but I also figured not many people (if any) are working right at the edges of that their system supports w.r.t. floats.
I can be convinced to take some time and nail this down 100% tho.

fcurella: I think it's good enough for now. Thank you so much!

fcurella:
Thank you so much!

tomage:
> Thank you so much!

No problemo! Happy to lend a hand, as long as it's useful :)

PR #1408, opened 2021-03-12
uk_UA address street titles have been set to proper language

fessua:
### What does this changes

This changes `'street_titles'` in providers.address.uk_UA 

### What was wrong

The values of `'street_titles'` in providers.address.uk_UA were not relative to Ukraine and not ukrainian language.

### How this fixes it

The values of `'street_titles'` were replaced to proper values in ukrainian language and relative to Ukraine.

Fixes #...


fcurella:
Thank you!

PR #1411, opened 2021-03-16
Fix ro_RO phone numbers generation.

feketemihai:
### What does this changes

Update Romanian phone numbers generation

### What was wrong

Generate x instead of numbers

### How this fixes it

Fixes #...


fcurella:
Thank you!

PR #1416, opened 2021-03-24
Ensures pytuple returns the correct number of items

### LINKED ISSUE #1415: pytuple can occassionally return fewer than expected elements
* Faker version: 6.1.1
* OS: Mac OSX 10.15.5

Sporadically using pytuple may result in the tuple having less than the requested elements (`nb_elements`) even when `variable_nb_elements` is set to False.

This happens because pytuple relies on pyset, returning `tuple(self.pyset(...))`. Because it delegates to a set rather than a list, any duplicate numbers generated will result in the set, and following that the tuple, having fewer numbers than expected.

Suggest that the appropriate fix might be to use `pylist` instead of `pyset`

### Steps to reproduce

1. Specify `nb_elements = 3` in a call to pytuple, `variable_nb_elements= False`
2. Repeat until the tuple 'randomly' contains less than 3 elements  

```python
import faker

fake = faker.Faker()

for x in range(10000):
    random_tuple = fake.pytuple(nb_elements=3, variable_nb_elements=False, value_types=[int])
    assert len(random_tuple) == 3, f"Tuple {random_tuple} not len 3 at iteration {x}"
```

### Expected behavior

When calling pytuple with `nb_elements = 3` and `variable_nb_elements = False` the tuple should always contain 3 elements, even if there are duplicate values.

### Actual behavior

Sporadically the tuple contains less than 3 elements.

rossjones:
### What does this changes

Currently, pytuple uses pyset, which means that if the generates happens
to generate duplicate items, the resulting set will only include on
instances of that item. When asking for explicitly n items, it is
possible to get a number of items that is anywhere between 1 and n inc.

By changing to use a _pyiterable as the pyset and pylist methods do, the
uniqueness of the values generated is no longer required and it behaves
correctly.

### What was wrong

The pytuple() method was incorrectly using pyset() to generate values 
instead of the more appropriate _pyiterable().

### How this fixes it

Ensures the values generated are passed to `tuple` directly, without 
being reduced in number by pyset.

Fixes #1415

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -123,7 +123,7 @@ class Provider(BaseProvider):
 
     def pytuple(self, nb_elements=10, variable_nb_elements=True, value_types=None, *allowed_types):
         return tuple(
-            self.pyset(
+            self._pyiterable(
                 nb_elements,
                 variable_nb_elements,
                 value_types,
--- a/tests/providers/test_python.py
+++ b/tests/providers/test_python.py
@@ -195,6 +195,21 @@ class TestPython(unittest.TestCase):
         with self.assertRaises(AssertionError):
             self.factory.pystr(min_chars=5, max_chars=5)
 
+    def test_pytuple(self):
+        with warnings.catch_warnings(record=True) as w:
+            some_tuple = Faker().pytuple()
+            assert len(w) == 0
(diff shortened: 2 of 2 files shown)


fcurella:
Thank you!

PR #1426, opened 2021-04-09
Extend Pesel To support greater range of dates

IlfirinPL:
### What does this changes

Allow to generate proper pesel for dates from range 1800-2299

### What was wrong

For dates outside 1900-2099 pesel generate incorrect values.

Information about generation is under link https://en.wikipedia.org/wiki/PESEL in section Birthdates

### How this fixes it

modify month date
The PESEL system has been designed to cover five centuries. To distinguish people born in different centuries, numbers are added to the MM field:
for birthdates between 1900 and 1999 – no change to MM field is made (see below)
for other birthdates:
2000–2099 – month field number is increased by 20
2100–2199 – month + 40
2200–2299 – month + 60
1800–1899 – month + 80

Fixes #...

### DIFF
--- a/faker/providers/person/pl_PL/__init__.py
+++ b/faker/providers/person/pl_PL/__init__.py
@@ -733,7 +733,7 @@ class Provider(PersonProvider):
             month = date_of_birth.month + 60
         else:
             raise ValueError("Date of birth is out of supported range 1800-2299")
-        
+
         pesel_date = f'{date_of_birth:%y}{month:02d}{date_of_birth:%d}'
 
         pesel_core = ''.join(map(str, (self.random_digit() for _ in range(3))))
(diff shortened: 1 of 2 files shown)


fcurella (changes requested):
Thank you! Could you add some tests for these scenarios?

IlfirinPL:
@fcurella  Yes, Will write some tests

IlfirinPL:
@fcurella  Tests added, documentation for function also update.

IlfirinPL:
@fcurella  
I see that formater on windows doesn't like dates old dates  "format %y requires year >= 1900 on Windows"
https://ci.appveyor.com/project/joke2k/faker/builds/38675793/job/kue9e4iel8w4nwyb#L1364
what do you this should be done skip those tests for windows or change way of extraction of values ?

IlfirinPL:
Update code with workaround issue with dates below 1900

fcurella:
Thank you!

PR #1431, opened 2021-04-22
Add tlds for th_TH internet provider

bact:
### What does this changes

- Add `tlds` for th_TH, in internet provider

### What was wrong

Internet provider previously not has th_TH locale

### How this fixes it

Add user name formats and TLDs for th_TH


fcurella (changes requested):
Thank you! Do you mind adding a test for the new provider?

faker/providers/person/th_TH/__init__.py
```diff
@@ -804,7 +804,7 @@ class Provider(PersonProvider):
         "ถนัดภาษา",
         "ถนัดรบ",
         "ถนัดรักษา",
-        "ถนัดหัดถกรรม",
+        "ถนัดหัตถกรรม",
```
fcurella, line 807: I dont see how those two lines are different. Why are they sowing up in the diff?

bact: They are two different characters, with a very similar shape (thus very common typo).
One is ด, another with ต.
The only difference is ต has a bend at the top curve of the character, but ด doesn't have the bend.

bact:
I will add the test.

faker/providers/internet/th_TH/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+            ("go.th", 40),
+            ("or.th", 40),
+            ("ac.th", 20),
+            ("net.th", 10),
+            ("mi.th", 5),
+            ("co", 10),
+            ("net", 20),
+            ("com", 150),
+            ("org", 50),
+        )
```
fcurella, line 31: can you add a trailing comma here?
```suggestion
        ),
```

fcurella (changes requested):

fcurella (approved):
Thank you!

PR #1433, opened 2021-04-25
doc: add OptionalProvider to community maintained providers

lyz-code:
### What does this changes

Add the [OptionalProvider](https://lyz-code.github.io/faker-optional/) to the community providers.

### What was wrong

It was "not easy" create data of type `Optional[Any]` with the existent providers.

### How this fixes it

This provider gives a wrapper over existent providers to return `None` some of the times.

```python
from faker import Faker
from faker_optional import OptionalProvider

fake = Faker()
fake.add_provider(OptionalProvider)

fake.optional_int()
# None

fake.optional_int()
# 1234
```


fcurella:
Thank you!

PR #1440, opened 2021-05-08
Create sk_SK person provider

slavkoja:
### What does this changes

It adds sk_SK person provider

### What was wrong

 sk_SK person provider is missing

### How this fixes it

The file was based on cs_CZ provider, but Slovak first and last names are used


fcurella (changes requested):
Thank you!

Could you please include the source of your data?

slavkoja:
Sorry, i miss your comment.

I took first names from our calendar (name days), from where i manually filter 100 "normal" names -- there are some, by my opinion, "exotic" or historic names, which i excluded.

The last names i start to collect from my memory, but this was enough, then i used https://slovnik.juls.savba.sk/?d=priezviska (our Academy of Science) from where i manually pick them to get 50.

Is this info enough?

fcurella:
Are there any statistical dataset that we can use for both first and last names?

slavkoja:
I am sorry, i do not understand what you want.

fcurella:
@slavkoja Is there any list of names compiled from official statistical data (eg: government census), rather than just made up by you?

The wikipedia page for ["Slovakian Names"](https://en.wikipedia.org/wiki/Slovak_name#cite_note-MINV1-2) mentions a page from the "Slovak Ministry of Interion", but the link is outdated and returns a 404. Could you look around that website and see if you can find it?

slavkoja:
> The wikipedia page for ["Slovakian Names"](https://en.wikipedia.org/wiki/Slovak_name#cite_note-MINV1-2) mentions a page from the "Slovak Ministry of Interion", but the link is outdated and returns a 404. Could you look around that website and see if you can find it?

I found two links, https://www.minv.sk/?tlacove-spravy-8&sprava=sofia-a-jakub-su-najcastejsie-mena-pre-deti-narodene-v-roku-2016 and https://www.minv.sk/?tlacove-spravy-8&sprava=najcastejsim-menom-ktorym-rodicia-pomenovali-svoje-dieta-v-roku-2013-je-jakub but i consider both as partially useful only, as people from 2013 and 2016 are now 8 and 5 years old...

Understand i properly, that you want to i prove names, which i provided? I am slowly lose the interest to merge it, as prove source for generate **fake** data is useless, as they have not be real, they have **look as real** only. While i understand, that you have no reason to believe my data, i spent time to create it and want to share them with others (as correcting existing data is always more simple as create them from scratch). If you consider my data as bad, be concrete, what and why is bad and do not dehonest my work, only because you cannot to decide correctness.

If you want another proving the source for **fake** data, i am not willing do it and be free to close this.

fcurella:
Our policy is to always look for an authoritative source for datasets of names. In some cases there simply aren't any available, and so we do occasionally allow non-authoritative sources, with a comment explaining how the list has been compiled.

We have this policy for a practical reason: different people have a different opinion on how much real the data should look like, and what a 'correct' dataset should be.

It may look good enough or correct to person A, but not to person B. And B will open a PR to "fix it" with their own version. Then person C comes along, and disagrees with this version. And so on.

But by using an external, more authoritative source of truth (Eg a governmental office or census), we are able to point to that source and essentially delegate all the subjectivity by saying "this is where we took the data. Have them fix it, or find a better source".

You will note that not all providers have a source in their dataset. This is mostly for historical reasons. We didn't always had this policy. We had to put it in place because we used to have a lot of PRs that would just keeping changing names over and over.

To be even more clear, we require to use  an authoritative source _if_ it is available. I'm not refusing this PR, I'm asking if you could look around for a list that you didn't had to compile yourself.

slavkoja:
Your response do not corresponds to project description:

> Faker is a Python package that generates fake data for you. Whether you need to bootstrap your database, create good-looking XML documents, fill-in your persistence to stress test it, or anonymize data taken from a production service, Faker is for you.

Between lines in your response i see, that it is not for providing fake data, but for providing real data for fake activities. No thanks, i do not want to contribute to it.

PR #1443, opened 2021-05-14
Fix/master/aba routing

### LINKED ISSUE #1437: Adding ABA Routing Number to Bank Provider
* Faker version:8.1.2

Is there any interest in adding ABA routing numbers to the Bank provider? https://en.wikipedia.org/wiki/ABA_routing_transit_number

This would be helpful for my use case and have code to generate but wanted to check before putting more effort in.

mstellon:
### What does this changes

Added aba() function to bank provided to generate valid ABA routing numbers. https://en.wikipedia.org/wiki/ABA_routing_transit_number

Fixes #1437

### DIFF
--- a/faker/providers/bank/__init__.py
+++ b/faker/providers/bank/__init__.py
@@ -1,7 +1,7 @@
-from math import ceil
 import re
 import string
 
+from math import ceil
 from string import ascii_uppercase
 
 from .. import BaseProvider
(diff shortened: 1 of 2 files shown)


fcurella (changes requested):
Could you fix the linting issues?

faker/providers/bank/__init__.py
```diff
@@ -28,6 +29,20 @@ class Provider(BaseProvider):
...
+        aba = f"{fed_num:02}{rand}"
+
+        #calculate check digit
+        d = [int(n) for n in aba]
+        chk_digit = 3*(d[0] + d[3] + d[6]) + 7*(d[1] + d[4] + d[7]) + d[2] + d[5] 
+        chk_digit = ceil(chk_digit/10)*10 - chk_digit
+
+        return f"{aba}{chk_digit}"
+
+
```
fcurella, line 45: Can you remove those two blank lines?

faker/providers/bank/__init__.py
```diff
@@ -1,3 +1,4 @@
+from math import ceil
```
fcurella, line 1: we use `isort` to sort imports. Can you move this import down to Line 5? (there should be a blank line between the `import ...` block and the `from ... import ...` block)

mstellon:
Fixed linting issues. Thanks!

fcurella:
Thank you!

PR #1461, opened 2021-06-11
Fix #1460: left_digits ignored if min_value given

### LINKED ISSUE #1460: pydecimal left_digits ignored if min_value or max_value is specified
`max_value` should be set to `10 ** (left_digits -1) - epsilon` before this:

https://github.com/joke2k/faker/blob/d9f4b00b9134e6dfbb09cc1caa81c912b79c3c7c/faker/providers/python/__init__.py#L92-L102

Use cases for using both include:

- `min_value=0`, since `positive=True` disallows `0` (a bug in itself IMO, but that's an age old debate!)
- `min_value` at all actually? As in, 4 left digits, but no less than 42 in value, for example.
- a `max_value` that has a semantically different reason for existing, so it's convenient to specify in addition to `left_digits` [^]

Work around is to specify a `max_value` (per above) instead of `left_digits` if `min_value` or `max_value` are needed too.

(I will have a PR for this shortly.)

[^] - e.g. `left_digits` could be a database requirement (`NUMERIC(left + right, right)`), but `max_value` something to do with the logic the fake is for.

OJFord:
### What does this changes

Allows `left_digits` and (one or both of) `min_value` & `max_value` to work together.

### What was wrong

Specifying `min_value` or `max_value` would cause those values (even if `max_value is None`) to be used for generating the random int for the left of d.p. in pyfloat and pydecimals, ignoring `left_digits`. (#1460)

### How this fixes it

This sets `max_value` to the maximum representable within `left_digits`, if unspecified. If both are specified, we validate that they're consistent (i.e. the given `max_value` must fit within the given `left_digits`).

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -97,8 +97,12 @@ class Provider(BaseProvider):
         sign = ''
         if (min_value is not None) or (max_value is not None):
             # Make sure left_digits still respected
-            if max_value is None and left_digits is not None:
-                max_value = 10 ** left_digits  # minus smallest representable, adjusted later
+            if left_digits is not None:
+                if max_value is None:
+                    max_value = 10 ** left_digits  # minus smallest representable, adjusted later
+                if min_value is None:
+                    min_value = -(10 ** left_digits)  # plus smallest representable, adjusted later
+
             if max_value is not None and max_value < 0:
                 max_value += 1  # as the random_int will be generated up to max_value - 1
             if min_value is not None and min_value < 0:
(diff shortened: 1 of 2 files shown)


OJFord:
Ah I should have created the PR before pushing the fix commit, to get the tests to run on both. I did verify myself that it fails though.

fcurella (approved):

faker/providers/python/__init__.py
```diff
@@ -69,10 +70,13 @@ def pyfloat(self, left_digits=None, right_digits=None, positive=False,
         if positive and min_value is not None and min_value <= 0:
             raise ValueError(
                 'Cannot combine positive=True with negative or zero min_value')
+        if left_digits is not None and max_value is not None and math.ceil(math.log10(max_value)) > left_digits:
```
OJFord, line 73: I just realised this doesn't allow `max_value=0` (log will raise an exception) - and in that case, or actually in any case, this should be checking the larger absolute value of min or max, not just max_value.

I'll fix that (and a test, since it's clearly not covered) later.

OJFord: Fixed

OJFord:
Thanks!

Issue #1445, opened 2021-05-19
Locale fr_FR : siren() and siret() methods generate invalid values


cestlacata:
* Faker version: 8.1.2
* OS: Windows 10

For locale fr_FR siren() and siret() methods generate incorrect values.
https://faker.readthedocs.io/en/master/locales/fr_FR.html#faker.providers.company.fr_FR.Provider.siren
https://faker.readthedocs.io/en/master/locales/fr_FR.html#faker.providers.company.fr_FR.Provider.siret

A valid SIREN has 9 digits = 8 digits + 1 Luhn Key
A valid SIRET has 14 digits = 9 digits (SIREN) + 4 digits + 1 Luhn Key

### Expected behavior

This a a valid solution :

```python
class SirenSiretProvider(BaseProvider):
    def siren(self):
        siren = self.numerify("########")
        return siren + str(calculate_luhn(siren))

    def siret(self):
        siret = self.siren() + self.numerify("####")
        return siret + str(calculate_luhn(siret))
```

### Actual behavior

siren() and siret() only generate random numbers.

fcurella:
Thank you for the report @cestlacata ! Do you mind submitting a Pull Request?

cestlacata:
> Thank you for the report @cestlacata ! Do you mind submitting a Pull Request?

No sorry

fcurella:
No worries. I don't have time to work on it, so I'm adding the 'good first issue' label for  increasedvisibility. Maybe we'll find someone available to fix it.

babastienne:
I dettected the same bug today.
I found a way to generate valid SIRET and SIREN numbers.

I will submit a pull request soon to fix this bug.

babastienne:
The pull request https://github.com/joke2k/faker/pull/1465 fix the problem reported in this issue.

daamien:
New attempt at fixing this

https://github.com/joke2k/faker/pull/1773

daamien:
Thanks !

PR #1465, opened 2021-06-12
Fix siret and siren generation for fr_FR locale company information

babastienne:
### What does this changes

Fix SIREN and SIRET generation for company fr_FR.

The twoo fields ned to respect the luhn algorythm to be valid.

### What was wrong

The two fields where generated randomly and therefore were not valid.

### How this fixes it

Integrate the luhn control number at the end of each field.

More information about this fix on the related issue #1445

### DIFF
--- a/faker/providers/company/fr_FR/__init__.py
+++ b/faker/providers/company/fr_FR/__init__.py
@@ -1,3 +1,5 @@
+from faker.utils.checksums import calculate_luhn
+
 from .. import Provider as CompanyProvider
 
 
@@ -52,8 +54,6 @@ class Provider(CompanyProvider):
         'en toute sécurité')
 
     company_suffixes = ('SA', 'S.A.', 'SARL', 'S.A.R.L.', 'S.A.S.', 'et Fils')
-
-    siren_format = "### ### ###"
 
     def catch_phrase_noun(self):
         """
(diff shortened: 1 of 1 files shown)


faker/providers/company/fr_FR/__init__.py
```diff
@@ -110,21 +110,28 @@ def _is_catch_phrase_valid(self, catch_phrase):
...
+        return f'{code[:3]} {code[3:6]} {code[6:]}{luhn_checksum}'
 
     def siret(self, max_sequential_digits=2):
         """
         Generates a siret number (14 digits).
         It is in fact the result of the concatenation of a siren number (9 digits),
         a sequential number (4 digits) and a control number (1 digit) concatenation.
         If $max_sequential_digits is invalid, it is set to 2.
+
+        The siret number is formated as '### ### ### #####'.
```
fcurella, line 126:
```suggestion
        The siret number is formatted as '### ### ### #####'.
```

fcurella (changes requested):
Thank you! Could you add a test that checks if the checksum is valid?

gp4git:
Hi,
I need this fix.
Is it possible to release it ?

Regards

stefan6419846:
The author @babastienne apparently never answered back on the request from @fcurella to add a corresponding test and to fix the typo. Feel free to submit an improved version of this PR yourself for future inclusion into a released package version.

gp4git:
Hi Stefan,
unfortunetaly, i'm not a python developer but I can confirm that this current PR fixed the bug.
I tested it across a python script (populate.py) that uses python3 faker which is released 
with a PostgreSQL extension called "postgresql_anonymizer".
So, for me, except for the typo, this current PR solves the problem.

Regards

stefan6419846:
This is no current PR, but has been closed about half a year ago, while being created more than a year ago.

Unfortunately, the bug which has been fixed here is not represented by a test, especially that the checksum actually is correct, as pointed out by Flavio above. Adding this should not be a big deal when looking at the existing tests if one is familiar enough with SIRET and SIREN numbers, which I am not.

daamien:
As far as i can tell, the test would be 

```python
from faker import Faker
from faker.utils.checksums import luhn_checksum
luhn_checksum(Faker('fr_FR').siret().replace(' ','')) == 0
```

PR #1467, opened 2021-06-18
Avoid collision for `date` in datetime provider (#1130)

### LINKED ISSUE #1130: Error in Provider date_time for method fake.date() colliding with faker.date_time.date()
* Faker version: 4.02
* OS: MAC, LINUX

The documenation for the date_time provider for the function 

`fake.date(pattern='%Y-%m-%d', end_datetime=None)` 

 Returning 

> Exception has occurred: TypeError
> function missing required argument 'year' (pos 1)
>   File "/Users/test/test-py/faker_test.py", line 15, in <module>
>     date = fake.date(pattern='%Y-%m-%d', end_datetime=None)

Which I think is clashing with the signature for:

```
class datetime(real_datetime):
    def strftime(self, fmt):
        return strftime(self, fmt)

    def combine(self, date, time):
        return datetime(date.year, date.month, date.day,
                        time.hour, time.minute, time.microsecond,
                        time.tzinfo)

    def date(self):        <===============================
        return date(self.year, self.month, self.day)

```

### Steps to reproduce

```
from faker import Faker
from faker.providers import ( date_time)

fake = Faker()
fake.add_provider(date_time)
date = fake.date(pattern='%Y-%m-%d', end_datetime=None)
```

### Expected behavior

Should produce a random string
# '2000-07-13'

### Actual behavior

> Exception has occurred: TypeError
> function missing required argument 'year' (pos 1)
>   File "/Users/test/test-py/faker_test.py", line 15, in <module>
>     date = fake.date(pattern='%Y-%m-%d', end_datetime=None)

mattshin:
Rename datetime.date -> dtdate to not overlap with faker.date()

### What does this changes

rename datetime.date import and usages

### What was wrong

trying to access the date function from a faker instance would use the datetime.date class instead

Fixes #1130

Alternatively, i can submit a PR that makes the provider date function private (`_date`) to explicitly note that this should not be externally used. That would have been a significant breaking change so i went down this route instead.

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -1,7 +1,9 @@
 import re
 
 from calendar import timegm
-from datetime import MAXYEAR, date, datetime, timedelta
+from datetime import MAXYEAR
+from datetime import date as dtdate
+from datetime import datetime, timedelta
 
 from dateutil import relativedelta
 from dateutil.tz import gettz, tzlocal, tzutc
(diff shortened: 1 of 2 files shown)


fcurella:
Thank you so much!

PR #1475, opened 2021-07-01
Skip UnsupportedFeature's on the command line

### LINKED ISSUE #1474: faker command line fails with "faker.exceptions.UnsupportedFeature: `image` requires the `Pillow` python library."
* Faker version: 8.8.2
* OS: Ubuntu

It looks like the image faker added in #1254 / 8.5.0 broke the faker command line?

is the fix to catch `UnsupportedFeature` in cli.py, and maybe print a warning?

### Steps to reproduce

1. Install faker in a virtualenv
1. run in the terminal `faker`

### Expected behavior

cli should output fake data

### Actual behavior

Fails with exception:

faker.exceptions.UnsupportedFeature: `image` requires the `Pillow` python library.

therefromhere:
### What does this changes

Catches UnsupportedFeature exception when running the `faker` command without specifying a fake if Pillow isn't installed.

### What was wrong

UnsupportedFeature was raised by the `image` faker if Pillow wasn't installed (see #1474)

### How this fixes it

Calls `get_formatters`, building up a list of excluded fakers if UnsupportedFeature is raised.

Fixes #1474

### DIFF
--- a/faker/cli.py
+++ b/faker/cli.py
@@ -7,7 +7,7 @@ import sys
 from pathlib import Path
 from typing import Any, Dict, List, Optional, TextIO
 
-from faker import VERSION, Faker, documentor
+from faker import VERSION, Faker, documentor, exceptions
 from faker.config import AVAILABLE_LOCALES, DEFAULT_LOCALE, META_PROVIDERS_MODULES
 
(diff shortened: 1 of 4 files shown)


fcurella:
Thank you!

PR #1476, opened 2021-07-02
fix #1471

### LINKED ISSUE #1471: fake email data does not have '@' char when using 'fake.csv()'
* Faker version: 8.8.2
* OS: Windows

fake email data does not have `@` char when using `fake.csv()`

### Steps to reproduce
```
Faker.seed(4321)
fake.csv(
    header=('email',),
    data_columns=('{{email}}',),
    num_rows=1,
    include_row_ids=False,
)
```
### Expected behavior
```
'"email"\r\n"milesphilip@yahoo.com"\r\n'
```

### Actual behavior
the `2` should be `@``
```
'"email"\r\n"milesphilip2yahoo.com"\r\n'
```

___

FYI The issue seems to be specific to `fake.csv()` since the ffg. :
```
Faker.seed(4321)
fake.email()
```
works as expected
```
'milesphilip@yahoo.com'
```

lucas860529:
### What does this changes

get rid of bothify in pystr_format

### What was wrong

numerify which is in bothify replaces "@" to a digit or empty string and make email addresses wrong 

### How this fixes it

Each formatter already calls numerify, so there is no need to call bothify at the end. 
If it doesn't work, please let me know.

Fixes #1471

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -48,7 +48,7 @@ class Provider(BaseProvider):
             )
 
     def pystr_format(self, string_format='?#-###{{random_int}}{{random_letter}}', letters=string.ascii_letters):
-        return self.bothify(self.generator.parse(string_format), letters=letters)
+        return self.lexify(self.generator.parse(string_format), letters=letters)
 
     def pyfloat(self, left_digits=None, right_digits=None, positive=False,
                 min_value=None, max_value=None):


faker/providers/python/__init__.py
```diff
@@ -48,7 +48,7 @@ def pystr(self, min_chars=None, max_chars=20):
             )
 
     def pystr_format(self, string_format='?#-###{{random_int}}{{random_letter}}', letters=string.ascii_letters):
-        return self.bothify(self.generator.parse(string_format), letters=letters)
+        return self.generator.parse(string_format)
```
fcurella, line 51: this way won't we loose the option of passing a custom alphabet with `letters=`? Do we have any provider using this feature?

PR #1477, opened 2021-07-05
Add UAE phone numbers [ar_AE]

sshishov:
### What does this changes

Add UAE cellphone, telephone, service and toll numbers

### What was wrong

There were no localized phone numbers for UAE region

### How this fixes it

UAE localization is added for `phone_number` provider

---
NOTE: I am using `f-strings` in the generation of full regex for `phone_number` test. Please check if it is okay, or should the `format` be used instead.


fcurella (approved):
we support python 3.6 and up, so f-strings are ok 👍🏼

fcurella:
Thank you!

PR #1478, opened 2021-07-07
Update __init__.py

iii235:
Added a new provider

### What does this changes

Added a provider that can return md5 hashes from a .txt list, allowing you to create more realistic hashes.


fcurella (changes requested):
I would rather add an optional `values` argument to the existing `md5` method. Creating methods for every possible use case is not going to be very maintainable.

Issue #1484, opened 2021-07-22
Return value of `random_letters` is not always a list


svisser:
* Faker version: 8.10.1
* OS: Ubuntu 20.04

The return type of `random_letters` is not always a list as indicated in the docstring:

https://github.com/joke2k/faker/blob/2dac486e6d3b5f018feb524f6fa19917ec10299e/faker/providers/__init__.py#L194-L195

### Steps to reproduce

```pycon
>>> from faker import Faker
>>> f = Faker()
>>> f.random_letters(length=0)
[]
>>> f.random_letters(length=1)
('Y',)
>>> f.random_letters(length=2)
['o', 'K']
>>> f.random_letters(length=3)
['L', 'k', 'M']
```

### Expected behavior

`type(f.random_letters(length=n))` should be list when `n=1`

### Actual behavior

`type(f.random_letters(length=n))` has type tuple when `n=1`

fcurella:
Thank you for the report @svisser ! Do you have time to submit a Pull Request?

svisser:
Thanks for merging in the PR :ok_hand: I'm happy for this issue to be closed but I'll leave it open in case there is anything related to docs / changelog / release management that is left to do here.

PR #1485, opened 2021-07-22
Return list instead of tuple in choices_distribution

svisser:
### What does this changes

This is a proposed fix for https://github.com/joke2k/faker/issues/1484.

It appears this line has come up in the past: https://github.com/joke2k/faker/issues/1367#issuecomment-758345350 and https://github.com/joke2k/faker/pull/1368#pullrequestreview-566519393.

### What was wrong

We should not return a list here as that type is not the same as what `random.choices` (below it) would return (assuming the default implementation in the standard library). The `else` part below it also uses a `choices` list to return at the end.

### How this fixes it

This adds a test case for `random_letters` in general and it specifically asserts that the return value is a `list`, as mentioned in the docstring of `random_letters()`.

### DIFF
--- a/faker/utils/distribution.py
+++ b/faker/utils/distribution.py
@@ -61,7 +61,7 @@ def choices_distribution(
 
     if hasattr(random, 'choices'):
         if length == 1 and p is None:
-            return (random.choice(a),)
+            return [random.choice(a)]
         else:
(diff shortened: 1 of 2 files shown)


fcurella:
Thank you!

PR #1487, opened 2021-07-26
it_IT: concerning surname may cause hassle when showing faker results

### LINKED ISSUE #1486: it_IT: concerning surname may cause hassle when showing faker results
Removing 'Mussolini' from the list of italian surnames to avoid any concerns when results are exposed to public.

This is a similar situation to PR #379 and it's main intention is to make it safe for developers to use the results of this library in a publicly facing context.

lordgordon:
Fixes  #1486


fcurella (approved):

fcurella:
Thank you!

PR #1488, opened 2021-07-26
Fix spelling in ar_AE and ar_PS phone number providers

svisser:
### What does this changes

Fix spelling of attributes in phone number providers.

### What was wrong

Spelling was not correct.

### How this fixes it

This fixes the spelling but this may be a breaking change in case someone relies on these attributes.


fcurella:
Thank you!

PR #1491, opened 2021-07-31
add missing diacritics in French last names

laowantong:
### What does this changes

Add some missing diacritics, e.g., Herve -> Hervé, Francois -> François, Etienne -> Étienne

### What was wrong

All diacritics are missing from the French last names. First names seem ok, though.

### How this fixes it

Gracefully.


fcurella:
Thank you!

PR #1493, opened 2021-08-09
Add ru_RU internet provider tests

pishchalnikov:
### What does this changes

Adds ru_RU internet provider tests


fcurella:
Thank you!

PR #1497, opened 2021-08-20
Added currency formats in pt_BR

leogregianin:
### What does this changes
Added currency formats in pt_BR

### What was wrong
Did not exist

### How this fixes it
Added new folder in currency/pt_BR and new file __init__.py and added tests for this.


fcurella:
Thank you! <3

PR #1498, opened 2021-08-20
Added datetime formats in pt_BR

leogregianin:
What does this changes
Added date time formats in pt_BR

What was wrong
Did not exist

How this fixes it
Added new folder in date_time/pt_BR and new file init.py and added 2 tests for this.


fcurella:
Thank you!

PR #1499, opened 2021-08-20
Added non-geographic phone numbers and added services phones formats in pt_BR

leogregianin:
What does this changes
 - Added non-geographic phone numbers ([source](https://pt.wikipedia.org/wiki/N%C3%BAmeros_de_telefone_no_Brasil#N%C3%A3o_Geogr%C3%A1ficos))
 - Added services phones formats ([source](https://pt.wikipedia.org/wiki/N%C3%BAmeros_de_telefone_no_Brasil#C%C3%B3digo_de_Acesso_a_Servi%C3%A7os_de_Utilidade_P%C3%BAblica_e_de_Apoio_ao_STFC))

What was wrong
 - Non-geographic phone number did not exist in valid pt_BR formats and services phones formats did not exist.

How this fixes it
 - Added non-geographic phone numbers in formats and change regex tests for this.
 - Added a new method called _`service_phone_number`_ in providers and added _`test_service_phone`_ test case for this.

### DIFF
--- a/faker/providers/phone_number/pt_BR/__init__.py
+++ b/faker/providers/phone_number/pt_BR/__init__.py
@@ -75,6 +75,14 @@ class Provider(PhoneNumberProvider):
         '71 ####-####',
         '81 ####-####',
         '84 ####-####',
+        '0300 ### ####',
+        '0500 ### ####',
+        '0800 ### ####',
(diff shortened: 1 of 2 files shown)


fcurella:
Thank you!

PR #1508, opened 2021-09-09
Remove period/fullstop from en_GB prefixes

stephengruppetta:
### What does this changes

Removes periods/fullstops from titles/prefixes in en_GB as these are not used in British English, unlike American English

### What was wrong

Titles includes period/fullstop:
Mr.
Dr.
etc

### How this fixes it

Fullstops removed

Fixes #...


fcurella:
Thank you!

Issue #1482, opened 2021-07-12
BrokenPipeError when interupting cli output


therefromhere:
* Faker version: 8.10.1
* OS: Ubuntu 20.04

Brief summary of the issue goes here.

Stopping the faker cli produces a BrokenPipeError, which is a bit untidy.

I think this could be fixed by catching the BrokenPipeError as per https://docs.python.org/3/library/signal.html#note-on-sigpipe (via https://stackoverflow.com/a/58517082/8331 ) ?

### Steps to reproduce

In terminal

`faker | head`

### Expected behavior

Command should cleanly output and shut down

### Actual behavior

Output is:

```
$ faker | head

### faker.providers.address

	fake.address()                                                                                                                           # PO Box 4
	                                                                                                                                           Grahaming 924
	                                                                                                                                           2
	fake.bothify(text='## ??', letters='abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ')                                               # 38 iv
	fake.building_number()                                                                                                                   # 9
	fake.city()                                                                                                                              # Allenneath
	fake.city_prefix()                                                                                                                       # Mount
Traceback (most recent call last):
  File "/home/johnc/.pyenv/versions/3.9.4/bin/faker", line 8, in <module>
    sys.exit(execute_from_command_line())
  File "/home/johnc/.pyenv/versions/3.9.4/lib/python3.9/site-packages/faker/cli.py", line 271, in execute_from_command_line
    command.execute()
  File "/home/johnc/.pyenv/versions/3.9.4/lib/python3.9/site-packages/faker/cli.py", line 248, in execute
    print_doc(arguments.fake,
  File "/home/johnc/.pyenv/versions/3.9.4/lib/python3.9/site-packages/faker/cli.py", line 101, in print_doc
    print_provider(doc, provider, fakers, output=output)
  File "/home/johnc/.pyenv/versions/3.9.4/lib/python3.9/site-packages/faker/cli.py", line 47, in print_provider
    print(f"\t{signature:<{margin}}{separator} {line[i * remains:(i + 1) * remains]}", file=output)
BrokenPipeError: [Errno 32] Broken pipe
```

fcurella:
Thank you for reporting this @therefromhere ! Do you have time to work on a Pull Request?

david-xander:
- Faker version 8.12.1
- Mac OS X

[...]

PR #1513, opened 2021-09-17
Solving broken pipe error when interrupting cli output

david-xander:
### What does this changes

- When using faker through the cli with no provider given now faker outputs this useful text: 'Please use a provider or type '?' for help'. 
- It also solves the broken pipe error issue reported in issue #1482. 
- And it actually prints the promised help message shown in the first change.
- A test case for the cli has been added in order to keep faker from failing when no provider given (ie: typing just 'faker', instead of 'faker name').

### What was wrong

- There was no broken pipe except catch in case something failed during execution.
- Using faker with no params produced a weird string, that was reported with the broken pipe error in issue #1482 
- There were no help message.

### How this fixes it

- Takes into account the received arguments in case none given.
- It try catches the broken pip exception, in case something goes wrong executing faker.
- It displays a message when no argument(no provider and/or options) is given.
- It actually print a help message according to the previous point, when you type '?'.
- The the parameters is being accessed as a public property and it's initialization with the argparser definition has been changed to an specific method of the Command class, in order to be able to access the parameters defined without needing to call the execute() method and also to be able to use the print_help() method of the argparse module.

### DIFF
--- a/tests/cli/test_cli.py
+++ b/tests/cli/test_cli.py
@@ -1,0 +1,19 @@
+import pytest
+import unittest
+
+from importlib import import_module
+from pathlib import Path
+
+from faker.cli import execute_from_command_line
+
+
+class TestingCli(unittest.TestCase):
+    @pytest.fixture(autouse=True)
+    def capsys(self, capsys):
+        self.capsys = capsys
+
+    def test_using_no_command(self):
+        with pytest.raises(SystemExit) as pytest_wrapped_e:
+            execute_from_command_line(['faker'])
+
+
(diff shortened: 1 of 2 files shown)


fcurella (changes requested):
We'll also need to update the [section in the README](https://github.com/joke2k/faker#command-line-usage)

tests/cli/test_cli.py
```diff
@@ -0,0 +1,19 @@
...
+class TestingCli(unittest.TestCase):
+    @pytest.fixture(autouse=True)
+    def capsys(self, capsys):
+        self.capsys = capsys
+
+    def test_using_no_command(self):
+        with pytest.raises(SystemExit) as pytest_wrapped_e:
+            execute_from_command_line(['faker'])
+
+
```
fcurella, line 19: can you remove this extra line?

fcurella:
closing for inactivity

Issue #1517, opened 2021-09-22
Add DateTimeProvider for Bangladesh(bangla)


jabertuhin:
I want to add DateTimeProvider for Bengali week and month names.

fcurella:
Thank you interest, @jabertuhin !

You can use any of the existing DateTimeProvider as a starting point. For example, you could copy the [Indian provider](https://github.com/joke2k/faker/blob/master/faker/providers/date_time/hi_IN/__init__.py) into `faker/providers/date_time/bn_BD/__init__.py` and modify it from there.

PR #1518, opened 2021-09-23
Task/jabertuhin/1517/add datetime provider for bd

jabertuhin:
### What does this changes

This pull-request addresses issue #1517 . I have created new `bn_BD` package under date_time package.

### What was wrong

N/A

### How this fixes it

N/A

Fixes #...

N/A


faker/providers/date_time/bn_BD/__init__.py
```diff
@@ -0,0 +1,37 @@
+from .. import Provider as DateTimeProvider
+
+
+class Provider(DateTimeProvider):
```
fcurella, line 4: There's some extra space at the end of this line that's making the linter fail. Could you remove it?
```suggestion
class Provider(DateTimeProvider):
```

fcurella (changes requested):

faker/providers/date_time/bn_BD/__init__.py
jabertuhin: Thanks for the feedback. I have updated my pull-request.

fcurella:
Thank you!

PR #1521, opened 2021-09-28
added bg_BG color provider and tests

iradonov:
### What does this changes

Creates a bg_BG color provider and tests.


fcurella:
Thank you!

PR #1524, opened 2021-10-01
Update README to not use Python 2.7 prefixes

limedaniel:
### What does this changes

Remove the `u` prefix from the sample output.  As 2.7 doesn't appear to be supported (anymore?) I figure this must have been missed in the update.

### What was wrong

The docs included output that didn't happen when actually using the program.

### How this fixes it

It's a documentation fix.


fcurella:
Thank you!

PR #1526, opened 2021-10-02
Changes default email generator to use safe domain names

### LINKED ISSUE #1523: Default to Safe Email Addresses
By default `faker.email()` will generate email addresses that are potentially valid, especially considering `@gmail.com`, `@yahoo.com` and other free providers are included in the data. 

This can be problematic considering there are a few potential outcomes that can come of this that can affect the developers using `faker` or the people who have unwittingly had their emails generated as part of a test suite.

While there is the argument that developers should be configuring mailers in test environments to not send outbound, using the in-built `ascii_safe_email()` or provide the optional `domain` argument, I also think that `faker` would benefit from having the `email()` provider use the safe domains by default and require developers to enable unsafe email address generation.

### Expected behavior

Here was something I very quickly put together, to demonstrate what I'd like to see happen, with how it could be used by a developer:

```python
@lowercase
def email(self, unsafe=False, domain=None):
    if domain:
      email = f'{self.user_name()}@{domain}'
    if unsafe:
      pattern = self.random_element(self.email_formats)
      email = "".join(self.generator.parse(pattern).split(" "))
    else:
      domain = self.random_element(self.safe_domain_name)
      email = f'{self.user_name()}@{domain}'
    return email
```

Which when used by the developer would look something like:
```
fake = Faker()

# safe by default
print(fake.email())

# enable unsafe mode
print(fake.email(unsafe=True))

# use a domain
print(fake.email(domain="mycooldomain.com"))
```

```
>>> justinramos@domain.com
>>> justinramos@jackson-cuevas.net
>>> justinramos@mycooldomain.com
```

### Actual behavior

Currently, the default behaviour is for `faker` to generate email addresses that may be valid. 

I'd be more than happy to open a PR to make these changes if you'd like to pursue this approach!

Thanks!

errbufferoverfl:
### What does this changes

This PR changes the default behaviour of the `email()` provider by using the `safe_domain_names` list to seed new email addresses rather than the `free_email_domains` and `domain_name` lists.

### What was wrong

By default `email()` will generate email addresses that are potentially valid, especially considering `@gmail.com`, `@yahoo.com` and other free providers are included in the data.

This can be problematic considering there are a few potential outcomes that can come of this that can affect the developers using faker or the people who have unwittingly had their emails generated as part of a test suite.

### How this fixes it

This PR introduces a new `safe` optional argument that by default is set to `True`. 

The `email()` function now checks for this new optional argument -- if set to `True` an email will be generated in a similar way an email is generated for a provided domain -- by generating a username and then setting a random domain from the `safe_domain_names` list.

If `safe` is set to `False` the original generation flow is followed.

Fixes #1523

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -119,9 +119,11 @@ class Provider(BaseProvider):
         return string
 
     @lowercase
-    def email(self, domain=None):
+    def email(self, safe=True, domain=None):
         if domain:
             email = f'{self.user_name()}@{domain}'
+        elif safe:
+            email = f'{self.user_name()}@{self.safe_domain_name}'
         else:
             pattern = self.random_element(self.email_formats)
             email = "".join(self.generator.parse(pattern).split(" "))
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -123,7 +123,7 @@ class Provider(BaseProvider):
         if domain:
             email = f'{self.user_name()}@{domain}'
         elif safe:
-            email = f'{self.user_name()}@{self.safe_domain_name}'
+            email = f'{self.user_name()}@{self.safe_domain_name()}'
         else:
             pattern = self.random_element(self.email_formats)
             email = "".join(self.generator.parse(pattern).split(" "))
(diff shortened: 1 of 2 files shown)


fcurella (approved):

fcurella:
Thank you so much!

PR #1528, opened 2021-10-02
Add date_time provider for nl_NL

MarcelRobeer:
### What does this changes

Added a date_time provider for Dutch (`nl_NL`). 

### What was wrong

`faker` dit not include days of the week and month names for the `nl_NL` locale.

### How this fixes it
- Added an `nl_NL` folder in `faker/providers/date_time/`
- Added an `__init__.py` file in the `nl_NL` folder, with the same format as `faker/providers/date_time/de_DE/__init__.py`
- Changed the days of the week to Dutch and the month names to Dutch (my native language). Unlike German, these are not uppercased as Dutch does not uppercase days of the week and months of the year (see e.g. https://www.babbel.com/en/magazine/a-guide-to-how-to-write-the-date-in-dutch).
- Added the `TestNlNl` test case to `tests/providers/test_date_time.py`, with the same format as other date_time test cases


fcurella:
Thank you!

PR #1529, opened 2021-10-04
Fix bug in pydecimal when returning max/min

wilbertom:
When pydecimal returns the max or min values, it returns the exact
value provided by the user. The problem is that these arguments are
usually provided as float/int literals in tests, making the return
value not decimals sometimes.

### What does this changes

This patch makes sure we return a Decimal when returning the min or max.

### What was wrong

We were returning ints/floats if the user provided them as arguments.


wilbertom:
Seems like a problem in CI and not my changes, ignoring.

fcurella:
@joke2k can you remove appveyor? I don't think we need it anymore

fcurella:
Thank you so much! <3

PR #1531, opened 2021-10-06
Add `date_time` provider for `el_GR`

nikosmichas:
### What does this changes

Add date_time provider for Greek language

### What was wrong

There was no provider for days and months in Greek language

### How this fixes it

Implemented the provider


fcurella:
Thank you!

PR #1535, opened 2021-10-07
Added currency provider for nl_NL

MarcelRobeer:
### What does this changes

Added a currency provider for Dutch (`nl_NL`). 

### What was wrong

`faker` dit not include the correct price format (Euro sign before value) for the `nl_NL` locale.

### How this fixes it
- Added an `nl_NL` folder in `faker/providers/currency/`
- Added an `__init__.py` file in the `nl_NL` folder, with the same format as `faker/providers/date_time/fr_FR/__init__.py`
- Changed the format to place the Euro sign before the value, and not add a breaking space between the Euro sign and value


fcurella:
Thank you!

PR #1542, opened 2021-10-10
Dynamic provider

omri374:
### What does this change do

This change introduces a dynamic provider which allows users to create providers on the fly given a set of values. These values are passed to the init function and not hardcoded, therefore they could be used if the user has a list of elements coming from files or other sources of data.

Example code:
```python
from faker import Faker
from faker.providers import DynamicProvider

medical_professions_provider = DynamicProvider(
             provider_name="medical_profession",
             elements=["dr.", "doctor", "nurse", "surgeon", "clerk"], #could also be loaded from file
)

fake = Faker()
fake.add_provider(medical_professions_provider)
fake.medical_profession()
```
Output:
```
 "dr."
```

This change will allow users to customize the elements Faker samples from, which is especially useful when generating synthetic data for ML model training, where the number of elements could be very large, or is gathered from some existing data source.

If you think this PR is valuable, I would be happy to add documentation as well.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -260,11 +260,14 @@ How to create a Provider
     fake.foo()
     # 'bar'
 
+
 How to create a Dynamic Provider
 ------------------------
+
 Dynamic providers can read elements from an external source.
 
 .. code:: python
+
     from faker import Faker
     from faker.providers import DynamicProvider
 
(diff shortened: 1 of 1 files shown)


omri374:
Hi @fcurella what do you think about this PR? Worth adding into Faker?

omri374:
@fcurella fixed linting issues.

omri374:
fixed isort

PR #1551, opened 2021-10-13
Improve randomnes of pydecimal for different scenarios.

### LINKED ISSUE #1549: pydecimal is not working really well when use max_value with low number
* Faker version: 9.2.0
* OS: python:3.8.8-slim (Docker) 

pydecimal is not working really well when use max_value with low number and use positive option or min_value with a close value to max_value. When you use this combination you usually get the max_value all the time. Because you always have set `max_random_digits =100` https://github.com/joke2k/faker/blob/2c857e1fb64d0797bf352995815dbca56de23b8f/faker/providers/python/__init__.py#L178

The same behavior if you use min_value option with negative number with a close max_value. It always get max number o min number

### Steps to reproduce

 `fake.pydecimal(positive=True,  max_value=200)`

or `fake.pydecimal(min_value=0,  max_value=200)`

### Expected behavior

You like have a random decimal between 0 and 200 

### Actual behavior

You always get Decimal('200') because this line of code https://github.com/joke2k/faker/blob/2c857e1fb64d0797bf352995815dbca56de23b8f/faker/providers/python/__init__.py#L205

To solve this you can use same function like pyfloat inside your code https://github.com/joke2k/faker/blob/2c857e1fb64d0797bf352995815dbca56de23b8f/faker/providers/python/__init__.py#L79-L81

antonioIrizar:
### What does this changes
Improve randomnes of pydecimal for different scenarios.

### What was wrong

1. It usually go to max_value or min_value when it has a low numbers.
2. It always happen when you set a min_value for really big numbers. They need bigger than 10*100"

### How this fixes it
Redo the code that calculates it

Fixes #1549

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -176,20 +176,10 @@ class Provider(BaseProvider):
 
         # if either left or right digits are not specified we randomly choose a length
         max_random_digits = 100
-        minimum_left_digits = len(str(min_value)) if min_value is not None else 1
-        if left_digits is None and right_digits is None:
-            right_digits = self.random_int(1, max_random_digits)
-            left_digits = self.random_int(minimum_left_digits, max_random_digits)
-        if left_digits is not None and right_digits is None:
-            right_digits = self.random_int(1, max_random_digits)
-        if left_digits is None and right_digits is not None:
-            left_digits = self.random_int(minimum_left_digits, max_random_digits)
-
-        left_number = ''.join([str(self.random_digit()) for i in range(0, left_digits)]) or '0'
-        if right_digits is not None:
-            right_number = ''.join([str(self.random_digit()) for i in range(0, right_digits)])
-        else:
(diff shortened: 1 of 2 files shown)


faker/providers/python/__init__.py
```diff
@@ -198,6 +188,26 @@ def pydecimal(self, left_digits=None, right_digits=None, positive=False,
         else:
             sign = '+' if positive else self.random_element(('+', '-'))
 
+        if sign == '+':
+            if max_value is not None:
+                left_number = str(self.random_int(max(min_value or 0, 0), max_value))
+            else:
+                min_left_digits = math.ceil(math.log10(max(min_value or 1, 1)))
+                left_digits = left_digits or self.random_int(min_left_digits, max_left_random_digits)
```
ndrwkim, line 196: I believe this has inadvertently caused pydecimal to not work as intended when `left_digits=0`.

PR #1553, opened 2021-10-14
Animal Provider & Automotive/VIN

Gumdish:
### What does this changes

Add animal provider. Name, scientific name and tests.

Add to automotive provider to generate VIN.


fcurella:
Thank you @Gumdish !

I'm not sure an animal provider is a use-case common enough to be included in `Faker`. If you can re-package it as an external package, I'll be happy to include it in our [list of community providers](https://github.com/joke2k/faker/blob/master/docs/communityproviders.rst).

PR #1536, opened 2021-10-07
Typing [Fixed #1489!]

MarcelRobeer:
### What does this change?
Added type annotation hints to all functions/classes in `faker`!

### What was wrong

`faker` did not include type annotations (issue #1489).

### How this fixes it
- Added automatic type annotations with `monkeytype` (https://monkeytype.readthedocs.io/en/latest/) for all files in which it did not throw errors.
- Manual linter fixes.
- Manual fixes of missing type annotations or unknown type annotations (e.g. type `None`). Did a lot of code analysis myself, but would also like to credit @nicarl for his earlier work on type annotations (e.g. in `documentory.py` and `generator.py`).
- `mypy` compliant (`mypy -m faker`). **Also added `mypy -m faker` test to `tox.ini`!**
- `flake8` compliant (`flake8 faker tests`)
- `isort` compliant (`python3 -m isort --check-only --diff .`)

PR #1557, opened 2021-10-20
Add typing-extensions as dependency and pin mypy version

nicarl:
### What does this changes

The PR adds typing-exntesions as a dependency in `setup.py` and pins the version of mypy to 0.910.

### What was wrong

@obieler raised the issue that `from typing_extensions import Literal` does not work on python 3.6 and 3.7 without installing `typing-extensions` (https://github.com/joke2k/faker/pull/1536).
Furthermore the mypy version was not pinned. It is advised to pin the mypy version to avoid that new typing errors appear due to changes in mypy, which would make the CI and errors harder to reproduce (https://sethmlarson.dev/blog/2021-10-18/tests-arent-enough-case-study-after-adding-types-to-urllib3).

### How this fixes it

typing-extensions is added as a dependency and solves this issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,6 +68,6 @@ setup(
     install_requires=[
         "python-dateutil>=2.4",
         "text-unidecode==1.3",
-        "mypy==0.910",
+        "typing-extensions>=3.10.0.2",
     ],
 )


fcurella:
In `install_requires`, I'd rather require just `typing-extensions` than the whole `mypy`.

Q: if we do that, does it make  the `--install-types` option in `mypy --install-types` redundant?

nicarl:
> In `install_requires`, I'd rather require just `typing-extensions` than the whole `mypy`.
> 
> Q: if we do that, does it make the `--install-types` option in `mypy --install-types` redundant?

Valid point concerning the dependency, better to add just `typing-extensions`.
We still require the `--install-types` flag, as some libraries do not contain types themselves but stubs exists. They need to be installed from https://github.com/python/typeshed. For our case `python-dateutil` has only stubs which need to be installed.

(details can be found here: https://mypy.readthedocs.io/en/stable/installed_packages.html#using-installed-packages)

fcurella:
Thank you!

PR #1564, opened 2021-11-02
Fix min/max param order in `random_int` call in `pydecimal`

richarddewit:
### What does this changes

Fixes the `min_value`/`max_value` param order in the `random_int` call in `pydecimal`.

### What was wrong

_Problem introduced in #1551_

After updating my tests failed all of a sudden with the message:

```
ValueError: empty range for randrange() (1000, 101, -899)
```
I use `pydecimal` in my factories:

```python
factory.Faker("pydecimal", right_digits=2, min_value=1, max_value=1000)
```
and I noticed something off in the stacktrace:

```python
---8<---
../.virtualenvs/myproject/lib/python3.9/site-packages/factory/faker.py:48: in evaluate
    return subfaker.format(self.provider, **extra)
../.virtualenvs/myproject/lib/python3.9/site-packages/faker/generator.py:79: in format
../.virtualenvs/myproject/lib/python3.9/site-packages/faker/providers/python/__init__.py:200: in pydecimal
    left_number = str(self.random_int(max(max_value or 0, 0), abs(min_value)))
#                                          ^                       ^
../.virtualenvs/myproject/lib/python3.9/site-packages/faker/providers/__init__.py:321: in random_int
    return self.generator.random.randrange(min, max + 1, step)
```

### How this fixes it

Swap the `min_value` and `max_value` params.

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -197,7 +197,7 @@ class Provider(BaseProvider):
                 left_number = "".join([str(self.random_digit()) for i in range(0, left_digits)]) or "0"
         else:
             if min_value is not None:
-                left_number = str(self.random_int(max(min_value or 0, 0), abs(max_value)))
+                left_number = str(self.random_int(min=max(min_value or 0, 0), max=abs(max_value)))
             else:
                 min_left_digits = math.ceil(math.log10(abs(min(max_value or 1, 1))))
                 left_digits = left_digits or self.random_int(min_left_digits, max_left_random_digits)
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -197,7 +197,7 @@ class Provider(BaseProvider):
                 left_number = "".join([str(self.random_digit()) for i in range(0, left_digits)]) or "0"
         else:
             if min_value is not None:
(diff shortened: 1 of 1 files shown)


faker/providers/python/__init__.py
```diff
@@ -197,7 +197,7 @@ def pydecimal(self, left_digits=None, right_digits=None, positive=False, min_val
                 left_number = "".join([str(self.random_digit()) for i in range(0, left_digits)]) or "0"
         else:
             if min_value is not None:
-                left_number = str(self.random_int(max(max_value or 0, 0), abs(min_value)))
+                left_number = str(self.random_int(max(min_value or 0, 0), abs(max_value)))
```
fcurella, line 200: to avoid further confusion, can we also use named arguments?

```suggestion
                left_number = str(self.random_int(min=max(min_value or 0, 0), max=abs(max_value)))
```

fcurella (changes requested):

fcurella:
@richarddewit [Some tests](https://github.com/joke2k/faker/runs/4170474132?check_suite_focus=true#step:5:1365) are failing. Could you take a look?

richarddewit:
I cannot seem to reproduce this with additional tests. Maybe the problem lies elsewhere. Investigating,,,

PR #1566, opened 2021-11-10
Fix pydecimal with left_digits=0 not setting the left digit to 0

ndrwkim:
Fix `pydecimal` not setting the digit left of the decimal point to 0 when given `left_digits=0`.


OJFord (approved):
Nice one.

fcurella:
Thank you <3

PR #1567, opened 2021-11-12
Relax ElementsType to admit e.g. Sets

### LINKED ISSUE #1565: `ElementsType` requires `Sequence[T]`, not just `Iterable[T]`?
The definition for `ElementsType` currently requires a sequence, or dict keys:

https://github.com/joke2k/faker/blob/7093e5ab9b5e894e04807130b75dc940c7a9d53a/faker/providers/__init__.py#L18

I can't see why it needs to be a _sequence_ though, i.e. having any order - wouldn't `Iterable[T]` suffice?

I hit this by using `random_element(elements)`, where `elements: Set[str]`. I can work around it for now as `random_element(tuple(elements))` of course.

OJFord:
### What does this changes

The type `ElementsType` to permit a broader range of arguments.

### What was wrong

`random_element` could not be called with an argument that was a set of choices, for example.

### How this fixes it

By relaxing the annotation to a `Collection`, we allow the `Sequence`, `Dict`, and `KeysView` that previously worked (this change is backwards compatible) but also allow `Set`, `MutableSet`, `ItemsView`, and `ValuesView`.

https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

The ultimate relaxation might be `Iterable[T] + Sized` - but that would only benefit something custom, nothing in `collections` today has `Iterable` & `Sized` but not `Container` (the three being `Collection`), and I haven't tested it.

Fixes #1565

### DIFF
--- a/faker/providers/__init__.py
+++ b/faker/providers/__init__.py
@@ -2,7 +2,7 @@ import re
 import string
 
 from collections import OrderedDict
-from typing import Any, Dict, KeysView, List, Optional, Sequence, TypeVar, Union
+from typing import Any, Collection, List, Optional, Sequence, TypeVar, Union
 
 from ..generator import Generator
 from ..utils.distribution import choices_distribution, choices_distribution_unique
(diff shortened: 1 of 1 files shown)


nicarl (approved):
:+1:

fcurella:
Thank you! 💚

Issue #1576, opened 2021-12-13
`future_date` doesn't seem to be using the `tzinfo` argument


alexei:
* Faker version: 9.9.0
* OS: macOS

`future_date` doesn't seem to be using the `tzinfo` argument. From the [source code](https://github.com/joke2k/faker/blob/e2ee6af810124da6f27e51942484bd7570322f7e/faker/providers/date_time/__init__.py#L2055-L2066):

```python
    def future_date(self, end_date: DateParseType = "+30d", tzinfo: Optional[TzInfo] = None) -> dtdate:
        """
        Get a Date object based on a random date between 1 day from now and a
        given date.
        Accepts date strings that can be recognized by strtotime().

        :param end_date Defaults to "+30d"
        :param tzinfo: timezone, instance of datetime.tzinfo subclass
        :example dtdate('2030-01-01')
        :return dtdate
        """
        return self.date_between(start_date="+1d", end_date=end_date)
```

This was [mentioned before](https://github.com/joke2k/faker/issues/1214), but the same author [closed the fix](https://github.com/joke2k/faker/pull/1215) without a resolution.

### Steps to reproduce

I'm not sure how to reproduce this.

### Expected behavior

The docs for both `future_date` and `past_date` mention they accept a `tzinfo` argument, but the source code reveals it's ignored. Either the argument should be dropped or it should be used.

### Actual behavior

As stated and illustrated in the code above, the `tzinfo` argument is ignored.

PR #1577, opened 2021-12-13
fix: update future, past date generators to use tzinfo

alexei:
### What does this changes

Fixes https://github.com/joke2k/faker/issues/1576

### What was wrong

Both methods accepted the timezones, but they were ignored.

### How this fixes it

I updated the relevant methods to pass the timezone argument to the underlying helpers.

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -2024,20 +2024,23 @@ class Provider(BaseProvider):
         else:
             return (datetime(1970, 1, 1, tzinfo=tzutc()) + timedelta(seconds=ts)).astimezone(tzinfo)
 
-    def date_between(self, start_date: DateParseType = "-30y", end_date: DateParseType = "today") -> dtdate:
+    def date_between(
+        self, start_date: DateParseType = "-30y", end_date: DateParseType = "today", tzinfo: Optional[TzInfo] = None
+    ) -> dtdate:
         """
         Get a Date object based on a random date between two given dates.
         Accepts date strings that can be recognized by strtotime().
 
         :param start_date Defaults to 30 years ago
         :param end_date Defaults to "today"
+        :param tzinfo: timezone, instance of datetime.tzinfo subclass
         :example Date('1999-02-02')
         :return Date
         """
 
         start_date = self._parse_date(start_date)
         end_date = self._parse_date(end_date)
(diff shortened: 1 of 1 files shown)


fcurella:
Thank you! Could you add some tests for regressions?

alexei:
> Could you add some tests for regressions?

@fcurella I'm thinking about freezing time to e.g. 17+00 and checking that `future_date() < future_date(offset=+12)` holds true. However...

I did a little digging and I'm a little confused about something and would like to clear things up before moving any further. Looking at the code I see a chain like:

```
future_date
    -> date_between
        -> date_between_dates
            -> date_time_between_dates.date
```

I found `tzinfo` still has no effect due to `date_between` converting the arguments to dates, so the tzinfo gets stripped. Simply replacing calls to `_parse_date()` with `_parse_date_time` wouldn't work either, because `_parse_date` returns dates while `_parse_date_time` returns timestamps, so to get to the actual date I'd need to first convert to `timestamp_to_datetime`. That all seems a little backwards.

However, looking at `future_datetime` I see something very different:

```
future_datetime
    -> date_time_between
```

In light of the above, I have a couple of questions:

1. `date_between` almost aliases `date_between_dates`. What is the difference between and what is the need for the two? 
2. `date_time_between` and `date_time_between_dates` seem to achieve the same thing, yet have different implementations. What is the difference? Btw, there's also `date_time` which has it's own way of achieving a date time between 1970 and "now".
3. Is there a particular reason why `_parse_date` returns `date`s and `_parse_date_time` returns `int`s?
4. Why don't the `date_*` methods simply call their `date_time_*` equivalents? Is it by design or chance? With this approach it seems everything would simply converge to a handful of methods.
5. What is your availability to refactoring this module? Or openness to having someone else do it.

fcurella:
Hi @alexei , thank you for digging into this.

Most of the mess is for historical reasons. Originally, this library was a port of https://github.com/fzaninotto/Faker, which is written in PHP. Sometimes the porting got more literal than it should have.

There is also a lot of hidden complexity when it comes to calculate dates in Windows, especially Win32, but we should have enough test coverage around that to allow us a refactor.

In short: some complexity here is plain debris, but some is there because computers :)

I'd really like it to refactor it, just be aware there might be some surprising bug along the way.

alexei:
Let's continue the conversation in https://github.com/joke2k/faker/issues/1591

PR #1584, opened 2021-12-29
Fix/master/az locale

sevdimali:
### What does this changes

1. removed unused code from the automotive provider.
2. removed currency in which currency code doesn't exist.
3. removed some jobs that are not relevant.
4. removed duplicate surnames from unisex surnames.

### What was wrong

small bugs

### How this fixes it

Already fixed

Fixes #...


fcurella:
merged in 512365aa9625fd555028a2bbce927c7be4919b91

Issue #1585, opened 2022-01-02
postal zip code not correctly related to address


FrancYescO:
When generating an address, zip codes seems are totally unrelated to the generated address/country

i'm just taking it_IT as example (maybe a difficult one...), ZIP = CAP

The zip code in the most case is 1:1 with a city, but there are some cases of 1:n (i.e. big cities have multiple zip, based on the specific street) and some rare case of n:1 (same zip for multiple cities)
here is a good feed for italy
https://github.com/matteocontrini/comuni-json/blob/master/comuni.json

this feed can also help to have a good match between city and province (actually also unrelated that will cause to have an invalid address)

i know can be a little challenging and overkill to have an address related to the street, but at least attach it the the related city, or at least to country should be a must have

fcurella:
It's always a trade-off between how realistic we want the data to be, the complexity of the code, and the ease of maintenance.

I'm also wary of using the specific feed you mention, because their license is a little murky. I don't blame the author – they did the best under the circumstances. My worry is that distro packagers will have an issue with it.

If you have time to work an Pull Request that uses a data source with a clear license (even if it's a little incomplete or outdated), I'll be happy to review it :)

FrancYescO:
for the first point, i think at least when using the full address func, we should find a way to have a realistic address, as most application are checking the `ZIP - City` relation, or at lest `ZIP - Country`..

for an italy dataset i'll try to find something with a clear license... but as the source is from a gov entity i don't think using the dataset posted above can be an issue..

fcurella:
> i don't think using the dataset posted above can be an issue..

I dont _think_ so either. The problem is in that ambiguity. What I'm supposed to tell a distro maintainer raising (legitimate) questions?

I would prefer to tradeoff accuracy and completeness for a clear license, in this case. If that's possible.

FrancYescO:
For italy dataset..
https://github.com/Samurai016/Comuni-ITA
https://comuni-ita.herokuapp.com/api/comuni
https://comuni-ita.herokuapp.com/api/province
or
https://github.com/italia/anpr/blob/master/src/_static/xlsx/tab_comuni.xlsx (no way to link to a zip code)

For a more general zip-country-city relation... i think we can find various feeds around, just need to know if is in the interest of the project to embed it in this way...
https://github.com/Zeeshanahmad4/Zip-code-of-all-countries-cities-in-the-world-CSV-TXT-SQL-DATABASE
https://github.com/midwire/free_zipcode_data

[...]

PR #1593, opened 2022-01-07
add geonames db of italian cap-city-province and full match when generating full address

FrancYescO:
also improved format for building_number_formats and secondary_address_formats

Maybe the code can be improved but ... 🤷

Fix (at least for italy) https://github.com/joke2k/faker/issues/1585

PS. i've used this snippet to "convert to dictionary" the geonames db
```
import csv
fulldict = {}
with open('IT.txt', newline = '') as games:                                                                                          
	game_reader = csv.reader(games, delimiter='\t')
	for game in game_reader:
		if game[1] not in fulldict:
			fulldict[game[1]] = [[game[2],game[6]]]
		else:
			fulldict[game[1]].append([game[2],game[6]])

print(fulldict)
```


FrancYescO:
> faker/providers/phone_number/en_PH/__init__.py:17:121: E501 line too long (122 > 120 characters)

dafaq? en_PH was jus untouched..

faker/providers/address/it_IT/__init__.py
```diff
@@ -511,11 +17616,19 @@ class Provider(AddressProvider):
...
-    secondary_address_formats = ("Appartamento ##", "Piano #")
+    address_formats = ("{{street_address}}\n{{postcode_city_province}}",)
+    secondary_address_formats = ("Appartamento @#", "Piano #")
+
+    def postcode_city_province(self) -> str:
+        cap = self.postcode()
+        rand_city_prov = self.random_element(self.cap_city_province[cap])  # type: List[str]
+        return cap + ", " + rand_city_prov[0] + " (" + rand_city_prov[1] + ")"
+
+    def city(self):
```
fcurella: can you add a type hint here too?

```suggestion
    def city(self) -> str:
```

faker/providers/address/it_IT/__init__.py
```diff
@@ -511,11 +17616,19 @@ class Provider(AddressProvider):
...
+        "{{street_name}}, {{building_number}} {{secondary_address}}",
     )
-    address_formats = ("{{street_address}}\n{{city}}, {{postcode}} {{state}} ({{state_abbr}})",)
-    secondary_address_formats = ("Appartamento ##", "Piano #")
+    address_formats = ("{{street_address}}\n{{postcode_city_province}}",)
+    secondary_address_formats = ("Appartamento @#", "Piano #")
+
+    def postcode_city_province(self) -> str:
+        cap = self.postcode()
+        rand_city_prov = self.random_element(self.cap_city_province[cap])  # type: List[str]
```
fcurella: we fully support type hints, so you dont need the py2-compatible syntax
```suggestion
        rand_city_prov: List[str] = self.random_element(self.cap_city_province[cap])
```

fcurella (changes requested):
Can you add some tests for `postcode_city_province` and `city`?

fcurella:
@FrancYescO  Don't worry about that extra commit, we're going to squash them anyway. But please let's keep the comments family-friendly :)

FrancYescO:
> Can you add some tests for `postcode_city_province` and `city`?

can you suggest a good place for the test case?

fcurella:
@FrancYescO You should add a testcase in [tests/providers/test_address.py](https://github.com/joke2k/faker/blob/master/tests/providers/test_address.py)

fcurella:
Perfetto! Grazie! <3

PR #1599, opened 2022-01-30
Added more support for da_DK

epetrovski:
Brief summary of the changes:

- Added more Danish street and city names
- Existing Danish country names were in Swedish, fixed.
- Fixed incorrect Danish street name format
- Added Danish license plate format
- Rename folder for Danish personal names so they can be found
- Rename folder for Danish phone numbers so they can be found
- Removed (0) from Danish phone number format. Not used.
- Added Danish bank account format
- Added Danish week and month names
- Added 1.000 common Danish words for text generation
- Added ~900 Danish job titles
- Added Danish colors
- Added Danish price tag format
- Added Danish company name format


faker/providers/address/da_DK/__init__.py
```diff
@@ -6,41 +6,1719 @@ class Provider(AddressProvider):
...
     street_address_formats = ("{{street_name}} {{building_number}}",)
 
     street_prefixes = (
-        "Gade",
-        "Allé",
-        "Parkere",
-        "Baggade",
-        "Gyde",
-        "Motorvej",
+        "Aberdeen",
```
fcurella, line 14: Those don't look like street prefixes. Maybe we should rename this property to `street_names`

faker/providers/job/da_DK/__init__.py
```diff
@@ -0,0 +1,962 @@
+from .. import Provider as BaseProvider
+
+
+class Provider(BaseProvider):
+    """
+    Source: https://star.dk/it/saadan-arbejder-vi-med-it-i-styrelsen/oversigt-over-stillingsbetegnelser-og-kvalifikationer/
```
fcurella, line 6: This line is making the linter fail because it's too long. Can you mark it with `NOQA`?

```suggestion
    Source: https://star.dk/it/saadan-arbejder-vi-med-it-i-styrelsen/oversigt-over-stillingsbetegnelser-og-kvalifikationer/.  # NOQA
```

fcurella (changes requested):

faker/providers/address/da_DK/__init__.py
epetrovski: I took all street names from Copenhagen that end with the suffixes "gade" (street), "vej" (road) etc. and deleted the suffix part. So new street names can be generated from prefix and suffix.  Prefixes are usually the name of a plant, person, Nordic god, city etc.. If this is not in keeping with the terminology, I can change it?

fcurella: yes, let's change the name of the property.

epetrovski: Done. But I had to change the method name to dk_street_name() as to not overwrite the street_name() method from the AddressProvider parent class.

fcurella: Why do you want to keep the logic from the parent method around? Can't you rename just the property, and the parent method would just work?

epetrovski:
@fcurella Could you help me out with this formatting error? Can't seem to make it pass.

epetrovski:
I don't get why this is failing. Other locale providers don't even have a street_prefix()?

fcurella:
I fixed it by just removeing the test for `street_prefix`. Thank you for your patience!

PR #1600, opened 2022-01-30
Fix language setting for provider

### LINKED ISSUE #1595: Incorrect language set on Provider during locale fallback
* Faker version: 11.3.0

When faker is initialised with a locale not implemented by some of the built-in providers, it falls back to the DEFAULT_LOCALE for these providers. However, the language set on the provider’s instance in `__lang__` is still the requested locale and not the locale in effect.

This is due to `Factory._find_provider_class` not returning the locale it chose, in: https://github.com/joke2k/faker/blob/001ddee39c33b6b82196fe6a5ecc131bca3b964c/faker/factory.py#L102-L104 to `Factory._get_provider_class` which then proceeds to return the locale value as it was passed in at the first place. Thus, `provider.__lang__` does not contain the found locale (as the variable name `lang_found` would suggest): https://github.com/joke2k/faker/blob/001ddee39c33b6b82196fe6a5ecc131bca3b964c/faker/factory.py#L61

### Expected behavior

`provider.__lang__` should be set to the actual language / locale being used by the provider.

### Actual behavior

`provider.__lang__` is set to the locale that was requested but is not offered by this provider.

interDist:
### What does this change

The `__lang__` attribute of a provider will always be set to the locale being used in practice, if the provider is localized. Otherwise, it will be set to `None`.

### What was wrong

Whenever the user requested a specific locale but it was not supported by the provider, the `default_locale` of the provider or the `DEFAULT_LOCALE` would be used but the internal `__lang__` attribute of the provider would nevertheless be set to the requested locale.
For example, if the user requested `es_CO` version of the `faker.providers.foobar` provider with offered locales `fr_CA`, `en_US`, `zh_TW`, the user would be handed back the `faker.providers.en_US.foobar` provider, with its `__lang__` attribute set to `es_CO`.

### How this fixes it

The `__lang__` attribute is set to the locale offered by the provider, or `None` in case the provider is not localized, instead of the locale requested by the user.

Fixes #1595.

### DIFF
--- a/faker/factory.py
+++ b/faker/factory.py
@@ -54,7 +54,7 @@ class Factory:
             if prov_name == "faker.providers":
                 continue
 
-            prov_cls, lang_found = cls._get_provider_class(prov_name, locale)
+            prov_cls, lang_found, _ = cls._find_provider_class(prov_name, locale)
             provider = prov_cls(faker)
             provider.__use_weighting__ = use_weighting
             provider.__provider__ = prov_name
@@ -64,31 +64,14 @@ class Factory:
         return faker
 
     @classmethod
-    def _get_provider_class(cls, provider: str, locale: Optional[str] = "") -> Tuple[Any, Optional[str]]:
-
-        provider_class = cls._find_provider_class(provider, locale)
-
-        if provider_class:
-            return provider_class, locale
-
-        if locale and locale != DEFAULT_LOCALE:
-            # fallback to default locale
-            provider_class = cls._find_provider_class(provider, DEFAULT_LOCALE)
-            if provider_class:
-                return provider_class, DEFAULT_LOCALE
-
-        # fallback to no locale
-        provider_class = cls._find_provider_class(provider)
-        if provider_class:
-            return provider_class, None
-
-        msg = f"Unable to find provider `{provider}` with locale `{locale}`"
-        raise ValueError(msg)
-
-    @classmethod
-    def _find_provider_class(cls, provider_path: str, locale: Optional[str] = None) -> Any:
+    def _find_provider_class(
+        cls,
+        provider_path: str,
+        locale: Optional[str] = None,
+    ) -> Tuple[Any, Optional[str], Optional[str]]:
 
         provider_module = import_module(provider_path)
+        default_locale = getattr(provider_module, "default_locale", "")
 
         if getattr(provider_module, "localized", False):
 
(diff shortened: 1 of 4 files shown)


faker/factory.py
```diff
@@ -64,31 +64,14 @@ def create(
         return faker
 
     @classmethod
-    def _get_provider_class(cls, provider: str, locale: Optional[str] = "") -> Tuple[Any, Optional[str]]:
```
interDist, line 67: completely dropped this method as only its first 3 lines are in fact used.

tests/providers/test_ssn.py
```diff
@@ -876,6 +876,33 @@ def test_pt_BR_rg(self):
                 assert re.search(r"^\d{9}$", to_test)
 
 
+class TestNlBE(unittest.TestCase):
```
interDist, line 879: moved from test_factory.py

tests/test_factory.py
```diff
@@ -270,42 +340,6 @@ def test_pyfloat_same_min_max(self):
         with pytest.raises(ValueError):
             assert fake.pyfloat(min_value=9999, max_value=9999)
 
-    def test_us_ssn_valid(self):
```
interDist, line 273: this seems to be identical to a test already implemented in test_ssn.py.

tests/test_factory.py
```diff
@@ -270,42 +340,6 @@ def test_pyfloat_same_min_max(self):
...
-        for i in range(1000):
-            ssn = provider.ssn()
-            assert len(ssn) == 11
-            assert ssn[0] != "9"
-            assert ssn[0:3] != "666"
-            assert ssn[0:3] != "000"
-            assert ssn[4:6] != "00"
-            assert ssn[7:11] != "0000"
-
-    def test_nl_BE_ssn_valid(self):
```
interDist, line 286: moved over to test_ssn.py ; not sure why this was in test_factory.py.

tests/providers/test_ssn.py
fcurella: Thank you for cleaning this up, but could you make these changes in a separate PR?

fcurella (changes requested):

tests/providers/test_ssn.py
interDist: No problem.

interDist:
What should I do about `black` failing for 4 source files? These are providers that I have not touched in this PR.

fcurella:
@interDist You can you rebase your PR against `master`, or merge `master` into your branch. That should fix it.

tests/providers/test_ssn.py
interDist: See #1606.

fcurella: I appreciate it. #1606 is merged :)

fcurella (approved):

fcurella:
Thank you for fixing this!

PR #1605, opened 2022-02-03
Use os.urandom in binary if unseeded

AlmightyYakob:
### What does this change

Updates the `binary` method, to increase performance.

### What was wrong

The existing code for the `binary` method is very slow (30-40x slower!) compared to `os.urandom`. For anything larger than a trivial amount of bytes, the list comprehension adds a very large amount of overhead.

### How this fixes it

If the instance is unseeded, `os.urandom` is used to generate random bytes in the `binary` method. Otherwise, the original method is used.

### DIFF
--- a/faker/generator.py
+++ b/faker/generator.py
@@ -17,11 +17,12 @@ class Generator:
         "arguments": {},
     }
 
+    _is_seeded = False
+
     def __init__(self, **config: Dict) -> None:
         self.providers: List["BaseProvider"] = []
         self.__config = dict(list(self.__config.items()) + list(config.items()))
         self.__random = random
-        self._is_seeded = False
 
     def add_provider(self, provider: "BaseProvider") -> None:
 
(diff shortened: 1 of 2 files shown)


AlmightyYakob:
> Could we modify [`seed`](https://github.com/joke2k/faker/blob/master/faker/generator.py?rgh-link-date=2022-02-04T15%3A58%3A03Z#L72-L73) and [`seed_instance`](https://github.com/joke2k/faker/blob/master/faker/generator.py?rgh-link-date=2022-02-04T15%3A58%3A03Z#L62-L69) to set something like`cls._is_seeded = True`, and use that property to automatically switch behavior?

I think that'd be ideal. I had a similar idea but wasn't familiar enough with the codebase to seriously propose it.

fcurella (changes requested):
You'll need to update the tests as well, since the data will change

AlmightyYakob:
@fcurella I've fixed the tests, I was initializing the `_is_seeded` var wrong. However, now calling `seed` will set this var for all faker instances, without any real way to reset it. Is this expected/desired?

fcurella (approved):

PR #1607, opened 2022-02-05
Fix postcode format for es_ES locale

### LINKED ISSUE #1596: es_ES postalcode is not generating valid codes
* Faker version: 11.3
* OS: Any

When using postcode for es_ES and using it with a field that requires a valid Postal Code, it fails sometimes.
I will assume that there is no logic with postal code generation for Spain.

### Steps to reproduce

Generate postal codes

### Expected behavior

Get a valid Spain postal code

### Actual behavior

Unexpected. Many are wrong

----

I'll dig now into the code. Let's see if I can get some more information and fix it :thinking: Do not expect much from me

mondeja:
### What does this changes

Change generation of es_ES postcodes.

### What was wrong

Postcode formats in es_ES locale is generating invalid values.

### How this fixes it

The implementation is not exhaustive. Postal codes in Spain usually are in the range from ~~3000~~ 1000 to 52100 but not all codes are used, so this implementation generates a random number between that range. It still generates invalid fakes postcodes, but at least looks like valid ones. See ["List of postal codes in Spain" (wikipedia)](https://en.wikipedia.org/wiki/List_of_postal_codes_in_Spain).

Fixes #1596.

### DIFF
--- a/faker/providers/address/es_ES/__init__.py
+++ b/faker/providers/address/es_ES/__init__.py
@@ -135,6 +135,6 @@ class Provider(AddressProvider):
         return self.random_element(self.regions)
 
     def postcode(self) -> str:
-        return str(self.generator.random.randint(3000, 52100)).zfill(5)
+        return str(self.generator.random.randint(1000, 52100)).zfill(5)
 
     autonomous_community = region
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

Issue #1519, opened 2021-09-24
Add LoremProvider for Bengali(Bangladesh)


jabertuhin:
I want to add LoremProvider for bn_BD.

### Steps to reproduce
N/A

### Expected behavior
N/A

### Actual behavior
N/A

fcurella:
Hi, sorry about the late response!

Feel free to submit a Pull Request. You can just copy one of the existing providers and make your changes.

PR #1608, opened 2022-02-05
Task/jabertuhin/1519/add loremprovider for bn bd

jabertuhin:
### What does this changes

This pull-request resolves issue #1519 . I have created `bn_BD` package under lorem provider.

### What was wrong

NA

### How this fixes it

NA

### Fixes

NA


fcurella (changes requested):
Thank you! Could you add a simple test case for the new provider?

fcurella (approved):

PR #1609, opened 2022-02-08
Closes #1570 by adding cities to swiss

### LINKED ISSUE #1570: fake.address does not work properly for 'de-CH'
* Faker version: 9.8.3             build: pyhd8ed1ab_0    from: conda-forge
* OS: Windows 10 / anaconda

    fake = Faker('de_CH')
    fake.address()

does not work properly. 
Instead of producing a city/village, it produces a Canton
### Steps to reproduce

from faker import Faker
fake = Faker('de_CH')
print(fake.address())

### Expected behavior

Output should be: 
street housenumber
zip_code village

### Actual behavior

Output  is
street housenumber
zip_code canton

Kalbra:
Adding cities to swiss for better adress formating. As described in #1570
The format changes from:
```
street housenumber
zip_code canton
```
to:
```
street housenumber
zip_code village
```


fcurella (changes requested):
Thank you! Can you add a test case for the new `city_name` method?

Kalbra:
I added a test case, hope thats the right way to do it

fcurella (approved):
Thank you!

PR #1610, opened 2022-02-08
Closes #936 Adding lorem support for german

### LINKED ISSUE #936: wrong language in faker.word() method
### Steps to reproduce

1. from faker import Faker 
1. fake = Faker('de_DE')
1. fake.word()

### Expected behavior

Generated word should be in German language

### Actual behavior

Generated word is in Latin. The same situation is with Spanish word() method.

Kalbra:
This adds lorem support for the german language. As described in #936.
It uses 500 words of this site: https://www.gut1.de/grundwortschatz/grundwortschatz-500/
Some words are changed, because of grammatical duplications.

The language is also changed for `de_AT`, because it is basiclly the same. 
A test for `de_DE` is also added.


fcurella (changes requested):
Can you add tests for the `de_AT` locale as well?

Kalbra:
I add tests for de_AT, even if it is not really necessary, cause `de_AT` is the same as `de_DE`. But does not harm, so

fcurella (changes requested):
Thank you! Can you run `make lint` to fix the formatting issues?

Kalbra:
I reformated the files by using `black --line-length 120` like discribed in the style guide.

tests/providers/test_lorem.py
```diff
@@ -4,6 +4,8 @@
 
 from faker.providers.lorem.az_AZ import Provider as AzAzLoremProvider
 from faker.providers.lorem.cs_CZ import Provider as CsCzLoremProvider
+from faker.providers.lorem.de_DE import Provider as DeDeLoremProvider
+from faker.providers.lorem.de_AT import Provider as DeAtLoremProvider
```
fcurella, line 8: these need to be sorted alphabetically. You can run `make lint` to have [`isort`](https://pycqa.github.io/isort/) fix it for you.

fcurella (changes requested):

tests/providers/test_lorem.py
Kalbra: Sorry for my bad formatting skills, but now I ran `make lint` and added the changes in [e56ac0c](https://github.com/joke2k/faker/pull/1610/commits/e56ac0c0625b83f62940ff161ca2b64045b71ad9)

fcurella (approved):
Thank you!

Issue #1538, opened 2021-10-08
randrange complains about non-integer value when calling pyfloat.


0xffea:
* Faker version: 9.2.0
* OS: Debian

randrange complains about non-integer value when calling pyfloat.

https://docs.python.org/3.10/whatsnew/3.10.html
> Non-integer arguments to random.randrange() are deprecated. The ValueError is deprecated in favor of a TypeError. >><
> (Contributed by Serhiy Storchaka and Raymond Hettinger in bpo-37319.)

### Steps to reproduce
```
import faker
fake=faker.Faker()
fake.pyfloat(min_value=0.1)
```
### Expected behavior

pyfloat returns a random float >0.1

### Actual behavior

```
Successfully installed faker-9.2.0
# python
Python 3.10.0 (default, Oct  6 2021, 00:03:11) [GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import faker
>>> fake=faker.Faker()
>>> fake.pyfloat(min_value=0.1)
/usr/local/lib/python3.10/site-packages/faker/providers/__init__.py:115: DeprecationWarning: randrange() will raise TypeError in the future
  return self.generator.random.randrange(min, max + 1, step)
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/random.py", line 303, in randrange
    istart = _index(start)
TypeError: 'float' object cannot be interpreted as an integer

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.10/site-packages/faker/providers/python/__init__.py", line 110, in pyfloat
    left_number = self._safe_random_int(
  File "/usr/local/lib/python3.10/site-packages/faker/providers/python/__init__.py", line 147, in _safe_random_int
    return self.random_int(min_value, max_value - 1)
  File "/usr/local/lib/python3.10/site-packages/faker/providers/__init__.py", line 115, in random_int
    return self.generator.random.randrange(min, max + 1, step)
  File "/usr/local/lib/python3.10/random.py", line 309, in randrange
    raise ValueError("non-integer arg 1 for randrange()")
ValueError: non-integer arg 1 for randrange()
```

ozeranskiy:
If we want `fake.pyfloat(min_value=0.1)` to return a random `float >0.1`, then conversion to int will not return the expected value.

fcurella:
[...]

PR #1614, opened 2022-02-11
Prevent "non-integer arguments to randrange()" warnings on Python 3.10

jacksmith15:
### What does this changes

From Python 3.10, non-integer arguments to `randrange()` produce a warning. This PR removes a path in which `pyfloat` can emit that warning.

### What was wrong

The existing implementation of `faker.providers.python.Provider.pyfloat` may pass float values to `random_int`, which in turn passes them directly to `randrange`.

### How this fixes it

`random_int` is already annotated to only accept integers, so I've made the fix in the `pyfloat` implementation (casting the arguments to `int` first).

An alternative would be to annotate `pyfloat` as only accepting integers, but this could be seen as a breaking change.

Related: https://github.com/joke2k/faker/issues/1538

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -155,7 +155,7 @@ class Provider(BaseProvider):
 
         return result
 
-    def _safe_random_int(self, min_value, max_value, positive):
+    def _safe_random_int(self, min_value: float, max_value: float, positive: bool) -> int:
         orig_min_value = min_value
         orig_max_value = max_value
 
@@ -169,7 +169,7 @@ class Provider(BaseProvider):
         if min_value == max_value:
             return self._safe_random_int(orig_min_value, orig_max_value, positive)
         else:
-            return self.random_int(min_value, max_value - 1)
+            return self.random_int(int(min_value), int(max_value - 1))
 
     def pyint(self, min_value: int = 0, max_value: int = 9999, step: int = 1) -> int:
         return self.generator.random_int(min_value, max_value, step=step)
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

fcurella:
@coneybeare `.random_int` expects integers:
```python
    def random_int(self, min: int = 0, max: int = 9999, step: int = 1) -> int:
        """Generate a random integer between two integers ``min`` and ``max`` inclusive
        while observing the provided ``step`` value.

        This method is functionally equivalent to randomly sampling an integer
        from the sequence ``range(min, max + 1, step)``.

        :sample: min=0, max=15
        :sample: min=0, max=15, step=3
        """
        return self.generator.random.randrange(min, max + 1, step)
```

Is there something calling it with floats?

coneybeare:
> @coneybeare `.random_int` expects integers:
> 
> ```python
>     def random_int(self, min: int = 0, max: int = 9999, step: int = 1) -> int:
>         """Generate a random integer between two integers ``min`` and ``max`` inclusive
>         while observing the provided ``step`` value.
> 
>         This method is functionally equivalent to randomly sampling an integer
>         from the sequence ``range(min, max + 1, step)``.
> 
>         :sample: min=0, max=15
>         :sample: min=0, max=15, step=3
>         """
>         return self.generator.random.randrange(min, max + 1, step)
> ```
> 
> Is there something calling it with floats?

I have found that using `pydecimal` triggers the warning with the following signature:

```
Faker().pydecimal(min_value=0.00, max_value=100.00, right_digits=2)
```

Am I using `pydecimal` incorrectly? Are the `min_value` and `max_value` supposed to be an integer when dealing with decimals?

PR #1619, opened 2022-02-19
Add faker_biology as community provider

richarda23:
### What does this change

Adds 'faker_biology' project to list of community providers

### What was wrong
Nothing was wrong, this is an addition to list of community providers
to enable easier inclusion of fake bio-data for testing purposes

### How this fixes it

For software in the life-science or biomedical domain it can be useful to use
realistic bio-data for testing. This can be awkward and time-consuming to extract from specialist databases. faker_biology aims to provide data-sets in a way that is easy to call using standard fake.xxx() method calls.

faker_biology has an open Apache2 license and is available on PyPi, and has good test coverage. I hope to expand the range of data included in future, and any suggestions or contributions are welcome.


fcurella (approved):
Thank you! <3

PR #1621, opened 2022-02-23
added lorem provider for fa-ir language code

arminshoughi:
added Lorem provider for the Persian language


fcurella (approved):
Thank you!

PR #1632, opened 2022-04-17
Fix error in currency provider

zerogonal:
### What does this change

Fixed an error in the currency provider.

### What was wrong

The currency with the code KRW should be the South Korean won, not "Western Krahn Language"


fcurella:
Nice catch! Thank you <3

PR #1639, opened 2022-04-27
Add vi-VN automotive provider

zerogonal:
Added automotive provider for vi-VN locale.


zerogonal:
@fcurella I recommend that you put all my pull requests into one big version bump.

PR #1645, opened 2022-04-28
Improved it_IT ssn provider to match real building rules

DamienPirsy:
### What does this change

Changed surname and name creation part to match the real working of an italian SSN (Codice Fiscale); results are taken from actual (male or female,  based on generated gender) first names and surnames as generated by the localized relative providers.

### What was wrong

Original letters where just random characters, now they are taken from names and surnames as they would in a real life setting. Fiscal codes should be closer to real ones now

### How this fixes it

Used native Faker localized providers as source for names and surnames


fcurella (approved):
Thank you!

Issue #1161, opened 2020-04-22
Suggestion: Include "Codename" generator


airstrike:
I'll steal the description from http://www.codenamegenerator.com/ since it's succinct and thorough:

> What is a codename? A codename, codeword or cryptonym is a name used to refer clandestinely to another name or word. Codename, codeword and cryptonym are usually interchangeable. Codenames are used to identify projects while maintaining secrecy against rival concerns. In more simple terms, a codename is a name that you agree upon and you use to hide what you are talking about. In industry, codenames are often used to designate products that are in their developement phase to protect them from competitors. In the military, codenames are frequently used to designate missions and operations.

This is not exactly the same as the single word generator that's already available (which can yield words like "such") but a narrower in scope as it often includes proper nouns, adjectives, mythological names, locations and more, and possibly a combination of these words.

Happy to kick start the effort by providing an initial list of codenames if this has any chance of being added.

fcurella:
I think I would rather have two separate `adjective` and `noun` methods, and people can combine them to make their codenames. This would give users more flexibility while keeping the implementation simple:

```python
from faker import Faker

fake = Faker()

for _ in range(100):
    print(f"{fake.adjective()} {fake.noun()}")

# "slow llama"
# "fast rocket"
```

airstrike:
That works. You'd be favoring simplicity in implementation for miss out on the proper nouns which are very commonly used for codenames IRL e.g. Barbarossa, Manhattan, Longhorn, Atlantis, etc. Though arguably the user could pull from the names and addresses generators to get (some) of those as well if desired.

crd:
This feels like a good use case for a new [Community Provider](https://faker.readthedocs.io/en/master/communityproviders.html).

dkotschessa:
@airstrike I a,m looking for contributions to make to this project.  I have recently created a community provider and was looking for another issue to work on.  Has this been initiated in any way?  It looks like a good one to work on.

airstrike:
> @airstrike I a,m looking for contributions to make to this project. I have recently created a community provider and was looking for another issue to work on. Has this been initiated in any way? It looks like a good one to work on.

Great to hear! I have not started on this at all, so I'm sorry to report I don't have anything to provide off-the-shelf to help jump start this exercise, but if there's anything I can help with, please let me know

dkotschessa:
@airstrike Meet me over in https://github.com/dkotschessa/faker_codename/issues

[...]

PR #1648, opened 2022-05-03
Added english part-of-speech functionality.

e-panourgia:
# What does this change 
Inspired by the discussion on #1161, I extended the lorem provider, adding the functionality of generating words of specific parts of speech. Currently, this has been implemented only for the case of the English language. The user is able to randomly generate the most common english **verbs, nouns, adverbs** and **adjectives**, which I found scraping the web sites that I included in `faker/providers/lorem/en_US/__init__.py`

# Expected behavior
The script:
```python
from faker import Faker

fake = Faker()
print(fake.words(4, "noun"))
print(fake.words(3, "adverb"))
print(fake.words(2, "verb"))
print(fake.word("adjective"))
```
would print something like:
```
['rope', 'evidence', 'rate', 'phase']
['virtually', 'forever', 'necessarily']
['hit', 'construct']
nice
```


fcurella (approved):
Looks good! Thank you!

PR #1649, opened 2022-05-04
Fix fr-FR datetime provider and add tests for it

zerogonal:
I have fixed the fr-FR date_time provider and added tests for it.


fcurella (approved):
Thank you!

PR #1651, opened 2022-05-04
Add random git provider

Quantum-0:
### What does this change

Add new provider.
Inspired by https://ngneat.github.io/falso/

### DIFF
--- a/faker/providers/git/__init__.py
+++ b/faker/providers/git/__init__.py
@@ -11,5 +11,5 @@ class Provider(BaseProvider):
 
     def git_commit_sha(self) -> str:
         return '%040x' % random.randrange(16**40)
-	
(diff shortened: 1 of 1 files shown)


Quantum-0:
Probably I should add tests?

fcurella:
Thank you @Quantum-0 !

I can see how this could be useful, but I don't think it's common enough to be included in the main package. Would you mind implementing as an [external provider](https://github.com/joke2k/faker#how-to-create-a-provider)? I'll be happy to include it in the [community providers page](https://faker.readthedocs.io/en/master/communityproviders.html)

Quantum-0:
Hmm.. I think I'll try to do that later. When (if) that will be done, I'll open PR to include it in the community providers page

PR #1660, opened 2022-05-11
Fix/master/1618 firefox ua nondeterministic w fixed seed

### LINKED ISSUE #1618: Firefox user agents with fixed seed are non-deterministic
* Faker version: tested on 8.14.1
* OS: MacOS

Seeded calls to `user_agent.firefox()` do not produce deterministic results, dependening on *when* the code is executed.

### Steps to reproduce

1. Run
```python
from faker import Faker
faker = Faker()
faker.seed_instance(1)
print(faker.firefox())
# Executing this on Feb 16th 2022 using faker 8.14.1 returned the following:
# 'Mozilla/5.0 (X11; Linux i686; rv:1.9.5.20) Gecko/2012-01-03 17:18:41 Firefox/3.8'
```
2. *Wait a few an unspecified amount of time* (alternatively, set your computer's date to something else far enough from current date).
3. Run step 1 again. The result is different.

### Expected behavior

The output of the code in step 1 should always be the same given the same version of faker and same seed.

### Actual behavior

The output of the code in step 1 changes depending on system time.

### Why this happens

Firefox user agents contain a segment `"Gecko/<timestamp>"`. Faker generates that timestamp by calling `date_time.date_time_between(<hardcoded start time>)`, setting the beginning of the interval but not the end, which defaults to `"now"` (i.e. `datetime.datetime.now()`), changing the interval and *sometimes* the end result .

**Bonus:** The reason you sometimes have to wait for quite long for the value of the seeded call to `firefox()` to change seems to be the underlying call to `randint(a,b)`. When seeded, that function is very insensitive to changes in the right end of the interval. For example, given a fixed seed, `randint(0,100000)` and `randint(0,111111)` return the exact same value. This likely points to the use of moduli in the RNG implementation in that function but that's a different issue and I didn't want to sidetrack myself. That is the reason why the actual bug with user agents is hard to reproduce, it happens because when working with epochs numbers can be quite large.

### Conclusion

I understand why the firefox user agent is generated the way it is, you would not want timestamps in the future or capped to an arbitrary date. But this inevitably breaks any workflow that depends on seeding to generate user agents deterministically (it broke some tests I'm working on). 

For my particular use case I can develop around this (basically by not calling `user_agent()` ever) but I can't come up with a simple solution that doesn't simply cap the timestamps of the `"Gecko"` strings. What do repo maintainers/users think?

jhbuhrman:
### What does this change

Make Firefox UA output deterministic when the same seed is used.

### What was wrong

The range from which the random dates were generated were dependent on the current date and therefore nondeterministic.

### How this fixes it

Random dates are now called with a fixed end of range, far in the future (`datetime.datetime.max`).

Fixes #1618

This PR was kindly supported by [Ordina Pythoneers](https://www.ordina.nl/vakgebieden/python/).

### DIFF
--- a/faker/providers/user_agent/__init__.py
+++ b/faker/providers/user_agent/__init__.py
@@ -1,8 +1,11 @@
 import string
 
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from .. import BaseProvider, ElementsType
+
+
+_DT_ALMOST_MAX = datetime.max - timedelta(1.0)
 
 
 class Provider(BaseProvider):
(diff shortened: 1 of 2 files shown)


fcurella (changes requested):
Thank you! could you run `make lint` on your local copy to lint the code?

jhbuhrman:
Fixed issue from `make lint`.

fcurella (approved):
Thank you!

Issue #1506, opened 2021-09-09
Feature request: length kwarg for all providers to limit the number of characters in the output


raviriley:
* Faker version: Latest
* OS:

Brief summary of the issue goes here.

### Expected behavior

For example, `faker.job(max_length=30)` would behave as normal but produce no output that is longer than 30 characters long.

omri374:
If you have a predefined list of elements, you can overwrite the existing functionality and get Faker to return only elements from your list. These can be achieved using a custom provider or a dynamic provider.

Example:

```python
from faker import Faker
from faker.providers import DynamicProvider

# provide your own list of jobs and filter out those who are longer than a value:
max_length = 8
all_elements = ["professor", "doctor", "judge", "policeman", "policewoman", "fire fighter"]
max_length_elements = [element for element in all_elements if len(element) <= max_length]

max_length_jobs_provider = DynamicProvider(provider_name="job", elements=max_length_elements)

# Add the new provider
fake.add_provider(max_length_jobs_provider)

# Call Faker
fake.job()
```
'doctor'


Doing this across the board would either be inefficient (Faker would have to keep sampling until it finds an element of size < X) or very complex (as Faker creates elements based on other elements. E.g. to return a person's name, Faker, with some probability, would sample a first name + a last name. Adding the `max_length` to each provider would result in a first and last name each having max 8 chars, but the full name having max 17 (first + space + last). This can be taken care of but would probably require a significant amount of work.

Hope this helps!

ffeldhaus:
Having the ability to restrict the length of the output for all providers is an important feature for many tests where the input also requires a specific length or where edge cases should be tested.

In my opinion all providers should have the ability to specify `max_length` and `min_length`.

I understand that this is a difficult problem and in my opinion a workaround to create samples until I have one with the right size is not a good solution to the problem. That's why I rather have a proper solution for this important problem directly in Faker than trying to create a workaround for all providers by sampling the output.

fcurella:
I'm definitely open to ideas!

PR #1662, opened 2022-05-14
Add length constraints for Person providers

MetcalfeTom:
### What does this changes

This PR works on https://github.com/joke2k/faker/issues/1506 by adding maximal and minimal length for person providers

Continuation of https://github.com/joke2k/faker/pull/1615

### DIFF
--- a/faker/providers/person/__init__.py
+++ b/faker/providers/person/__init__.py
@@ -1,5 +1,5 @@
 from collections import OrderedDict
-from typing import Optional, Tuple, Union
+from typing import Optional, Tuple, Union, Dict
 
 from .. import BaseProvider, ElementsType
 
@@ -209,13 +209,13 @@ class Provider(BaseProvider):
     def first_name(self, min_length: Optional[int] = None, max_length: Optional[int] = None) -> str:
         return self.random_element(self.first_names, min_length, max_length)
 
-    def first_name_as_list(self) -> Union[tuple, OrderedDict, dict]:
+    def first_name_as_list(self) -> Union[Tuple[str, ...], OrderedDict[str, float], Dict[str, float]]:
         return self.first_names  # type: ignore[return-value]
 
     def last_name(self, min_length: Optional[int] = None, max_length: Optional[int] = None) -> str:
         return self.random_element(self.last_names, min_length, max_length)
 
-    def last_name_as_list(self) -> Union[tuple, OrderedDict, dict]:
+    def last_name_as_list(self) -> Union[Tuple[str, ...], OrderedDict[str, float], Dict[str, float]]:
         return self.last_names  # type: ignore[return-value]
 
     def name_male(self) -> str:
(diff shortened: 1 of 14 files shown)


faker/providers/person/__init__.py
```diff
@@ -205,11 +206,17 @@ def name(self) -> str:
...
         return self.generator.parse(pattern)
 
-    def first_name(self) -> str:
-        return self.random_element(self.first_names)
+    def first_name(self, min_length: Optional[int] = None, max_length: Optional[int] = None) -> str:
+        return self.random_element(self.first_names, min_length, max_length)
 
-    def last_name(self) -> str:
-        return self.random_element(self.last_names)
+    def first_name_as_list(self) -> Union[tuple, OrderedDict, dict]:
```
fcurella, line 212: I think we can get a little bit more specific withthe type hint here:
```suggestion
    def first_name_as_list(self) -> Union[Tuple[str, ...], OrderedDict[str, float], Dict[str, float]]:
```

MetcalfeTom: For all similar methods correct?  (e.g. `suffix_female_as_list()`)

fcurella: correct

MetcalfeTom: Done

MetcalfeTom:
@fcurella can you advise me on the appveyor build error?
```
The build phase is set to "MSBuild" mode (default), but no Visual Studio project or solution files were found in the root directory. If you are not building Visual Studio project switch build mode to "Script" and provide your custom build command.
```

Also how do you want to proceed with the other providers?  Do they all need updating?  In which case, do you want me to cover them all in this PR or add them via separate PRs?

fcurella:
@MetcalfeTom you can just ignore the Appveyor build.

> how do you want to proceed with the other providers? Do they all need updating?
Yes, they will all need to be updated.

> do you want me to cover them all in this PR or add them via separate PRs?

I'd rather have separate, smaller PRs.

MetcalfeTom:
> I'd rather have separate, smaller PRs.

Can do.  How do you want to proceed?

You could open up a new branch based on `master`.  I can change the base of this PR to that and begin opening up others for the providers.  After those are complete we can merge into `master`

fcurella:
> I can change the base of this PR to that and begin opening up others for the providers. After those are complete we can merge into master

I'd rather re-fork off of `master` each time for each PR. I think this way we could reduce merge conflicts along the way.

MetcalfeTom:
@fcurella as directed, I opened up PRs based on all applicable provider types.  Please review the changes here when you have time, as it will be the first to merge

MetcalfeTom:
@fcurella none of the providers in the library satisfy only this clause:

```python
if hasattr(self, "suffixes_male") and hasattr(self, "suffixes_female"):
    suffixes = tuple(self.suffixes_male) + tuple(self.suffixes_female)  # type: ignore[attr-defined]
    return suffixes
```

therefore I can't increase the test coverage of `faker/providers/person/__init__.py`

fcurella (approved):
Thank you!

fcurella:
Sorry @MetcalfeTom , I had to revert all the length-check changes because the test suite started failing.

MetcalfeTom:
@fcurella is there somewhere I can read more about the failing tests?  Trying to understand if I can help get this code back in

fcurella:
you can look at [this job](https://github.com/joke2k/faker/actions/runs/3244082077/jobs/5319758523).

MetcalfeTom:
@fcurella I can fix this by:

1. Adding additional elements to fit the length constraints
2. Changing the test criteria (seems arbitrary)

Probably option 1 is better.  Should I open a PR?

fcurella:
Let's go with option #1

PR #1663, opened 2022-05-21
Add support for generating relative file paths

Lukas0907:
### What does this change

Add support for generating relative file paths to the file provider.

### What was wrong

`file_path()` would always return an absolute file path, i.e., one that starts with `/`.

### How this fixes it

The argument `absolute=True` is added to `file_path()`. It's `True` by default to not change the current behavior but can be set to `False` in case the user wants a relative path.


fcurella (approved):
Thank you!

PR #1664, opened 2022-05-21
refactor(package): move to pyproject.toml

### LINKED ISSUE #1636: Move to `pyproject.toml` + `setup.cfg`

tplessas:
### What does this change

* Move package metadata to pyproject.toml (PEP 621)
* Move build system definition to pyproject.toml (PEP 517, 518)
* Move unsupported options to setup.cfg
* Refactor setup.py to only pass conditional zipsafe flag
* Will close #1636 

### What was wrong

#1636 requests moving the build configuration to `pyproject.toml` in order to ease the burden of switching to a non-`setuptools` build backend in the future.

### How this fixes it

Most metadata and build options that used to be supplied to `setup()` in `setup.py` have been moved to `pyproject.toml`, depending on current `setuptools` support [1] and the PEP 621 specification [2].

The two options whose inclusion in the above file is as of yet unsupported (`metadata`, `package_data`) were moved to `setup.cfg` instead [3]. Again, the reason most metadata lives in `pyproject.toml` is due to the expectation of switching to `poetry` as outlined in #1636.

The `zip_safe` option is generated conditionally and, as such, cannot be included in any of the above files - it remains in `setup.py`.

### Comments

* The new build configuration format was tested against the old by unpacking the wheels generated by `build` and running a [diff](https://github.com/joke2k/faker/files/8747804/diff.txt) (result linked).
* Unlike `setuptools`, `poetry` does not support dynamically loading the version number from a file. Since the `VERSION` file is only used during building as far as I can tell we could just configure `bumpversion` to change the entry in `pyproject.toml` instead.
* No clue what to do about the `zip_safe` flag - is there really any case where `!(zipimport.zipimporter in pkgutil.iter_importer_modules.registry.keys())`? Don't think `poetry` supports it as an option anyway.

[1] https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html
[2] https://peps.python.org/pep-0621/
[3] https://setuptools.pypa.io/en/latest/userguide/declarative_config.html

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,48 @@
+[project]
+name = "Faker"
+dynamic = ["version", "readme"]
+description = "Faker is a Python package that generates fake data for you."
+classifiers = [
+  "Development Status :: 5 - Production/Stable",
+  "Environment :: Console",
+  "Intended Audience :: Developers",
+  "Programming Language :: Python",
+  "Programming Language :: Python :: 3",
+  "Programming Language :: Python :: 3 :: Only",
+  "Programming Language :: Python :: 3.6",
+  "Programming Language :: Python :: 3.7",
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Implementation :: PyPy",
+  "Topic :: Software Development :: Libraries :: Python Modules",
+  "Topic :: Software Development :: Testing",
+  "Topic :: Utilities",
+  "License :: OSI Approved :: MIT License",
+]
+keywords = ["faker", "fixtures", "data", "test", "mock", "generator"]
+authors = [{name = "joke2k", email = "joke2k@gmail.com"}]
+urls = {homepage = "https://github.com/joke2k/faker"}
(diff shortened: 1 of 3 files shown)


fcurella:
The dynamic `zip_safe` was introduced in https://github.com/joke2k/faker/pull/212, and I believe it may affect "frozen" installations, such as `PyInstaller` or `cx_freeze` (see https://github.com/joke2k/faker/pull/927)

tplessas:
After some cursory research it seems that PyInstaller has implemented support for `pkgutil.iter_modules` since the dynamic `zip_safe` was included (see https://github.com/pyinstaller/pyinstaller/pull/5959). This means that we should be able to at least support PyInstaller with a static `True` flag (nothing of the sort for `cx_freeze` unfortunately).

That said, the issue of `poetry` not supporting the `zip_safe` flag at all still applies (only reference to it I found was this: https://github.com/python-poetry/poetry/issues/928, pretty stale too). As such, if `faker` being zipsafe is a hard requirement, it seems that the best way forward right now is sticking with `setuptools`. Also, `poetry` really shines at dependency management/specification, which does mean that `faker`, being from what I understand exclusively dependent on the stdlib, would not really have much to gain from such a transition.

PR #1665, opened 2022-05-21
Added fake misc data for the el_GR locale.

e-panourgia:
### What does this change
I extended the ```misc``` fake data generator to include the el_GR locale. More specifically, the user can now generate fake **gemstone**, **mountain**, **plant** and **space** data in Greek.

### Expected behavior
The script
```python
from faker import Faker

fake = Faker("el_GR")
print(fake.gemstone_name())
print(fake.mountain_name())
print(fake.plant_name())
print(fake.space_object_name())
print(fake.random_object_name())
```
would print something like
```
Μαργαριτάρι
Πεντέλη
Μαρούλια
Ποσειδώνας
Αιγάλεω
```

where the method ```random_object_name()``` returns a random object that is included in one of the other categories.


fcurella:
Closed via #1674

PR #1666, opened 2022-05-30
Minor docstring changes and RUT range corrections

nstuardod:
### What does this change
- Redefine `faker.providers.ssn.es_CL.Provider.company_rut()` to return values starting from 60,000,000.
- Extract bounds to class definition.
- Reformat and improve `faker.providers.ssn.es_CL.Provider` docstring.

### What was wrong
Docstring was formatted improperly, and ranges did not match with cited sources.

### How this fixes it
See first section.

### DIFF
--- a/faker/providers/ssn/es_CL/__init__.py
+++ b/faker/providers/ssn/es_CL/__init__.py
@@ -29,33 +29,39 @@ class Provider(BaseProvider):
     A Faker provider for the Chilean VAT IDs, also known as RUTs.
 
     Sources:
-    https://es.wikipedia.org/wiki/Rol_%C3%9Anico_Tributario
-    https://presslatam.cl/2018/04/el-problema-de-la-escasez-y-stock-disponible-de-los-ruts-en-chile/
+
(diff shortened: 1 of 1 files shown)


faker/providers/ssn/es_CL/__init__.py
```diff
@@ -29,33 +29,39 @@ class Provider(BaseProvider):
...
     def company_rut(self) -> str:
         """
-        :return: a random Chilean RUT between 32.000.000 and 99.999.999
+        :return: a random Chilean RUT between 60.000.000 and 99.999.999
         """
-        return self.rut(32000000, 99999999)
+        return self.rut(self.minimum_rut_company, self.maximum_rut_company)
 
-    def rut(self, min: int = 10, max: int = 99999999) -> str:
+    def rut(self, min_value: int = minimum_rut_person, max_value: int = maximum_rut_company) -> str:
```
fcurella, line 56: For backward compatibility, can we keep the same argument names?

fcurella (changes requested):

Issue #1628, opened 2022-03-10
Support for RFC 3339


1Mark:
Add support for RFC 3339

https://stackoverflow.com/questions/522251/whats-the-difference-between-iso-8601-and-rfc-3339-date-formats

PR #1670, opened 2022-06-09
Rfc3339

tplessas:
### What does this change

* Expose `sep` and `timespec` parameters for isoformat()
* Update documentation
* Update tests

### What was wrong

Inspiration for this PR was gotten from #1628 - the RFC 3999 [1] format for date-time representation is universal in computing, therefore allowing for its full support in `faker` is a worthwhile effort.

Most of the various valid date and time representations [2] are already well-supported by `faker` through pattern-passing to the relevant producing functions. Combined date-time representations are, however, currently limited to the default `isoformat()` output.

### How this fixes it

The easiest and most maintainable way to add support for RFC 3999 (as well as all ISO 8601 timestamp profiles) would be to expose the `sep` and `timespec` parameters in `isoformat()`. This way users can just pass the specifics of the profile configuration they would like the result to be returned in. Changes are non-breaking and fully tested.

[1] https://datatracker.ietf.org/doc/html/rfc3339
[2] https://ijmacd.github.io/rfc3339-iso8601/

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -1877,8 +1877,8 @@ class Provider(BaseProvider):
         self,
         tzinfo: Optional[TzInfo] = None,
         end_datetime: Optional[DateParseType] = None,
-        sep: Optional[str] = 'T',
-        timespec: Optional[str] = 'auto'
+        sep: Optional[str] = "T",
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

PR #1671, opened 2022-06-09
Fix area codes for French phone numbers

zerbusdetroy:
What does this change
Fix a wrong area code for French phone numbers ("58 " instead of "508")

What was wrong
French phone number generation could lead to invalid phone number like "058 XXXXXX"


nicarl (approved):
Good catch! Change looks good to me

fcurella (approved):
Thank you!

PR #1672, opened 2022-06-12
feat(providers/ssn/el-GR): rewrite provider

tplessas:
### What does this change

* Fix police ID generator
* Fix VAT ID generator
* Add TIN generator
* Add SSN generator
* Add tests
* Add sources and docs

### What was wrong

Existing generators for the `el-GR` locale would provide incorrect data. Otherwise important generators were not included.

### How this fixes it

Generating functions have been rewritten to conform with the format specifications referenced to in comments. More generators have been added. No breaking changes introduced, tests included.


fcurella (approved):
Thank you!

PR #1674, opened 2022-06-20
docs(docs/communityproviders.rst): Add faker-geoscience

tplessas:
### What does this change

* Add faker-geoscience to the community provider list

### Description

Me and @e-panourgia are collaborating on a new Faker community provider for data related to the Earth sciences - we used those collected for the purposes of #1665 to jump start the effort.

We'd be really glad if our project could be included in the list of community providers - it does fulfill all the basic requirements listed and we could use the visibility to invite contributions too.


fcurella (approved):
Thank you!

PR #1675, opened 2022-06-22
Upgrade same tr_TR providers

keyiflerolsun:
> person
> company
> internet
> currency

providers updated..


fcurella (changes requested):
Thank y ou! cOuld you fix the linting issues?

fcurella (approved):
Thank you!

PR #1678, opened 2022-06-30
Add a new Education Provider

matthttam8411:
I created a new education provider to populate public school information from schools in the United States. I may add additional information later but for now I needed this so I figured I'd publish so other's could use it too.

### What does this change

Links to a new community provider I built.

### What was wrong

I needed to generate school information such as school names and districts but couldn't find an easy way to do this without populating my own faker choice list.

### How this fixes it

I built my own provider using publicly available public school data.

Fixes #...


fcurella (approved):
Thank you!

fcurella (commented):
lint

fcurella (commented):
-

PR #1682, opened 2022-07-14
fix: change format fake license_plate local_PT

augutoraminelli:
### What does this change

In 2017, Brazil joined Mercosul economic block, where vehicle license plates followed the standard among the bloc's member countries. As a result, a new board reconfiguration was adopted.

### What was wrong

As of old format, it had "LLL - NNNN" configuration plate. With the new change, the fifth element of the plate becomes a number. Ex: "LLL-NLNN".

### How this fixes it

The change basically when creating a new fake plate, the fifth element becomes a letter. Ex: "LLL-NLNN".

Fixes #...


fcurella (approved):
Thank you!

PR #1683, opened 2022-07-14
build(setup.py): Add `project_urls`

tony:
### What does this change

Adds [`project_urls`](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls) to _setup.py_

### What was wrong

When on [Faker's PyPI page](https://pypi.org/project/Faker/) it's not possible to get to changelog in a single click

### How this fixes it

Adds links which will be visible on next PyPI release

<details>

<summary>Preview</summary>

image

</details>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,12 @@ setup(
     author="joke2k",
     author_email="joke2k@gmail.com",
     url="https://github.com/joke2k/faker",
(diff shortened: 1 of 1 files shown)


fcurella (approved):
Thank you!

PR #1687, opened 2022-07-22
Fix Provider._parse_date_time to support timestamps (#1668)

### LINKED ISSUE #1668: unix_time() with start_datetime crashes with OverflowError
* Faker version: 13.12.0
* OS: Linux 5.18.1-arch1-1

I want to create two unix timestamps: One with any value and the other one with a value higher than the first one (minimal example below). Does not work but crashes with an OverflowError (Traceback below).

### Steps to reproduce

Create a minimal example:
```
from faker import Faker

fake = Faker()
Faker.seed(0)

unix_time_1 = fake.unix_time()
unix_time_2 = fake.unix_time(start_datetime=unix_time_1)
```
Execute the example and get the following traceback:

```
  File "/home/…/venv/lib/python3.10/site-packages/faker/providers/date_time/__init__.py", line 1813, in unix_time
    start_datetime = self._parse_start_datetime(start_datetime)
  File "/home/…/venv/lib/python3.10/site-packages/faker/providers/date_time/__init__.py", line 1924, in _parse_start_datetime
    return cls._parse_date_time(value)
  File "/home/…/venv/lib/python3.10/site-packages/faker/providers/date_time/__init__.py", line 1981, in _parse_date_time
    return datetime_to_timestamp(now + timedelta(value))
OverflowError: date value out of range
```

### Expected behavior

`unix_time_2` should be a unix timestamp with a value higher than `unix_time_1`.

### Actual behavior

Crashes with OverflowError.

mochi-moshi:
### What does this change

Change `Provider._parse_date_time` to treat integer values as UNIX timestamps.

### What was wrong

`Provider._parse_date_time` when taking an integer as a parameter treats it as a timedelta in days where as `unix_time()` returns the POSIX timestamp in seconds. Either the creation of the timedelta object or its addition to `datetime.now()` would cause an
OverflowError.

### How this fixes it

Value is directly returned rather than being treated as a timedelta.

Fixes #1668

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -1983,7 +1983,7 @@ class Provider(BaseProvider):
             time_params = cls._parse_date_string(value)
             return datetime_to_timestamp(now + timedelta(**time_params))  # type: ignore
         if isinstance(value, int):
-            return datetime_to_timestamp(now + timedelta(value))
+            return value
         raise ParseError(f"Invalid format for date {value!r}")
 
     @classmethod
--- a/tests/providers/test_date_time.py
+++ b/tests/providers/test_date_time.py
@@ -125,7 +125,7 @@ class TestDateTime(unittest.TestCase):
         assert timestamp > now
         delta = timedelta(days=30)
         from_delta = DatetimeProvider._parse_date_time(delta)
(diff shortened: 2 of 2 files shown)


fcurella (approved):
Thank you!

PR #1694, opened 2022-07-25
Add length constraints for Company providers

MetcalfeTom:
### What does this change

This PR works on https://github.com/joke2k/faker/issues/1506 by adding maximal and minimal length for company providers

Continuation of https://github.com/joke2k/faker/pull/1662


fcurella (approved):
Thank you!

normic:
@fcurella so this is merged but reverted? 
Does this mean that there is currently no solution to this issue?

What is the recommended way to use Faker with company name (or several other providers) on a database field which of course has a max_length?
I even couldn't find a declaration what length I should expect from Faker.

stefan6419846:
See https://github.com/joke2k/faker/pull/1662#issuecomment-1326267609 for the latest related discussion.

PR #1698, opened 2022-08-11
Adds warning for bad usage of date_time_this_year

### LINKED ISSUE #1697: Faker allows date_time_this_year with unreasonable values
* Faker version: "13.3.1"
* OS: Linux

I just set up a faker like this (using factory boy):

```
key_expires = Faker("date_time_this_year", before_now=False)
```

Because `after_now` defaults to `False`, this sets both `after_now` and `before_now` to `False.  

As a result, `date_time_this_year` winds up just picking the time right now, because that's the closest thing that makes sense.

Would it be better to just have an assertion to prevent people from having both set to `False`? 

I could probably do a PR to land this. 

The only big consideration I can think of is that it might break some people that have it in the bad config, so we might need some good release notes.

mlissner:
### What does this change

Adds a warning when date_time_this_year has False for both before and after now.

### What was wrong

It was too easy to have a bad configuration, particularly since the defaults for before_now and after_now are not the same.

### How this fixes it

Adds a warning

Fixes #1697


mlissner:
Looks like tests probably need a kick or something, but I think this is pretty much ready.

mlissner:
Fixed typing error (I hope), rebased and merged.

stefan6419846:
Shouldn't we have a test for this as well?

mlissner:
> Shouldn't we have a test for this as well?

Just to test that a warning was emitted? Sounds kind of unnecessary to me, but I guess I could try to figure it out if the maintainers here wanted me to.

mlissner:
Looks like adding a test isn't too hard. Let me know if this is needed:

```
with self.assertWarns(Warning):
    do_something()
```

fcurella:
@mlissner yes, a test like that is what we need.

mlissner:
Sorry, I got a lot busier since I offered to do this. I don't have the time to set up tests and get them working just to assert that a warning is raised. I'd still love for this to be merged though, since the code itself makes the faker more useful, but I'll understand if nobody wants to take this over the line. I appreciate the great package though.

stefan6419846:
I would tackle this, but this would mean to open a different PR for this. If the *Allow edits from maintainers* option is enabled for this PR, it should be much easier for the owner/collaborators of this repository to do so.

stefan6419846:
@fcurella This is an option @mlissner has to set for this PR in the right column with the PR metadata, usually below the participants (example setting: https://user-images.githubusercontent.com/96178532/201717118-7849026a-2786-4a17-bb98-38c7b7445beb.png), as this allows all maintainers of *faker* to write to the `mlissner:1697-warn-about-unreasonable-values` branch.

mlissner:
Thanks @stefan6419846. I've got that box checked, for what it's worth, so I guess somebody can run with this if they want to.

PR #1700, opened 2022-08-15
Feature/provider/locale/bn bd/1685

saanpritom:
### What does this change

New **bn_BD** locale support is added for the following providers:

- address
- automotive
- bank
- color
- company
- currency
- geo
- internet
- job
- person
- phone_number
- ssn

Existing **bn_BD** locale support is updated for the following provider:

- date_time

**Reports on the development process**

- No new test cases have been written as the code coverage of the newly added codes are 100%

- All the TOX test cases have been passed except **FLAKE8**. Some errors are showing which do not belong to my codes

- The following two environments of TOX test couldn't be tested. As currently I do not have an active subscription of Travis CI service. So, requesting you to do the tests and let me know if it fails
- The providers which are not mentioned above are unnecessary. Because in Bangladesh people use en_US locale info for those providers
- Most of the cases just the data needed to translate but for some providers specially **address** and **person** some new methods are added to generate pattern
- Sensitive data has the source included

### What was wrong

N/A

### How this fixes it

Fixes #...

N/A


fcurella (approved):
Thank you!

Issue #1705, opened 2022-08-25
Wrong postcode generation


SanderFtn:
* Faker version: faker==13.14.0
* OS: Ubuntu 22.04.1 LTS

Using generate postcode() does not work properly

### Steps to reproduce

```
from faker import Faker
fake = Faker('uk_UA')
Faker.seed(0)
for _ in range(20):
    print(fake.postcode())  
``` 

### Expected behavior

The postcode consists of five digits (01000-99999)

### Actual behavior

The postcode consist of six digits

```
for _ in range(20):
    print(fake.postcode())    
    
92662
42146
32179
105932
58144
22000
88653
02653
64266
46874
73775
104332
03592
26608
85107
18246
103864
08727
107715
99332
```

PR #1709, opened 2022-08-30
Fixed uk_UA 5 digit postcode generation issue

ajinkyac03:
### What does this change
Changed the random no generation logic in postcode generation for uk_UA.

### What was wrong

- postcode() does not work properly for uk_UA
- The postcode consists of five digits (01000-99999)
- Actual behaviour was generating 6 digit postcodes sometimes.

### How this fixes it
- Postcode is generated between (01000-99999).

Fixes https://github.com/joke2k/faker/issues/1705


fcurella (approved):
Thank you!

PR #1710, opened 2022-08-30
Relax minimum version of `typing-extensions`

### LINKED ISSUE #1708: Lower bound for typing_extensions?
* Faker version: 14.1.0
* OS: OpenSUSE Leap 15.3

For old Python versions (before Python 3.8), *faker* requires the *typing_extensions* with at least version 3.10.0.2. Due to some other dependencies inside my project, I am required to use version 3.7.4.3.

### Steps to reproduce

1. `pip install faker`

### Expected behavior

*faker* works with `typing_extensions==3.7.4.3` as well, to allow better integration with larger systems which are stuck at such older versions for some reasons.

### Actual behavior

*faker* declares `typing_extensions>=3.10.0.2` (#1557), although I could see no issue when using it with `typing_extensions==3.7.4.3`.

stefan6419846:
### What does this change

The minimum version of `typing-extensions` is being relaxed, as reported in #1708.

### What was wrong

Version 3.10 is too new for some setups involving multiple packages, while there has been not apparent reason for this requirement.

### How this fixes it

Reduce the required version to 3.7.4.3 which did not cause issues in my tests.

Fixes #1708.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,6 +72,6 @@ setup(
     python_requires=">=3.6",
     install_requires=[
         "python-dateutil>=2.4",
-        "typing-extensions>=3.10.0.2;python_version<'3.8'",
+        "typing-extensions>=3.7.4.3;python_version<'3.8'",
     ],
 )


fcurella (approved):
Thank you!

PR #1711, opened 2022-08-30
Add support for prefix and suffix to `pystr` provider

### LINKED ISSUE #1707: Allow prefix and suffix values for pystr provider
* Faker version: 14.1.0
* OS: OpenSUSE Leap 15.3

I am trying to replace [`factory.fuzzy.FuzzyText`](https://factoryboy.readthedocs.io/en/stable/fuzzy.html#fuzzytext) with the *faker* equivalent, as the *fuzzy* implementations are going to be discontinued in the future. Unfortunately, *prefix* and *suffix* values for `Faker.pystr` are not supported at the moment.

### Expected behavior

An optional prefix and suffix are supported for the *pystr* provider.

stefan6419846:
### What does this change

Adds support for providing a prefix and a suffix for the `pystr` provider.

### What was wrong

The `pystr` provider did not allow to use prefixes and suffixes, which is required for migrating from `factory_boy.fuzzy.FuzzyText`.

### How this fixes it

Adds two new parameters for it.

Additionally, the `pystr` tests have been fixed itself, as they have not been running lately (missing `test_` prefix for the method) and the assertion for `min_chars == max_chars` was broken as well.

Fixes #1707.

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -51,20 +51,22 @@ class Provider(BaseProvider):
     def pybool(self) -> bool:
         return self.random_int(0, 1) == 1
 
-    def pystr(self, min_chars: Optional[int] = None, max_chars: int = 20) -> str:
+    def pystr(self, min_chars: Optional[int] = None, max_chars: int = 20, prefix: str = "", suffix: str = "") -> str:
         """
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

PR #1713, opened 2022-08-30
Improve pydecimal provider with min and max values

leo-naeka:
### What does this change

Improve numbers generation through the pydecimal provider, especially with `min_value` and `max_value`.

### What was wrong

The pydecimal provider was not accepting floats or decimals for min and max value boundaries.  
Hence it was not possible to cover some cases, like scoped percent values (consider any N-decimal number between 73,5% and 98%, so with `min_value=D("0.735")` and `max_value = D("0.98")`).

### How this fixes it

Support decimals and floats for `min_value` and `max_value`. Improve right number generation accordingly.

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -176,22 +176,23 @@ class Provider(BaseProvider):
 
     def pydecimal(
         self,
-        left_digits=None,
-        right_digits=None,
-        positive=False,
-        min_value=None,
-        max_value=None,
-    ):
+        left_digits: Optional[int] = None,
+        right_digits: Optional[int] = None,
+        positive: bool = False,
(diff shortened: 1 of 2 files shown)


fcurella (commented):
Thank you! Would you mind adding type hints to the signature?

leo-naeka:
> Thank you! Would you mind adding type hints to the signature?

For sure!

leo-naeka:
Just realized that I missed some bits… Fixed now.

The remaining mypy errors concern `abs()` return inferred as `object` (see https://github.com/python/mypy/issues/8601).

Tried wrapping it but not really convinced. Your thoughts on this?

fcurella:
ok, lwt's just not add the type hint for now. Sorry for the extra work

PR #1717, opened 2022-09-13
added banks for en-IN

ChiragMangukia:
### What does this change

Added banks for en-IN locale.

### What was wrong

No utility was implemented for en-IN banks

### How this fixes it

Added a list of Indian banks.


ChiragMangukia:
Will request new with some changes

PR #1719, opened 2022-09-18
Added banks for en-IN

ChiragMangukia:
### What does this change

Added banks for en-IN locales. It includes a list of Indian origin banks.

### What was wrong

Support for en-IN banks was not implemented

### How this fixes it

Added a new list of banks

Fixes #...


fcurella (changes requested):
Thank you! Could you add a test case for the new provider?

ChiragMangukia:
Added the test in #1724

fcurella:
superseded by https://github.com/joke2k/faker/pull/1724

PR #1729, opened 2022-09-28
Add enum provider

NazarioJL:
### What does this change

Adds an enum provider to allow users to generate values based on enums. 

#### Usage:
```python
from enum import Enum

from faker import Faker


class Color(Enum):
    RED = 1
    GREEN = 2
    BLUE = 3


fake = Faker()
fake.enum(Color)
# One of [Color.RED, Color.GREEN, Color.BLUE]
```

### DIFF
--- a/faker/documentor.py
+++ b/faker/documentor.py
@@ -1,7 +1,7 @@
 import inspect
 import warnings
-from enum import Enum, auto
 
+from enum import Enum, auto
 from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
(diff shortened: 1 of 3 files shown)


fcurella (changes requested):
Could you run `make lint` to fix the formatting issues?

NazarioJL:
@fcurella need me to do anything else?

fcurella:
No, thank you. I'll take it form here and fix the mypy issues :)

PR #1733, opened 2022-10-12
Fix ImportError on python <3.7.2

### LINKED ISSUE #1732: Version 15.1.0 not working in python 3.7
* Faker version: 15.1.0
* OS: Windows
* Python: 3.7

When importing Faker 15.1.0, it raises an exception:

In this line:
faker/providers/__init__.py
from typing import Any, Collection, List, Optional, OrderedDict, Sequence, TypeVar, Union
ImportError: cannot import name 'OrderedDict' from 'typing'

### Steps to reproduce

1. Install python 3.7
1. from typing import OrderedDict
1. it should raise an exception.

### Expected behavior

I would expect faker to be imported and working as expected.

### Actual behavior

ImportError: cannot import name 'OrderedDict' from 'typing'

matthewhughes934:
788350230154b6e2c151850f9e9c0b7f1abccbca added an import of `typing.OrderedDict` but this was added in Python 3.7.2[1] resulting in an import failure for python versions before this.

[1] https://docs.python.org/3/library/typing.html#typing.OrderedDict

### What does this change

Fix ImportError on python <3.7.2

### What was wrong

There is an `ImporError` in python versions before this, see #1732

### How this fixes it

Import from `typing_extensions` if not available in `typing`

Fixes #1732

### DIFF
--- a/faker/providers/__init__.py
+++ b/faker/providers/__init__.py
@@ -1,7 +1,12 @@
 import re
 import string
 
-from typing import Any, Collection, List, Optional, OrderedDict, Sequence, TypeVar, Union
+from typing import Any, Collection, List, Optional, Sequence, TypeVar, Union
+
+try:
+    from typing import OrderedDict
+except ImportError:
+    from typing_extensions import OrderedDict
 
 from ..generator import Generator
 from ..utils.distribution import choices_distribution, choices_distribution_unique
--- a/faker/providers/__init__.py
+++ b/faker/providers/__init__.py
@@ -1,14 +1,11 @@
 import re
 import string
 
+from collections import OrderedDict
 from typing import Any, Collection, List, Optional, Sequence, TypeVar, Union
 
-try:
-    from typing import OrderedDict
-except ImportError:
-    from typing_extensions import OrderedDict
-
 from ..generator import Generator
+from ..typing import OrderedDictType
 from ..utils.distribution import choices_distribution, choices_distribution_unique
 
 _re_hash = re.compile(r"#")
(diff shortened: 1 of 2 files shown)


matthewhughes934:
I guess accepting this change depends whether you want to support all versions of Python 3.7 or just the latest one? Note that there's nothing in this change preventing a regression in the future, that could be done with something like adding explicit CI tests on 3.7.0 i.e. comes down to the previous question.

fcurella (changes requested):
would you mind moving the `try .. catch` to `faker.compat` and import `OrderedDict from there?

I would also use `from ... import OrderedDict as OrderedDictType` to avoid confusion with `collections.OrderedDict`

matthewhughes934:
> would you mind moving the `try .. catch` to `faker.compat` and import `OrderedDict from there?

Just to check I'm not missing something: `fake.compat` needs to be created

> I would also use `from ... import OrderedDict as OrderedDictType` to avoid confusion with `collections.OrderedDict`

:+1:

fcurella:
@matthewhughes934 Soory, my bad. I meant [`faker.typing`](https://github.com/joke2k/faker/blob/master/faker/typing.py)

matthewhughes934:
> I would also use `from ... import OrderedDict as OrderedDictType` to avoid confusion with `collections.OrderedDict`

:thinking: is this distinction necessary, I think `typing.OrderedDict` (and `typing_extensions.OrderedDict`) are both just aliases of `collections.OrderedDict`, e.g. on Python 3.7.1

```python
>>> from typing_extensions import OrderedDict
>>> OrderedDict([("foo", 1)])
OrderedDict([('foo', 1)])
>>> OrderedDict([("foo", 1)])["foo"]
1
```

But happy to separate them if you want to maintain the distinction, I've done that with [903a2e9](https://github.com/joke2k/faker/pull/1733/commits/903a2e9e290dfb61292339c65aa82b84b73ee544)

fcurella:
> is this distinction necessary

It's not necessary, but it would save my sanity if I have to debug issues on Python < 3.9 :)

fcurella (approved):

PR #1734, opened 2022-10-12
fix-in-progress

karlduggan:
### What does this change

Added new method random_float and added it inside the pyfloat() method

### What was wrong

Min and max float data where being parsed to a method that only takes integers "random_int(int(min_value), int(max_value - 1))"
which contributed to the error

### How this fixes it

I added a new method "random_float(self, min: float = 0.0, max: float = 9999.0)" that takes in float values and uses the random.uniform method to return a random value between two given floats.


faker/providers/python/__init__.py
```diff
@@ -193,7 +194,7 @@ def _safe_random_int(self, min_value: float, max_value: float, positive: bool) -
             max_value = int(max_value - 1)
             if max_value < min_value:
                 max_value += 1
-            return self.random_int(min_value, max_value)
+            return self.random_float(min_value, max_value)
```
fcurella, line 197: `_safe_random_int` is designed to return an `int`, not a `float`. If you want to fix `pyfloat`, it will have to be fix at that level, or we'll have to refactor this method to be more generic.

fcurella (changes requested):

PR #1735, opened 2022-10-13
Fix type variable inside the ElementsType

nils-van-zuijlen:
### What does this change

Add type variables at every use point of the ElementsType type variable.

### What was wrong

As the inner type variables were not specified, they were assumed to be `<nothing>` in function declarations, and `Any` in variable declarations.

This is a problem in the following example:
```python
class DjangoChoiceProvider(providers.BaseProvider):
    def choice_django(self, choices: Collection[Tuple[str, str]]) -> str:
        """Random choice in django choice element"""
        return self.random_element(choices)[0]
```
Here, mypy complains because `self.random_element(choices)` is assumed to return `<nothing>`, and the subscription is assumed by it to fail.

### How this fixes it

The TypeVar is now bound in the function definition, and takes its value from the passed argument's types.

### DIFF
--- a/faker/providers/__init__.py
+++ b/faker/providers/__init__.py
@@ -405,7 +405,7 @@ class BaseProvider:
 
     def random_elements(
         self,
-        elements: ElementsType[T] = ("a", "b", "c"),
+        elements: ElementsType[T] = ("a", "b", "c"),  # type: ignore[assignment]
         length: Optional[int] = None,
         unique: bool = False,
         use_weighting: Optional[bool] = None,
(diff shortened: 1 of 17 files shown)


fcurella (approved):

fcurella:
Thank you!

fcurella (changes requested):
I've amde some change to `Provider.filter_by_length`. Could you rebase against `master`?

fcurella (commented):
Thank you!

PR #1737, opened 2022-10-15
Add color ID

rosdyana:
### What does this change

- Add color provider for ID (Indonesia)
- Add test for ID color provider


fcurella (changes requested):
Thank you! Could you run `make lint` to fix the formatting?

PR #1739, opened 2022-10-21
Sort and add new Arabic words

alfawal:
### What does this change

The order of names in person directory.

### What was wrong

Some variables contain unordered tuples.

### How this fixes it

- Adds new names.
- Sorts the existing Arabic values.


fcurella (changes requested):
Thank you! Could you upate your base with latest changes from `master`?

fcurella (approved):
I can take it form here. Thank you!

PR #1741, opened 2022-10-28
Fix pyfloat causing ValueError

### LINKED ISSUE #1718: ValueError: empty range for pyfloat
* Faker version: 14.0.0
* OS: macOS 12.5 (Monterey)

pyfloat throws ValueError when min_value and max_value have the same left digits. 

### Steps to reproduce

```
from faker import Faker

fake = Faker()

fake.pyfloat(min_value=1.25, max_value=1.75)
```

### Expected behavior

pyfloat returns a float between min_value and max_value

### Actual behavior

pyfloat throws exception

```
ValueError: empty range for randrange() (1, 1, 0)
```

confuzeus:
When given floats with same wholes, a
ValueError will be thrown by the randrange
method because the random_int method
will provide it with 2 integers of the same value.

Example: pyfloat(min_value=2.3, max_value=2.5) will call randrange(2,2) which isn't valid.

See line 181-194 for the actual meat of this PR. The other changes were caused by Black.

Fixes #1718

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -58,7 +58,13 @@ class Provider(BaseProvider):
     def pybool(self) -> bool:
         return self.random_int(0, 1) == 1
 
-    def pystr(self, min_chars: Optional[int] = None, max_chars: int = 20, prefix: str = "", suffix: str = "") -> str:
+    def pystr(
+        self,
+        min_chars: Optional[int] = None,
+        max_chars: int = 20,
+        prefix: str = "",
+        suffix: str = "",
+    ) -> str:
         """
         Generates a random string of upper and lowercase letters.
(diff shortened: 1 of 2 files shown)


confuzeus:
Tests are failing in the master branch but this code works.

fcurella (changes requested):
Thank you ! Can you add a test case for regression purposes?

confuzeus:
@fcurella Done at line 239.

Btw, I had to add the quick hack at line 218 to make sure the result is always >= max_value. A better algorithm could be implemented in the future.

fcurella (changes requested):
Can you fix the linting issues? You can run `make lint` to run the linters with our settings

confuzeus:
@fcurella Sorry auto formatting messed up the code. It's done now.

PR #1743, opened 2022-11-01
fixed issue #1742

### LINKED ISSUE #1742: Missing return in faker-master\faker\providers\address\en_US\__init__.py
* Faker version: current
* OS: Windows 10 x64

Missing return on row 496 (def state_abbr(self, include_territories: bool = True) -> str:)

### Steps to reproduce

1. invoke state_abbr()
2. check the output 

### Expected behavior
Should return an element from states_and_territories_abbr list

### Actual behavior
Returns an element from states_abbr list only

AssenD:
Fixes #1742

### DIFF
--- a/faker/providers/address/en_US/__init__.py
+++ b/faker/providers/address/en_US/__init__.py
@@ -493,7 +493,7 @@ class Provider(AddressProvider):
             If False, only states will be returned.
         """
         if include_territories:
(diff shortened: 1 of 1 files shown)


fcurella (approved):

fcurella (commented):
Thank you!

PR #1746, opened 2022-11-03
Update the word_list

BeRT2me:
Data from 2022-11-03

### What does this change

Updates the word list.

### What was wrong

Word list was outdated - and contained swear words.

### How this fixes it

Updates the word list, thereby removing the swear words.


fcurella (approved):
Thank you!

PR #1748, opened 2022-11-12
Fix implicit Optional issues

palfrey:
### What does this change

Corrects various types that had a `None` default, but a type without `None` being a valid value.

### What was wrong

Now that https://github.com/python/mypy/pull/13401 is merged, latest mypy doesn't permit this anymore

### How this fixes it

Makes them `Optional[X]` not `X` types

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,7 +162,7 @@ jobs:
       uses: liskin/gh-problem-matcher-wrap@v1
       with:
         linters: mypy
-        run: mypy --show-column-numbers --install-types --non-interactive --config mypy.ini faker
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -162,7 +162,7 @@ jobs:
       uses: liskin/gh-problem-matcher-wrap@v1
       with:
         linters: mypy
-        run: mypy --show-column-numbers --install-types --non-interactive --config mypy.ini faker
+        run: mypy --install-types --non-interactive --config mypy.ini faker
```
palfrey, line 165: This is now pulled into config so local running gets the same results

fcurella (approved):
Thank you!

PR #1749, opened 2022-11-13
Replace archived `faker-optional` with my alternative

PerchunPak:
[`faker-optional`](https://github.com/lyz-code/faker-optional) was archived 9 months ago. I recently created an alternative to replace it.

### DIFF
--- a/docs/communityproviders.rst
+++ b/docs/communityproviders.rst
@@ -42,9 +42,9 @@ Here's a list of Providers written by the community:
 +---------------+---------------------------+----------------------------------+
 | Wi-Fi ESSID   | Fake Wi-Fi ESSIDs.        | `faker_wifi_essid`_              |
 +---------------+---------------------------+----------------------------------+
-| Optional      | Wrap over other           | `faker_optional`_                |
-|               | providers to return       |                                  |
-|               | their value or `None`.    |                                  |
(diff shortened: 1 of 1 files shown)


stefan6419846:
Just from a general development perspective: Does it really make sense to create a completely new repository without the old commit history and publish a new package under a different name when the original repository explicitly states:

> No longer maintained: I use [pydantic-factories](https://github.com/Goldziher/pydantic-factories) instead, if you want to maintain this repo, I can transfer it to you.

PerchunPak:
I wrote the author, and he didn't answer for ~3 days. I also didn't take anything from the original repo, and wrote everything from zero.

If you look at [old implementation](https://github.com/lyz-code/faker-optional/blob/master/src/faker_optional/model.py), [and mine](https://github.com/PerchunPak/optional-faker/blob/master/optional_faker.py), you will see that they are completely different.

stefan6419846:
Thanks for the explanations - I did not have a closer look at the actual code and you were speaking of a fork as well.

fcurella (approved):
Thank you!

lyz-code:
I'm the author of the old code and I'm fine with the change. If I were quicker answering maybe I could have transferred the repo.

I take the chance to thank you all again for maintaining `faker` :)

PR #1751, opened 2022-11-18
Add missing number in SHA-256 docstring and add hyphen for SHA hashes

gustavi:
### What does this change

- Small typo: `SHA56` => `SHA256`
- Rename `SHA<N>` in `SHA-<N>` to match the name used in RFCs and Wikipedia

### What was wrong

See previous

### How this fixes it

See previous


fcurella (approved):
Thank you!

PR #1754, opened 2022-11-22
Fix/seeding

### LINKED ISSUE #1656: Faker returns one value when using multiple locales
* Faker version: 13.11
* OS: Mac OS Monterey
Computer: MacBook Pro (14-inch, 2021)

If using multiple locales, faker only makes one choice. This only happens when you select multiple locales

### Steps to reproduce

```
from faker import Faker

fake = Faker(["en_GB", "fr_FR", "en_IN"])
Faker.seed(0)
for _ in range(5):
    print(fake.date_time_this_year())

```

### Expected behavior
returns
```
2022-03-16 19:22:23
2022-03-23 16:00:20
2022-01-08 20:40:15
2022-02-20 06:38:22
2022-04-10 06:42:46
```

### Actual behavior
returns
```
2022-03-23 16:00:20
2022-03-23 16:00:20
2022-03-23 16:00:20
2022-03-23 16:00:20
2022-03-23 16:00:20

```

johntmyers:
fixes #1656 

### What does this change

This removes re-seeding the `Random()` instance on every call to `_select_factory()` which gets called every time a faker method is used. This code was only reached when both conditions are true:

- A seed is set via `Faker.seed()`
- More than one locale is specified

### What was wrong

Basically same as above, it is unnecessary to set the `Random()` instance's seed value every time a faker method is used. The `Random()` instance is seeded when making the call to `Faker.seed()` which is sufficient for making sure that the order of the locales selected is the same between subsequent instances of `Faker`

### How this fixes it

Remove the seeding that isn't needed and modified a test that asserts:

- Generating values when a seed is used and multiple locales are used does not generate the same value over and over
- Calling `Faker.seed()` and using a `Faker` instance in an identical way will produce the same fake output each time

Fixes #...

### DIFF
--- a/tests/test_proxy.py
+++ b/tests/test_proxy.py
@@ -141,7 +141,6 @@ class TestFakerProxyClass:
 
         assert first_list == second_list
 
-
     def test_seed_instance(self):
         locale = ["de_DE", "en-US", "en-PH", "ja_JP"]
         fake = Faker(locale)
--- a/tests/test_proxy.py
+++ b/tests/test_proxy.py
@@ -133,6 +133,8 @@ class TestFakerProxyClass:
         count = 5
         fake = Faker(["en_GB", "fr_FR", "en_IN"])
         first_list = [fake.name() for _ in range(count)]
+        # We convert the list to a set to remove duplicates and ensure
+        # that we have exactly `count` unique fake values
         assert len(set(first_list)) == count
 
         Faker.seed(2043)
--- a/tests/test_proxy.py
+++ b/tests/test_proxy.py
@@ -130,12 +130,14 @@ class TestFakerProxyClass:
 
     def test_seed_class_locales(self):
         Faker.seed(2043)
+        count = 5
         fake = Faker(["en_GB", "fr_FR", "en_IN"])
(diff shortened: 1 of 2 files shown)


tests/test_proxy.py
```diff
@@ -130,11 +130,17 @@ def test_seed_classmethod(self):
 
     def test_seed_class_locales(self):
         Faker.seed(2043)
+        count = 5
         fake = Faker(["en_GB", "fr_FR", "en_IN"])
-        name = fake.name()
+        first_list = [fake.name() for _ in range(count)]
+        assert len(set(first_list)) == count
```
fcurella, line 136: what's the reason for casting to a `set`?

fcurella: got it 👍🏼  Would you add a comment above the assertion?

fcurella:
@johntmyers You can run `make lint` to format the code with our settings.

tests/test_proxy.py
johntmyers: done

johntmyers:
@fcurella everything is passing now

fcurella (approved):
Thank you!

PR #1758, opened 2022-11-25
Fix IT SSN generation with latin chars in name or surname

### LINKED ISSUE #1757: ssn.it_IT.Provider generates SSNs shorter than 16 chars
* Faker version:  tested with 13.14.0,  15.3.3
* OS: Ubuntu 22.04

Generating an Italian SSN, when the last (or first) name returned by `person.it_IT.Provider` is 3 characters long and contains latin chars, the resulting string is shorter than expected (16 chars). In this scenario:

- string is not shorter than 3 chars, so it does not get padded
- latin char is neither a vowel nor a consonant, so it gets ignored

### Steps to reproduce

```
from unittest import mock 

from faker import Factory

generator = Factory.create("it_IT")
generator.last_name = mock.MagicMock(return_value="Foà")

assert len(generator.ssn()) == 16

AssertionError
```

### Expected behavior

The length of generated SSN should be 16 characters. Latin chars should be translated into equivalent ASCII chars (A for À, E for È and É, etc.)  

### Actual behavior

The length of generated SSN is 15 characters.

panicofr:
### What does this change

Fix IT SSN generation when Person provider returns first or last name with latin chars

### What was wrong

Latin chars was considered neither vowel nor consonant. They just got ignored, and the resulting SSN was wrong, or even shorter than expected length (16 chars) in case of short names/surnames.

### How this fixes it

This change applies a transliteration for dyacritic chars, translating them into the equivalents ASCII chars, according to the Italian SSN generation rules.

Fixes #1757

### DIFF
--- a/faker/providers/internet/el_GR/__init__.py
+++ b/faker/providers/internet/el_GR/__init__.py
@@ -68,7 +68,7 @@ def latinize(value: str) -> str:
 
     def replace_greek_character(match):
         matched = list(match.group(0))
-        value = map(lambda l: replace[search.find(l)], matched)
+        value = map(lambda letter: replace[search.find(letter)], matched)
         return "".join(value)
 
     return re.sub(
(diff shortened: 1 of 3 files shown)


fcurella (changes requested):
Thank you! Can you run `make lint` to format the code?

panicofr (commented):
Done 👍

fcurella (approved):
Thank you!

PR #1759, opened 2022-11-30
Suggest `faker --help` if CLI arguments are missing

### LINKED ISSUE #1187: Lots of output at terminal
If I just enter `faker` at the terminal and hit enter, is it supposed to produce nearly 50mb of output? The text flew by so fast, I piped it into a text file, and it was something like 46mb of text. I'm fairly confident this isn't intended, but I obviously can't say for sure, since I'm not the one who wrote it. If this isn't intended, I can provide more details.

mportesdev:
### What does this change

Attempt to make the command-line output of `faker` (invoked without arguments) more user friendly.

### What was wrong

Running `faker` prints several megabytes of sample data to stdout, which is confusing.

### How this fixes it

Running `faker` displays command's usage with an additional message `Try 'faker --help' for more information.`.

Fixes #1187

### DIFF
--- a/faker/cli.py
+++ b/faker/cli.py
@@ -267,6 +267,10 @@ examples:
             "first argument)",
         )
 
+        if not self.argv[1:]:
+            parser.print_usage(sys.stderr)
+            parser.exit(2, f"Try '{self.prog_name} --help' for more information.\n")
+
         arguments = parser.parse_args(self.argv[1:])
 
         if arguments.verbose:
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

fcurella:
we can always update the README

mportesdev:
Alternatively, we can limit the output to only the selected locale.

https://github.com/mportesdev/faker/commit/21af1b1ebada766522dceecaff86e104a16c6afc

For me, this reduces the output from 11 to 0.7 MB. Running `faker` gives me sample data for `en_US`, `faker -l de_DE` for German etc.

fcurella:
@mportesdev I like the idea of printing only the selected locale.

I also wonder if we should use `--sample` instead of `--help`

PR #1763, opened 2022-12-13
Update spanish phone number formats

GuillemCalidae:
According to the official specs:
https://avancedigital.mineco.gob.es/es-ES/Servicios/Numeracion/Documents/Guia_Numeracion.pdf

### What does this change

Adds restrictions to some phone numbers generations.

### What was wrong

Some phone numbers generated weren't valid according to official documentation. While the whole 8- and 9- prefixes are supposed to be landline numbers not all following numbers are assigned to an actual geographic area and are invalid.

### How this fixes it

Add phone numbers prefixes to increase the specificity of the numbers generated.

Also refactored the formats tuple to be generated from a product of prefixes and phone numbers, which makes sense now to avoid a list of 400+ elements.


fcurella (changes requested):
Can you run `make lint` to fix the formatting issues?

faker/providers/phone_number/es_ES/__init__.py
```diff
@@ -1,59 +1,129 @@
...
+    PHONE_FORMATS = (
+        "+34 xxx ### ###",
+        "+34 xxx######",
+        "+34 xxx ## ## ##",
+        "+34xxx ### ###",
+        "+34xxx######",
+        "+34xxx ## ## ##",
+    )
+    formats = tuple(
+        f'{phone_format.replace("xxx", prefix)}'
```
n1ngu, line 127: Or simply
```
    formats = tuple(
        phone_format.replace("xxx", prefix)
        for (prefix, phone_format) in itertools.product(PREFIXES, PHONE_FORMATS)
    )
```

n1ngu (changes requested):

fcurella (approved):
Thank you!

PR #1767, opened 2022-12-14
add code and test suite for fr_CA address provider

davebulaval:
Add a new address provider for `fr_CA` with the appropriate tests.


davebulaval:
The typing seems to fail since I use the `en_CA` provider and overload some stuff. A solution would be to remove class inheritance and copy-paste the code from `en_CA`.

fcurella (commented):
Thank you! I've fixed the typing issues on the provider's superclass.

PR #1772, opened 2022-12-28
Add market-data identifiers provider

rubenafo:
### What does this change

Adds a new provider (faker-marketdata) to generate fake market data widely used in financial services (ISIN, SEDOL, CUSIP, etc.). List of supported IDs is in the [documentation](https://github.com/rubenafo/faker_marketdata).

### What was wrong

No bugs fixed with this PR


fcurella (commented):
Thank you!

PR #1777, opened 2023-01-05
Fix name_female generates mixed names (#1770)

### LINKED ISSUE #1770: name_female generates mixed names
* Faker version: Lastest
* OS: NA

Using `name_female` actually ends up generating mixed names, the same is true for `name_male`.

### Steps to reproduce
```python
from faker import Faker

fake = Faker("en-GB")

for i in range(1, 100):
    print(fake["en-GB"].name_female())
```
Example output:
```
...
Mr. Russell Morgan
Mr. Vincent Frost
Ruth Perry
Jacob Harris-Fitzgerald
Frances Robinson
Guy Hayward
Pamela Sanders-Davis
Frederick Doyle
Diana Hayes
Diane Hudson
Mr. Leonard Simpson
Mrs. Natalie Wilson
...
```
### Expected behavior

Generates only female names.

### Actual behavior

Generates mixed names.

This appears to be because `formats_female` is not defined, which then causes the function to fall back to a general format.

https://github.com/joke2k/faker/blob/7e78eef2565c5ce78b879bf271f520e817f55d4f/faker/providers/person/__init__.py#L230-L236

nathanael-e:
### What does this change

Split formats in person/en_GB into formats_female and formats_male. 

### What was wrong

faker.name_female() and faker.name_male() would previously generate mixed names (both female and male names) when using locale en_GB. 

### How this fixes it

formats_female and formats_male are now defined and will no longer fall back on the general format when calling name_female() or name_male().

Fixes #1770

### DIFF
--- a/faker/providers/person/en_GB/__init__.py
+++ b/faker/providers/person/en_GB/__init__.py
@@ -4,22 +4,28 @@ from .. import Provider as PersonProvider
 
 
 class Provider(PersonProvider):
-    formats = (
-        "{{first_name_male}} {{last_name}}",
-        "{{first_name_male}} {{last_name}}",
-        "{{first_name_male}} {{last_name}}",
-        "{{first_name_male}} {{last_name}}",
-        "{{first_name_male}} {{last_name}}-{{last_name}}",
+
+    formats_female = (
         "{{first_name_female}} {{last_name}}",
         "{{first_name_female}} {{last_name}}",
         "{{first_name_female}} {{last_name}}",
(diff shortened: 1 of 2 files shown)


Dutcho:
This issue occurs wider than just `'en_GB'`, resulting in equally wrong `name_female()`/`name_male()` results for `'nl_NL'` (which I ran into) and `'en_IE'` (which I checked), and possibly more (but not `'en_US'`)

fcurella (approved):
I can take it from here. Thank you!

PR #1783, opened 2023-01-12
Add method to generate JSON as bytes

browniebroke:
### What does this change

Add wrapper method around the JSON provider to generate JSON as bytes.

### What was wrong

I use Faker through factory boy and couldn't find a convenient way to achieve that. The change is quite small so I thought it could be a nice addition to the library.

### How this fixes it

Wraps the existing JSON provider and encode its output to bytes.


fcurella (approved):
Thank you!

PR #1786, opened 2023-01-22
Add `faker-datasets` to the community providers table

cavokz:
Fixes https://github.com/elastic/faker-datasets/issues/7.


cavokz:
I fail to understand how to fix the error reported by appveyor, I also observed some of the other providers in the list but could not spot the trick.

Could you please give me an hint?

stefan6419846:
The AppVeyor builds have been failing for quite some time now and were always ignored, so no worries. GitHub Actions should not fail, which still requires approval by one of the maintainers to actually run as you are a first-time contributor.

cavokz:
> The AppVeyor builds have been failing for quite some time now and were always ignored, so no worries. GitHub Actions should not fail, which still requires approval by one of the maintainers to actually run as you are a first-time contributor.

Ok. Thank you!

PR #1789, opened 2023-02-09
Remove deprecated placeimg image provider

### LINKED ISSUE #1788: Remove image provider about to be shutdown
I randomly discovered that https://placeimg.com was about to be shut down (on june, 30th 2023). I believe it should therefore be removed

https://github.com/joke2k/faker/blob/7ca5eac81157c2792237dc1a3136a955d4464143/faker/providers/internet/__init__.py#L160

Wonskcalb:
### What does this change

Image placeholder services

### What was wrong

The service will close down on June 30th 2023. Since this could
create errors in users' code, better get rid of it in due time

### How this fixes it

It's removed from the code

Fixes #1788

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -157,7 +157,6 @@ class Provider(BaseProvider):
         "https://picsum.photos/{width}/{height}",
         "https://dummyimage.com/{width}x{height}",
         "https://placekitten.com/{width}/{height}",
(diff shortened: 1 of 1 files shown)


fcurella (approved):
Thank you!

Wonskcalb:
Happy to help (it's also my first FOSS contribution, although minimal, I'm jazzed :sweat_smile: )

PR #1793, opened 2023-02-11
Implement truth probability for boolean generator

Kolyunya:
### What was wrong
Currently, it's impossible to generate booleans that are biased towards `True` or `False` values.

### How this fixes it
This patch adds an optional `truth_probability` argument to the `pybool()` method which allows to generate booleans that are biased towards `True` or `False` values.

Having the new argument is optional and defaults to `50%`, this patch is fully backward compatible.

### Example
Imagine we need to generate random booleans but we only want 10% percent of them to be equal to `True`. After this patch it would be possible using the following code:

```python
boolean = faker.pybool(10)
boolean = faker.pybool(truth_probability=10)
```

### Side notes
Our [PHP friends](https://github.com/fzaninotto/Faker) already have this feature [implemented](https://github.com/fzaninotto/Faker#fakerprovidermiscellaneous) which proves that it is actually useful for the end-users. 

PHP implementation has `chanceOfGettingTrue` as the name of the argument, but I believe that `truth_probability` would fit better.

Seems like it's impossible to use `parametrize` in that test case because it extends `unittest.TestCase`, that's why I added a helper method to the test class.


fcurella (approved):
Thank you!

Kolyunya:
@fcurella my pleasure!

PR #1795, opened 2023-02-12
Added PresidioSentenceFaker to community providers

omri374:
### What does this change

Added a link to the `presidio-evaluator` python package, which contains a fake data generator for creating synthetic/augmented Named Entity Recognition datasets. It extends `Faker` by returning the start and end of each fake value, as a mechanism to auto-label datasets for NER training.

Presidio Evaluator: https://github.com/microsoft/presidio-research
Data generator module: https://github.com/microsoft/presidio-research/blob/master/presidio_evaluator/data_generator/README.md


fcurella (approved):
Thank you!

PR #1796, opened 2023-02-16
Add support for Python 3.11

mportesdev:
### What does this change

- add Python 3.11 to tox test environments
- add Python 3.11 to GitHub Actions test matrices
- add Python 3.11 to classifiers in setup.py


fcurella (approved):
Thank you!

PR #1797, opened 2023-02-17
Add argument validation to `pybool()`

Kolyunya:
### What was wrong
`pybool()` didn't have argument validation.

### How this fixes it
I've added argument validation to `pybool()`.

### What does this change
After the patch, an exception would be thrown when invalid `truth_probability` value is provided. 

I'm really not sure about this one, and you may consider this a redundant check.

But it seems to me that when someone provides a probability value of `-5` this may be a clear sign that something's wrong with the client code, thus it's better to report the problem instead of silently ignoring it.

Note that technically this patch may be considered as a breaking change.

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -59,9 +59,13 @@ class Provider(BaseProvider):
         """
         Generates a random boolean, optionally biased towards `True` or `False`.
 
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

Kolyunya:
@fcurella my pleasure!

PR #1802, opened 2023-02-24
Add zh_TW date_time provider

### LINKED ISSUE #1801: Support Minguo calendar years
* Faker version: 17.0.0
* OS: macOS

[Minguo calendar years](https://en.wikipedia.org/wiki/Republic_of_China_calendar) are commonly and officially used in Taiwan. It's not a difficult conversion from Gregorian years, but it would be helpful and convenient to have a function to generate random years in the Minguo calendar, maybe under the `zh_TW` provider.

ew2664:
### What does this change

Adds zh_TW date_time provider, which provides month and day of week names in traditional Chinese. Also supports generating [Minguo calendar years](https://en.wikipedia.org/wiki/Republic_of_China_calendar), which is commonly and officially used in Taiwan.

### What was wrong

See #1801 

### How this fixes it

Provides a `minguo_year()` function under the `zh_TW` provider that works similarly to the `year()` function, except the generated Gregorian years are converted to the Minguo calendar.

Fixes #1801

### DIFF
--- a/tests/providers/test_date_time.py
+++ b/tests/providers/test_date_time.py
@@ -40,6 +40,7 @@ from faker.providers.date_time.sl_SI import Provider as SlSiProvider
 from faker.providers.date_time.ta_IN import Provider as TaInProvider
 from faker.providers.date_time.tr_TR import Provider as TrTrProvider
 from faker.providers.date_time.zh_CN import Provider as ZhCnProvider
+from faker.providers.date_time.zh_TW import Provider as ZhTwProvider
 
 
 def is64bit():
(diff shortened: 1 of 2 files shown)


tests/providers/test_date_time.py
```diff
@@ -1192,8 +1193,27 @@ def test_day(self):
...
+
+    def test_month(self):
+        month = self.fake.month_name()
+        assert month in ZhTwProvider.MONTH_NAMES.values()
+
+    def test_minguo_year(self):
+        year = self.fake.minguo_year()
+        assert isinstance(year, str)
+        assert year.isdigit()
+        assert 1 <= len(year) <= 3
```
fcurella, line 1216: could you also add a test for the inherited `year()` method?

fcurella (changes requested):

fcurella (approved):
Thank you!

PR #1804, opened 2023-02-24
Add SBN provider

### LINKED ISSUE #1803: Support for 9 digit SBN
* Faker version: 17.0.0
* OS: Windows 10

Before the formation of the ISBN there was a non-international Standard Book Number (SBN) used for a few years. While no longer used today, it might be nice to include SBNs as they are still on many books from the time. I'd be happy to work on a PR for this.

dag2226:
### What does this change

Adds an SBN provider that generates (pre-International) Standard Book Numbers.

### What was wrong

See #1803 

### How this fixes it

Takes the current ISBN provider implementation and modifies it to match the older SBN model. This includes removing the usage of EAN and registration group in number generation, as well as editing the `_check_digit` method to work for SBN.

Fixes #1803

### DIFF
--- a/faker/providers/sbn/__init__.py
+++ b/faker/providers/sbn/__init__.py
@@ -1,0 +1,53 @@
+from typing import List, Tuple
+
+from faker.providers.sbn.rules import RegistrantRule
+
+from .. import BaseProvider
+from .rules import RULES
+from .sbn import SBN, SBN9
+
+
+class Provider(BaseProvider):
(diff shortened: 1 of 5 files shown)


fcurella (changes requested):
Thank you! Could you run `make lint` to format the code?

dag2226:
My apologies for the mess, this is my first contribution and I'm still getting bogged down by a few things. Should I close this PR and open a new one with a cleaner commit history?

PR #1809, opened 2023-02-28
Use the official name for North Macedonia due to the name change in 2019

takis:
Translations verified using Wikipedia and Google Translate.

The name change was the result of the Prespa Agreement: https://en.wikipedia.org/wiki/Prespa_Agreement


fcurella (approved):
Thank you!

faker/providers/address/uk_UA/__init__.py
```diff
@@ -651,6 +650,7 @@ class Provider(AddressProvider):
         "Південний Судан",
         "Польща",
         "Португалія",
+        "Респу́бліка Півні́чна Македо́нія",
```
PerchunPak, line 653: This is actually cursed. Symbols, like those on letters `у́`, `ч` (lol, GitHub doesn't recognize it because it's an illegal symbol) and `о́` are not wrote in words. Those are used only for marking pronunciation in Ukrainian. So the right name here should be:

```
Північна Республіка Македонія
```

P.S. I see that some translations lack word `Republic`, how about add `TODO` comment to all of the changed lines, with content like `Please review this string, should be "Republic of North Macedonia". See #1809`.
P.S.2 I see that some languages have more than 2 changes, what does it mean?

fcurella: Thank you noticing @PerchunPak, I'll fix this entry.

PR #1810, opened 2023-02-28
Add nl_BE automotive provider

takis:


fcurella (changes requested):
Could you add a test case for the new provider?

takis:
Hi Flavio,

Sure! I've added one now.

The black check fails, but the entire file wasn't black-formatted before my patch either. I tried to follow the same
style for the regex, which is what black dislikes.

```python
pattern = foo(
    r"[A-Z]|"
    r"[0-9]"
)
```

Which black would prefer to see as:
```python
pattern = foo(
    r"[A-Z]|[0-9]"
)
```
And of course for such a short regex I would agree.

takis:
Hi @fcurella 

black complaining was caused by the same issue as in the #1820 pull request: I was using the default black configuration for formatting the code omitting the` --line-length 120` using in the Makefile.

fcurella (approved):
Thanks you!

PR #1820, opened 2023-03-13
Add Dutch lorem provider

takis:
Provide a lorem provider for both nl_BE and nl_NL locales.


fcurella (changes requested):
can you run `make lint` once more?

takis:
Sure!

I see the issue now: I was using black without command line parameters as I typically use a `pyproject.tml` file containing the black configuration options.

fcurella:
Thank you!

PR #1821, opened 2023-03-13
Issue #1812 - added territory zip codes

### LINKED ISSUE #1812: states_postcode fails with territories
* Faker version: since 15.1.2
* OS:  All

en_US state fix in 15.1.2 added territories, which are not in states_postcode

Looking up postcode by state fails with territories enabled, but didn't fail prior to 15.1.2.

The fix made here:
  https://github.com/joke2k/faker/blob/master/faker/providers/address/en_US/__init__.py#L496
Combined with the default to include territories here:
  https://github.com/joke2k/faker/blob/master/faker/providers/address/en_US/__init__.py#L488
Causes this to fail whereas it did not prior to the correction made in 15.1.2.
  https://github.com/joke2k/faker/blob/master/faker/providers/address/en_US/__init__.py#L505

### Steps to reproduce

1. fake.add_provider(providers.address.en_US.Provider)
2. state = fake.state_abbr()  # include_territories=True is default.  user can fix by setting to False?
3. repeatedly call fake.zipcode_in_state(state)  # or alternatively pass in GU or other territory

### Expected behavior
Ideally this would not give an error when combining these functions.
It looks like territories to have zip codes, so best solution may be to update the state/postalcode mapping.

--> from internet:  "Within the U.S. Territories, American Samoa (postal abbreviation AS) uses zip code 96799, and Guam (postal abbreviation GU) uses zip codes in the range 96910–96932."

### Actual behavior
Fails on Step 3 noted above.

johnekent:
### What does this change

Added postal codes for territories.

### What was wrong

Default state abbreviations returned territories, which were not in the postal code map.

### How this fixes it

Update postal code map with territories, including length handing.

Fixes #1812

### DIFF
--- a/faker/providers/address/en_US/__init__.py
+++ b/faker/providers/address/en_US/__init__.py
@@ -419,6 +419,15 @@ class Provider(AddressProvider):
         "WV": (24701, 26886),
         "WI": (53001, 54990),
         "WY": (82001, 83128),
+        # Territories - incomplete ranges with accurate subsets - https://www.geonames.org/postalcode-search.html
+        "AS": (96799, 96799),
+        "FM": (96941, 96944),
+        "GU": (96910, 96932),
+        "MH": (96960, 96970),
+        "MP": (96950, 96952),
+        "PW": (96940, 96940),
+        "PR": (600, 799),
+        "VI": (801, 805),
     }
 
     territories_abbr = (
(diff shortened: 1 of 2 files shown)


fcurella (approved):

fcurella (approved):
Thank you!

Issue #1779, opened 2023-01-09
Add dynamic type python generator


dariocurr:
* Faker version: 15.3.4
* OS: linux

### Feature request

It would be awesome to dynamically generate object passing the type as argument. Something like:

```py
from typing import Type

def generate_random(object_type: Type, **kwargs):
    if object_type == int:
        callable = fake.pyint
    elif object_type == float:
        callable = fake.pyfloat
    elif object_type == bytes:
        callable = fake.binary
    elif object_type == str:
        callable = fake.pystr
    elif object_type == bool:
        callable = fake.pybool
    elif object_type == dict:
        callable = fake.pydict
    elif object_type == list:
        callable = fake.pylist
    elif object_type == set:
        callable = fake.pyset
    elif object_type == tuple:
        callable = fake.pytuple
   callable(**kwargs)
```

I took a look around and I didn't find anything providing or discussing this feature

stefan6419846:
Just out of curiosity: Do you have any special use-case in mind which makes using the single methods this complicated?

dariocurr:
The type of the object may not be certain in some use cases.
For example:
 I am testing the behavior of a component that contains a dict, whose key can have different types and must be specified in the component's metadata.
To check that everything works well, I read the metadata and extract the key type. Then to test the correct insertion I have to generate an object of that type not known a priori.

That's why currently to generate such random object I have implemented the `if-elif` by myself

fcurella:
This seems reasonable. Do you have time to prepare a Pull Request?

dariocurr:
I guess I could work on it but I would like to receive some guidelines.

For example, should I take in account just the methods, then object types, in the `python` [provider](https://faker.readthedocs.io/en/master/providers/faker.providers.python.html)?

stefan6419846:
Starting with the Python providers should be sufficient for now in my opinion. Adding support for other providers (like for file or dates and times) could still be added in a future change if useful for somebody.

dariocurr:
I'm closing this issue since #1822 has been merged

PR #1822, opened 2023-03-14
Implement generate pyobject by type #1779

dariocurr:
### What does this change

It solves issue #1779

Note: I had hard time dealing with `mypy` and `kwargs`

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -55,6 +55,40 @@ class Provider(BaseProvider):
             allowed_types = ()
         return tuple(value_types) + tuple(allowed_types)
 
+    def pyobject(
+        self,
+        object_type: Optional[Type[Union[bool, str, float, int, tuple, set, list, Iterable, dict]]] = None,
+    ) -> Optional[Union[bool, str, float, int, tuple, set, list, Iterable, dict]]:
+        """
+        Generates a random object passing the type desired.
+
+        :object_type: the type of the object to generate.
+        :return: the random object generated.
+        :raises ValueError: if the object type passed is not supported
+        """
+        if object_type is None:
+            return None
+        elif object_type == bool:
+            return self.pybool()
+        elif object_type == str:
+            return self.pystr()
+        elif object_type == float:
+            return self.pyfloat()
+        elif object_type == int:
+            return self.pyint()
+        elif object_type == tuple:
+            return self.pytuple()
+        elif object_type == set:
+            return self.pyset()
+        elif object_type == list:
+            return self.pylist()
+        elif object_type == Iterable:
+            return self.pyiterable()
+        elif object_type == dict:
+            return self.pydict()
+        else:
+            raise ValueError("Object type `{object_type}` is not supported by `pyobject` function")
+
     def pybool(self, truth_probability: int = 50) -> bool:
         """
         Generates a random boolean, optionally biased towards `True` or `False`.
--- a/tests/providers/test_python.py
+++ b/tests/providers/test_python.py
@@ -3,7 +3,7 @@ import sys
 import unittest
 import warnings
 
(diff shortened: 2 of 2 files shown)


faker/providers/python/__init__.py
```diff
@@ -55,6 +55,40 @@ def _check_signature(self, value_types: Optional[TypesSpec], allowed_types: Opti
             allowed_types = ()
         return tuple(value_types) + tuple(allowed_types)
 
+    def pyobject(
+        self,
+        object_type: Optional[Type[Union[bool, str, float, int, tuple, set, list, Iterable, dict]]] = None,
+    ) -> Optional[Union[bool, str, float, int, tuple, set, list, Iterable, dict]]:
```
fcurella, line 61: I think we should be able to use some version of [`TypeVar`](https://docs.python.org/3/library/typing.html#typing.TypeVar) here:

```python3
TObjectType = TypeVar('TObjectType', Type[Union[None, bool, str, float, int, tuple, set, list, Iterable, dict]])

class Provider:
    ...
    def pyobject(self, object_type: TObjectType = None): -> TObjectType:
        ...
```

tests/providers/test_python.py
```diff
@@ -3,13 +3,46 @@
...
+        (str, {}),
+        (float, {}),
+        (int, {}),
+        (tuple, {}),
+        (set, {}),
+        (list, {}),
+        (Iterable, {}),
+        (dict, {}),
+    ),
+)
```
fcurella, line 28: I dont see the point in defining `kwargs`, since it's always an empty dict. We could just have:

```suggestion
@pytest.mark.parametrize(
    "object_type",
    (
        (None,),
        (bool,),
        (str,),
        (float,),
        (int,),
        (tuple,),
        (set,),
        (list,),
        (Iterable,),
        (dict,),
    ),
)
```

tests/providers/test_python.py
```diff
@@ -3,13 +3,46 @@
...
+def test_pyobject(
+    object_type: Optional[Union[bool, str, float, int, tuple, set, list, Iterable, dict]],
```
fcurella, line 30: If we define a `TypeVar`, we can import it and use it here

fcurella (approved):

faker/providers/python/__init__.py
dariocurr: Good suggestion.

Beware of the return type, which is not a type

fcurella: oh, that's right. I wonder if something like this would work?

```python3
TObjectType = TypeVar('TObjectType', Union[None, bool, str, float, int, tuple, set, list, Iterable, dict])

class Provider:
    ...
    def pyobject(self, object_type: Type[TObjectType] = None): -> TObjectType:
        ...
```

dariocurr: It doesn't work with mypy

tests/providers/test_python.py
```diff
@@ -3,13 +3,46 @@
...
+        (tuple,),
+        (set,),
+        (list,),
+        (Iterable,),
+        (dict,),
+    ),
+)
+def test_pyobject(
+    object_type: Optional[Union[bool, str, float, int, tuple, set, list, Iterable, dict]],
+    kwargs: Dict[str, Any],
```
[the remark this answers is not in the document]
dariocurr: Solved with https://github.com/joke2k/faker/pull/1822/commits/7671aadc1193105b565592c01bf29ec499c8e6ef

fcurella (approved):
Thank you!

tests/providers/test_python.py
dariocurr: Same as [here](https://github.com/joke2k/faker/pull/1822#discussion_r1135813843)

dariocurr:
Just to have an update. When will it be merged?

PR #1826, opened 2023-03-20
Ensure pyfloat honors min and max values

### LINKED ISSUE #1825: pyfloat produces values outside of the allowed range.
* Faker version: 17.6.0
* OS: Windows 10

`pyfloat` produces values outside of the allowed range.
It appears this is related to the fact that the provided `min_value` and `max_value` are overwritten see [line 150](https://github.com/joke2k/faker/blob/master/faker/providers/python/__init__.py#L150). So even the final check to see if the result exceeds the `max_value` will not catch these scenarios. 

### Steps to reproduce

```python
for i in range(10000):
    min_x, max_x = (-120.123, -115.321)
    min_y, max_y = (20.123, 30.321)
    x= fake.pyfloat(min_value=min_x, max_value=max_x)
    y= fake.pyfloat(min_value=min_y, max_value=max_y)
    if x < min_x:
        print(f'Exceeds min {x} < {min_x}')
    if x > max_x:
        print(f'Exceeds max {x} > {min_x}')
    if y < min_y:
        print(f'Exceeds min {y} < {min_y}')
    if y > max_y:
        print(f'Exceeds max {y} > {min_y}')
```

Partial output:
```
Exceeds max -115.0 > -115.321
Exceeds max -115.23 > -115.321
Exceeds max -115.1989873 > -115.321
Exceeds max -115.282293711644 > -115.321
Exceeds min 20.10361238 < 20.123
Exceeds max -115.127203600243 > -115.321
Exceeds max -115.1 > -115.321
Exceeds min 20.0 < 20.123
Exceeds max -115.19968337045 > -115.321
Exceeds max -115.198 > -115.321
Exceeds max -115.12414928 > -115.321
Exceeds min 20.0 < 20.123
```

### Expected behavior

Values should be bound by the `min_value` and `max_value`

### Actual behavior

Values are exceeding the allowed boundary

mvanderlee:
### What does this change

Ensure we don't override the user input.
Validate that we're within the allowed range, and if not, ensure we are with some randomness. 

### What was wrong

As shown in #1825, `pyfloat` would reliably produce results that exceeded the provided values.

### How this fixes it

Copy the arguments and modify the copy instead of the original arguments.
Check that we're in the required range, and if not, ensure we are.

Fixes #1825.

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -144,20 +144,23 @@ class Provider(BaseProvider):
 
         sign = ""
         if (min_value is not None) or (max_value is not None):
+            # Copy values to ensure we're not modifying the original values and thus going out of bounds
+            left_min_value = min_value
+            left_max_value = max_value
             # Make sure left_digits still respected
             if left_digits is not None:
                 if max_value is None:
-                    max_value = 10**left_digits  # minus smallest representable, adjusted later
+                    left_max_value = 10**left_digits  # minus smallest representable, adjusted later
                 if min_value is None:
-                    min_value = -(10**left_digits)  # plus smallest representable, adjusted later
+                    left_min_value = -(10**left_digits)  # plus smallest representable, adjusted later
 
             if max_value is not None and max_value < 0:
-                max_value += 1  # as the random_int will be generated up to max_value - 1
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

PR #1827, opened 2023-03-21
Fix value error formatting

dariocurr:
### What does this change

Fix the `ValueError` formatting

### What was wrong

The error was not printed 

### How this fixes it

I literally add just a char to format the message error (my mistake, I forgot it in #1822)

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -87,7 +87,7 @@ class Provider(BaseProvider):
         elif object_type == dict:
             return self.pydict()
         else:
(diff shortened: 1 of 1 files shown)


stefan6419846:
We might want to cover this by a corresponding test to avoid such an issue.

dariocurr:
Done with latest commit 331498d

dariocurr:
I have no idea why the tests fail

fcurella:
@dariocurr Looks like just a fluke, I've re-run that test for you

fcurella (approved):
Thank you!

PR #1828, opened 2023-03-21
Separate freely-associated states & US territories

DavidCain:
There are three sovereign states that are members of the Compact of Free Association. These states are not US territories, and having `state_abbr()` include them by default can cause problems with other tooling.

For the purposes of a valid deliverable address, it's useful to include these freely-associated states by default. But it's also helpful to be able to dynamically exclude them. (Furthermore, there's good reason to simply be correct about how these sovereign states are referred to).

-----

### What does this change

Exposes a custom boolean (no change in default behavior) which allows
exclusion of freely-associated states from US states & territories.

### What was wrong

We'd categorized these sovereign states as US territories.

### How this fixes it

An optional param to `state_abbr()` allows exclusion of these freely-associated
states while still preserving existing functionality.

### DIFF
--- a/faker/providers/address/en_US/__init__.py
+++ b/faker/providers/address/en_US/__init__.py
@@ -1,5 +1,5 @@
 from collections import OrderedDict
-from typing import Optional
+from typing import Optional, Set
 
 from ..en import Provider as AddressProvider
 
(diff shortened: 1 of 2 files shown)


fcurella:
cc @johnekent

fcurella (approved):
Thank you!

DavidCain:
Wow, thank _you_ for the speedy review, @fcurella ! Will push a quick commit to fix `black`.

PR #1829, opened 2023-03-23
Fix determinism in `state_abbr()`

DavidCain:
Unfortunately, by changing the code to use sets (unordered by definition) from simple tuples, I broke the determinism of `state_abbr()`. Apologies! Changing it back to use a simple ordered collection fixes the problem.

(This probably makes sense to ship as 18.3.1, since there's no change in the API)


fcurella (approved):
Thank you!

PR #1832, opened 2023-03-27
Update README.rst

daisyye0730:
updated "seeding generator section"

### What does this change

Brief summary of the changes.

### What was wrong

In the "Seeding the Generator section", the following sentence sounds awkward "Calling the same methods with the same version of faker and seed produces the same results."

### How this fixes it

Rephrased it to "Seed produces the same result when the same methods with the same version of faker are called."


fcurella (approved):
Thank you! This way it sounds so much better!

PR #1834, opened 2023-03-27
Working hours optional parameter

### LINKED ISSUE #1775: Suggestion: Generate Date Time values only for at certain times (e.g. Workdays, Weekends)
Hi, 

Using `fake.date_time()` or `fake.date_time_between()` return values that are quite unrealistic for certain purposes. For my usecase I would love to have a feature that restrict the returned dates to only working days (mon-fri), maybe even only typical working hours (e.g. 08:00-17:00). 

I could imagine that as an optional parameter in the function call: `fake.date_time_between(filter="weekdays")`

Other filters could be "weekend", "monday", "nighttime", "daytime" etc.

ulasonat:
### What does this change

This change adds an optional `is_working_hours` Boolean parameter to the date_time-related functions in Faker. When set to `True`, the generated datetime objects will be restricted to working hours (08:00-17:00) and weekdays. If set to `False` or not used, the functions work as before.

### What was wrong

The original issue reported on GitHub (https://github.com/joke2k/faker/issues/1775) mentioned the need for an additional parameter to control the generated datetime objects to be within working hours and weekdays.

### How this fixes it

The proposed changes include adding the `is_working_hours` parameter to the relevant functions and adjusting the datetime generation logic based on its value. A new helper function `adjust_random_date_to_working` is introduced to handle the required adjustments for working hours and weekdays.

Fixes #1775.

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -7,6 +7,7 @@ from datetime import datetime
 from datetime import time as dttime
 from datetime import timedelta
 from datetime import tzinfo as TzInfo
+from datetime import timezone
 from typing import Any, Callable, Dict, Iterator, Optional, Tuple, Union
 
 from dateutil import relativedelta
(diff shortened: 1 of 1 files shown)


stefan6419846:
Instead of saying `is_working_hours` with fixed values, could we make this a time range instead, so users can decide which range they are interested in? Then this functionality could be used for multiple purposes, resolving my concerns about the not always clear definition of working hours as well.

Additionally, it seems like the functionality remains untested for now.

By the way: The issue mentions weekday limits as well.

fcurella:
I'm afraid this functionality might soon get too complicated:

* working hours and working days might differ across Countries
* lots of countries don't even have the concept of working hours
* Some countries have different working hours for different part sof the country (eg: good luckk finding anything open in Milan, Italy before 10am).
* Some countries have different working hours _and_ working days depending on the type of business, eg a food stores vs an office

My worry is that we start we the simple 8-5pm use case, and then we'll get PR after PR asking for new functionality. Perhaps this would be better served by writing a Community Provider?

PR #1836, opened 2023-03-31
currency_symbol

### LINKED ISSUE #1835: Some currency symbols missing from currency provider
* Faker version: v4.0.1 onwards
* OS: Archlinux

Some currencies do not have any symbols associated with them.

### Steps to reproduce

```
from faker import Faker
fake = Faker()
currency = fake.currency()  # returns ("GGP", "Guernsey pound")
currency = ("GGP", "Guernsey pound")
symbol = fake.currency_symbol(code=currency[0])
print(symbol)
```

### Expected behavior

```
# should print the pound sign
'£'
```

### Actual behavior

```
faker/providers/currency/__init__.py", line 326, in currency_symbol
    return self.currency_symbols[code]
 KeyError: 'GGP'
```

evoludigit:
### What does this change

- adds symbol for all listed currencies in the currency symbol dictionary
- ¤ is chosen as default when the provided currency code does not have a corresponding symbol.

### What was wrong

KeyError raised when the currency_symbol function was provided with a non existing currency code.

### How this fixes it

- completion of the currency symbol dictionary
- default currency symbol when code is not present in the currency symbol dictionary

Fixes #1835


stefan6419846:
Could you please add corresponding tests as well?

PR #1838, opened 2023-04-02
Currency symbol completion

### LINKED ISSUE #1835: Some currency symbols missing from currency provider

evoludigit:
*Note: I preferred to close the previous PR and create a new pull request so as to have everything under a single commit.*

### What does this change

- adds symbol for all listed currencies in the currency symbol dictionary
- ¤ is chosen as default when the provided currency code does not have a corresponding symbol.

### What was wrong

- some currency_code were missing from the currency_symbols dictionary
- no default value provided by the currency_symbol function, thus raising a KeyError

### How this fixes it

- completion of the currency symbol dictionary
- default currency symbol when code is not present in the currency_symbol dictionary
- a KeyError is still raised when the currency_symbol function is called with a non valid currency code

Fixes #1835


fcurella (changes requested):
Could run `make lint` to fix the formatting? issues?

evoludigit:
Yes, sorry I am using precommit in my projects and forgot to install black / flake8 / mypy and run make lint before commiting. I am not sure however if this is sufficient however.

PR #1840, opened 2023-04-04
fix: use correct union type for provider argument

### LINKED ISSUE #1839: Incorrect type annotation on Generator.add_provider
* Faker version: 18.3.2
* OS: MacOS

The type annotation on `Generator.add_provider` is needlessly restrictive - the implementation clearly supports passing either a type or an instance, but the type annotation only allows instances:

```python
    # faker/generator.py
    def add_provider(self, provider: "BaseProvider") -> None:
        if isinstance(provider, type):
            provider = provider(self)
```

### Steps to reproduce

1. Write code that adds a provider class to a `Generator`
2. Run mypy

```python
class MyCustomProvider(BaseProvider):
    pass

fake = Factory.create("en_US")
fake.add_provider(MyCustomProvider)
```

### Expected behavior

`mypy` check should pass

### Actual behavior

`mypy` produces an error:

```
error: Argument 1 to "add_provider" of "Generator" has incompatible type "Type[MyCustomProvider]"; expected "BaseProvider"  [arg-type]
```

DanielSchaffer:
### What does this change

The type annotation on `Generator.add_provider`

### What was wrong

See #1839

### How this fixes it

Adds the correct union type annotation

Fixes #1839

### DIFF
--- a/faker/generator.py
+++ b/faker/generator.py
@@ -1,7 +1,7 @@
 import random as random_module
 import re
 
-from typing import TYPE_CHECKING, Any, Callable, Dict, Hashable, List, Optional
+from typing import TYPE_CHECKING, Any, Callable, Dict, Hashable, List, Optional, Type, Union
 
 from .typing import SeedType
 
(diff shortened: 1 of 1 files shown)


fcurella (approved):
Thank you!

DanielSchaffer:
@fcurella No problem! I don't understand where that doc8 error is coming from, looking at the diff, I'm not seeing any trailing whitespace 🤔

fcurella:
It wasn't you, it was introduced in https://github.com/joke2k/faker/commit/5d4e3785c8d84fdf808a771c34f3e0f62db1c2ca.

I'll fix it on `master` and you can rebase against it

PR #1841, opened 2023-04-04
Formatting modern

MarquisTheCoder:
### What does this change

I changed the formatting inside the isinstance function

### What was wrong

the formatting wasnt modern python and could be improved 

### How this fixes it
uses string interpolation
Fixes #...


fcurella (approved):
Thank you!

Issue #1824, opened 2023-03-19
The pydecimal function raises exception when using decimal places.


mvanderlee:
* Faker version: 17.6.0
* OS: Windows 10

Brief summary of the issue goes here.

The `pydecimal` function raises exception when using decimal places.

### Steps to reproduce

1. 
```python
from faker import Faker
fake = Faker()
fake.pydecimal(min_value=-122.190402, max_value=0, right_digits=6)

>> ValueError: non-integer stop for randrange()
```

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
File [/~/.pyenv/pyenv-win/versions/3.11.1/Lib/random.py:317), in Random.randrange(self, start, stop, step)
    316 try:
--> 317     istop = _index(stop)
    318 except TypeError:

TypeError: 'float' object cannot be interpreted as an integer

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
Cell In[80], line 1
----> 1 fake.pydecimal(min_value=-122.190402, max_value=0, right_digits=6)

File [\.venv\Lib\site-packages\faker\providers\python\__init__.py:262](/.venv/Lib/site-packages/faker/providers/python/__init__.py:262), in Provider.pydecimal(self, left_digits, right_digits, positive, min_value, max_value)
    260 else:
    261     if min_value is not None:
--> 262         left_number = str(self.random_int(max(max_value or 0, 0), abs(min_value)))
    263     else:
    264         min_left_digits = math.ceil(math.log10(abs(min(max_value or 1, 1))))

File [\.venv\Lib\site-packages\faker\providers\__init__.py:316](/.venv/Lib/site-packages/faker/providers/__init__.py:316), in BaseProvider.random_int(self, min, max, step)
    306 def random_int(self, min: int = 0, max: int = 9999, step: int = 1) -> int:
    307     """Generate a random integer between two integers ``min`` and ``max`` inclusive
    308     while observing the provided ``step`` value.
    309 
   (...)
    314     :sample: min=0, max=15, step=3
    315     """
--> 316     return self.generator.random.randrange(min, max + 1, step)

File [~\.pyenv\pyenv-win\versions\3.11.1\Lib\random.py:323](~/.pyenv/pyenv-win/versions/3.11.1/Lib/random.py:323), in Random.randrange(self, start, stop, step)
    320     if istop != stop:
    321         _warn('randrange() will raise TypeError in the future',
    322               DeprecationWarning, 2)
[...]

PR #1842, opened 2023-04-05
1824 fix min max

MarquisTheCoder:
### What does this change

I rounded the min/max values in the pydecimal function definition

```python

# line: 254 file:providers/python/__init__.py
def pydecimal(
        self,
        left_digits=None,
        right_digits=None,
        positive=False,
        min_value=None,
        max_value=None,
    ):
        """ returns valid decimal when given floating point numbers without changing too much code"""
        min_value = round(min_value)
        max_value = round(max_value)

```
### What was wrong

As stated in issue 1824 when a developer inserts a floating point into the min or max value 
this conflicts with pythons in-built randrange function. 
https://github.com/joke2k/faker/issues/1824

### How this fixes it

simple fix just rounds the number of the min or max value because they don't need
to be floating points for this function to work in an consistent manner. 

Fixes #...
1824


stefan6419846:
Could you please add a corresponding test as well and fix the formatting issues detected by `black`?

PR #1844, opened 2023-04-10
Added es_PE person

JsonVladimir:
Added es_PE to person, I recently used the library and noticed that there were no common names from Peru, I decided to propose that the file be added.


fcurella (changes requested):
Could you add some test cases for the new providers?

fcurella (changes requested):
Can yuou fix the remaining linting issue? You can run `make lint` to format using the project's settings.

JsonVladimir:
> Can yuou fix the remaining linting issue? You can run `make lint` to format using the project's settings.

When I run black it tells me there is 1 file that needs to be reformatted but when i format it it tells me no changes were made.

```
C:\GitHub\faker> python -m black --check --line-length 120 C:\GitHub\faker
would reformat C:\GitHub\faker\tests\providers\test_phone_number.py

Oh no! 💥 💔 💥
1 file would be reformatted, 683 files would be left unchanged.
C:\GitHub\faker> python -m black C:\GitHub\faker\tests\providers\test_phone_number.py
**All done! ✨ 🍰 ✨
1 file _left unchanged_.**
```
image
image

fcurella:
@JsonVladimir I'm not seeing `--line-length 120` on the second command you called. You can just run `make lint` and it should run it with the correct arguments for you

tests/providers/test_bank.py
```diff
@@ -241,6 +242,56 @@ def test_clabe_bank_code(self, faker, num_samples):
...
+    def test_bban(self, faker, num_samples):
+        for _ in range(num_samples):
+            assert re.fullmatch(r"[A-Z]{4}\d{18}", faker.bban())
+
+    def test_iban(self, faker, num_samples):
+        for _ in range(num_samples):
+            iban = faker.iban()
+            assert is_valid_iban(iban)
+            assert iban[:2] == EsPeBankProvider.country_code
+            assert re.fullmatch(r"\d{2}[A-Z]{4}}\d{18}", iban[2:])
```
fcurella, line 257: I think I see an extra `}` in `{4}}`

fcurella (changes requested):
Could you check the failing tests?

tests/providers/test_phone_number.py
```diff
@@ -285,6 +285,21 @@ def test_phone_number(self, faker, num_samples):
...
+
+    def test_phone_number(self, faker, num_samples):
+        pattern: Pattern = re.compile(
+            r"(\+?51|\(\+51\))?(9\d{8})|"
+            r"(\+?51|\(\+51\))?((9\d{2}) \d{3} \d{3})|"
+            r"(9\d{8})|"
+            r"((9\d{2}) \d{3} \d{3})"
+        )
+        for _ in range(num_samples):
+            phone_number = faker.phone_number_pe()
```
fcurella, line 299: the method is called `phone_number()`:

```suggestion
            phone_number = faker.phone_number()
```

fcurella (changes requested):

tests/providers/test_phone_number.py
```diff
@@ -285,6 +285,12 @@ def test_phone_number(self, faker, num_samples):
             assert pattern.fullmatch(phone_number)
 
 
+class TestEsPE:
+    """Test es_PE phone number provider methods"""
+
+    pass
```
fcurella, line 291: Can we actually test something? :P

fcurella (changes requested):

Issue #1847, opened 2023-04-18
Suggestion: create simple_cellphone_number format


dlwrnc:
* Faker version: faker 18.4.0
* OS: mac os monterey

### Ideal behavior

Ideal behavior would be a situation where it's possible to create only phone numbers of format ##########. I've found a few forms that are built to only expect a "basic" ten-digit phone number. 

### Actual behavior

Actual behavior is that there's no way to define the ideal phone number format so narrowly in the existing code.

My suggested change is to the en_US file, adding a format type like:

`simple_format =  ('##########', '###-###-####')`

and then a 

```
    def simple_cellphone_number(self) -> str:
        pattern: str = self.random_element(self.simple_format)
        return self.numerify(self.generator.parse(pattern))
```

happy to make this change myself. :)

PR #1848, opened 2023-04-22
add a separate basic number format for US

dlwrnc:
issue ticket: https://github.com/joke2k/faker/issues/1847

### What does this change

Adds a 'basic phone number' type to en_US to comply with some form requirements, as well as a unit test.

### What was wrong

Some forms do not allow for phone formats with extensions or country codes. As a result, the faker get_phone function for en_US doesn't work for all cases. 

### How this fixes it

Adds a function that allows the user to only use 'basic' ten-digit phone numbers and standard formatting.


dlwrnc:
@fcurella thanks for the catch -- linting complete.

PR #1851, opened 2023-04-24
Fix zh_TW ssn provider rule

cyanghsieh:
### What does this change
Implement current rule of generating ssn in locale zh_TW.

### What was wrong

The rule in the repo has not updated for years and was wrong. Gender and checksum were not considered.

### How this fixes it

Based on the public information of the generating rules and implement so.

### DIFF
--- a/faker/providers/ssn/zh_TW/__init__.py
+++ b/faker/providers/ssn/zh_TW/__init__.py
@@ -1,43 +1,47 @@
 from .. import Provider as SsnProvider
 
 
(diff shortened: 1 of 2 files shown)


cyanghsieh:
add test_ssn and apply Black-formatting based on the last check result

fcurella:
Don't worry about the coverall test. It was set up to be too strict, and I don't have a way to change it

fcurella (approved):

PR #1852, opened 2023-04-24
Improve fr_FR provider postcode

tonial:
### What does this change

Improve postcode for fr_FR address provider

### What was wrong

The current provider may generate postcodes for invalid departments (any number starting with 96 is invalid)

### How this fixes it

Generate postcode from department codes

NB: I'm not sure how to test this commit : I could check that the generated postcode always match a department, but the test would be flaky before the fix...

### DIFF
--- a/faker/providers/address/fr_FR/__init__.py
+++ b/faker/providers/address/fr_FR/__init__.py
@@ -46,7 +46,6 @@ class Provider(AddressProvider):
     address_formats = ("{{street_address}}\n{{postcode}} {{city}}",)
 
     building_number_formats = ("%", "%#", "%#", "%#", "%##")
-    postcode_formats = ("#####",)
     countries = (
         "Afghanistan",
         "Afrique du sud",
@@ -467,3 +466,13 @@ class Provider(AddressProvider):
         :example: '59'
         """
         return self.department()[0]
+
+    def postcode(self) -> str:
+        """
+        Randomly returns a postcode generated from existing french depertment number.
+        exemple: '33260'
+        """
+        department = self.department_number()
+        if department in ["2A", "2B"]:
+            department = "20"
+        return f"{department}{self.random_number(digits=5 - len(department))}"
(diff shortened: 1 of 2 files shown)


faker/providers/address/fr_FR/__init__.py
```diff
@@ -467,3 +466,12 @@ def department_number(self) -> str:
...
         """
         return self.department()[0]
+
+    def postcode(self) -> str:
+        """
+        Randomly returns a postcode generated from existing french depertment number.
+
+        exemple: '33260'
+        """
+        department = self.department()
```
fcurella, line 476: did you mean to call `self.department_number()`?

faker/providers/address/fr_FR/__init__.py
```diff
@@ -467,3 +466,12 @@ def department_number(self) -> str:
...
+        return department + self.random_number(digits=5 - len(department))
```
fcurella, line 477: if we're concatenating strings, I'd rather use string formatting (f-strings are fine). Just to make it more evident.

fcurella (changes requested):

faker/providers/address/fr_FR/__init__.py
tonial: :facepalm:

faker/providers/address/fr_FR/__init__.py
tonial: Done !

fcurella:
we could just check that the first 2 characters are a valid department code? Better than nothing :)

tonial:
I added the check.
And thanks to it, I noticed I previously made a mistake as if the department code is 2A or 2B, the postcode starts with 20....

fcurella (commented):
Could you fix the linting issue?

fcurella (changes requested):
Could you fix the linting issue?

tonial:
I fixed it :)

fcurella (approved):
Thank you!

PR #1857, opened 2023-05-03
Add zh_TW automotive license_plate rules and test functions

cyanghsieh:
### What does this change

add zh_TW automotive license_plate rules

### What was wrong

was not implemented ever. 

### How this fixes it

will generate the fake license plate data in zh_TW locale

Fixes #...


fcurella (approved):
Thank you!

PR #1858, opened 2023-05-03
Add zh_CN automotive license_plate rules and test functions

cyanghsieh:
### What does this change

add zh_CN automotive license_plate rules

### What was wrong

Was not implemented before.

### How this fixes it

will generate the fake license plate data in zh_CN locale


fcurella (approved):
Thank you!

Issue #1853, opened 2023-04-25
Suggestion: Fake Passport, Driving License and County


psmukherjee009:
* Faker version:
* OS:

Currently I needed Fake Passport, Driving License and County and didn't find it in Faker. Can we have these?

### Steps to reproduce

1. step 1
1. step 2
1. step 3

### Expected behavior

X should be ...

### Actual behavior

X is ...

stefan6419846:
You can probably try to implement this yourself and create a PR, but there is no existing implementation yet.

PR #1860, opened 2023-05-03
Passport Provider and EN_US Implementation

llw2128:
### What does this change
Added a passport faking functionality 

### What was wrong
#1853
Addressed one of the desired functionalities

### How this fixes it
This adds a new provider that fakes passport data, specifically numbers, issue and expiry dates, and birthdays

### DIFF
--- a/faker/providers/passport/__init__.py
+++ b/faker/providers/passport/__init__.py
@@ -1,3 +1,4 @@
+import datetime
 import re
 
 from string import ascii_uppercase
@@ -12,12 +13,12 @@ class Provider(BaseProvider):
 
     passport_number_formats: ElementsType = ()
 
-    def passport_DOB(self):
-        """Generate a datetime date of birth."""
+    def passport_dob(self) -> datetime.date:
+        """Generate a datetime date of bisrth."""
         birthday = self.generator.date_of_birth()
         return birthday
 
-    def passport_owner(self, gender="X"):
+    def passport_owner(self, gender: str = "X") -> tuple[str, str]:
         """Generate a given_name and surname for a passport owner
         The ``gender`` argument is sex of a passport owner which is a one character string
         that is either male, female, or non-binary.
--- a/faker/providers/passport/__init__.py
+++ b/faker/providers/passport/__init__.py
@@ -2,6 +2,7 @@ import datetime
 import re
 
 from string import ascii_uppercase
+from typing import Tuple
 
 from .. import BaseProvider, ElementsType
 
@@ -18,9 +19,9 @@ class Provider(BaseProvider):
         birthday = self.generator.date_of_birth()
         return birthday
 
-    def passport_owner(self, gender: str = "X") -> tuple[str, str]:
+    def passport_owner(self, gender: str = "X") -> Tuple[str, str]:
         """Generate a given_name and surname for a passport owner
-        The ``gender`` argument is sex of a passport owner which is a one character string
+        The ``gender`` argument is the gender marker of a passport owner, which is a one character string
         that is either male, female, or non-binary.
         """
         if gender == "M":
--- a/faker/providers/passport/__init__.py
+++ b/faker/providers/passport/__init__.py
@@ -1,0 +1,30 @@
+import re
+
+from string import ascii_uppercase
+
+from .. import BaseProvider, ElementsType
+
+localized = True
+
+
+class Provider(BaseProvider):
+    """Implement default Passport provider for Faker."""
+    nationality: ElementsType = ()
+    passport_number_formats: ElementsType = ()
+
+    def nationality(self)-> str:
+        return self.nationality
+    
+    def passport_owner(self):
+        given_name = self.generator.parse("{{first_name}} ")
+        surname  = self.generator.parse("{{last_name}} ")
+        return given_name, surname
+    
+    def passport_number(self) -> str:
+        """Generate a passport number"""
+        temp = re.sub(
+            r"\?",
+            lambda x: self.random_element(ascii_uppercase),
+            self.random_element(self.passport_number_formats),
+        )
+        return self.numerify(temp)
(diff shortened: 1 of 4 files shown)


faker/providers/passport/en_US/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+
+    DEFAULT_DAY = date.today()
+    passport_number_formats = (
+        # NGP
+        "?########",
+        # Pre-NGP
+        "#########",
+    )
+
+    def passport_Dates(self, birthday=DEFAULT_DAY):
```
fcurella, line 25: we use lowercase-only method names

faker/providers/passport/en_US/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+class Provider(PassportProvider):
+    """Implement passport provider for ``en_US`` locale.
+
+    Sources:
+
+    - https://travel.state.gov/content/travel/en/passports/passport-help/next-generation-passport.html
+    - https://www.vitalrecordsonline.com/glossary/passport-book-number
+    """
+
+    DEFAULT_DAY = date.today()
```
fcurella, line 17: I'm afraid this would only get evaluated at declaration time, and not at runtime, resulting in the wrong value. Eg imagine the typical webserver scenario, where the process starts and lasts for days. The `DEFAULT_DAY` would be stuck at the day the webserver was launched.

faker/providers/passport/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+
+localized = True
+
+
+class Provider(BaseProvider):
+    """Implement default Passport provider for Faker."""
+
+    passport_number_formats: ElementsType = ()
+
+    def passport_DOB(self):
```
fcurella, line 15: can we add type hinting?

faker/providers/passport/en_US/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+
+        Sources:
+
+        -https://travel.state.gov/content/travel/en/passports/passport-help/faqs.html
+        """
+        birth_date = birthday.strftime("%d ") + birthday.strftime("%b ") + birthday.strftime("%Y")
+        today = date.today()
+        age = (today - birthday).days // 365
+        if age < 16:
+            EXPIRY_YEARS = 5
```
fcurella, line 40: we can have `expiry_years` lowercase.

faker/providers/passport/en_US/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+            issue_date_fromat = issue_date.strftime("%d ") + issue_date.strftime("%b ") + issue_date.strftime("%Y")
+            expiry_date_format = expiry_date.strftime("%d ") + expiry_date.strftime("%b ") + expiry_date.strftime("%Y")
+            return birth_date, issue_date_fromat, expiry_date_format
+
+        else:
+            EXPIRY_YEARS = 10
+            issue_date = self.generator.date_time_between(today - timedelta(days=EXPIRY_YEARS * 365 - 1), today)
+            age_issue = (issue_date.date() - birthday).days // 365
+
+            if age_issue < 16:
```
fcurella, line 62: You're already catching `age < 16` in the first case of this`if`. I'm not sure how we would end up in this branch here.

faker/providers/passport/en_US/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+
+        -https://travel.state.gov/content/travel/en/passports/passport-help/faqs.html
+        """
+        birth_date = birthday.strftime("%d ") + birthday.strftime("%b ") + birthday.strftime("%Y")
+        today = date.today()
+        age = (today - birthday).days // 365
+        if age < 16:
+            EXPIRY_YEARS = 5
+            issue_date = self.generator.date_time_between(today - timedelta(days=EXPIRY_YEARS * 365 - 1), today)
+            if age < 5:
```
fcurella, line 42: can you add a comment expaining why we need to special-case less-than-5yos?

fcurella (changes requested):

faker/providers/passport/en_US/__init__.py
llw2128: Reworked logic on this, updated to remove unnecessary conditions

faker/providers/passport/en_US/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+                expiry_years = 10
+
+            expiry_date = issue_date.replace(year=issue_date.year + expiry_years)
+
+            issue_date_fromat = issue_date.strftime("%d ") + issue_date.strftime("%b ") + issue_date.strftime("%Y")
+            expiry_date_format = expiry_date.strftime("%d ") + expiry_date.strftime("%b ") + expiry_date.strftime("%Y")
+            return birth_date, issue_date_fromat, expiry_date_format
+
+    def passport_gender(self, seed: int = 0) -> str:
+        """Generates a string representing the sex displayed on a passport
```
fcurella, line 75:
```suggestion
        """Generates a string representing the gender displayed on a passport
```

faker/providers/passport/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+    passport_number_formats: ElementsType = ()
+
+    def passport_dob(self) -> datetime.date:
+        """Generate a datetime date of bisrth."""
+        birthday = self.generator.date_of_birth()
+        return birthday
+
+    def passport_owner(self, gender: str = "X") -> tuple[str, str]:
+        """Generate a given_name and surname for a passport owner
+        The ``gender`` argument is sex of a passport owner which is a one character string
```
fcurella, line 23:
```suggestion
        The ``gender`` argument is the gender marker of a passport owner, which is a one character string
```

fcurella (approved):
Thank you! 🏳️‍🌈

PR #1864, opened 2023-05-11
adding state_code and state API

### LINKED ISSUE #1845: Suggestion: Address provider should generate state (where appropriate)
* Faker version: 18.4.0
* OS: macOS 13.2.1

The standard [address.Provider](https://faker.readthedocs.io/en/master/providers/faker.providers.address.html) doesn’t have a `state()` or `state_code()` method which would be useful (for countries which support that).

### Steps to reproduce

n/a

### Expected behavior

n/a

### Actual behavior

n/a

Sbhat92:
### What does this change

Adds state_code() API to address provider in en_US and state_name() in en_IN

### What was wrong

Address provider did not have state() and state_code() APIs

### How this fixes it
Adds the above API and corresponding test cases

Fixes #1845

### DIFF
--- a/tests/providers/test_address.py
+++ b/tests/providers/test_address.py
@@ -631,6 +631,12 @@ class TestEnUS:
         first = faker.state_abbr()
         faker.seed_instance(0)
         assert faker.state_abbr() == first
+    
+    def test_state_code(self, faker, num_samples):
+        for _ in range(num_samples):
+            state_code = faker.state_code()
+            assert not isinstance(state_code, str)
+            assert state_code in EnUsAddressProvider.states_abbr
 
 
 class TestEsCo:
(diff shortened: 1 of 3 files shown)


fcurella (changes requested):
Thank you! Could you run `make lint` to fix the formatting issues?

tests/providers/test_address.py
```diff
@@ -632,6 +632,12 @@ def test_state_abbr_determinism(self, faker):
         faker.seed_instance(0)
         assert faker.state_abbr() == first
 
+    def test_state_code(self, faker, num_samples):
+        for _ in range(num_samples):
+            state_code = faker.state_code()
+            assert not isinstance(state_code, str)
```
fcurella, line 638: Shouldn't `state_code` BE a string?

fcurella (changes requested):
Can you fix the new test?

weedySeaDragon:
This would be awesome to have! @Sbhat92 -- do you have the time to finish this?

PR #1867, opened 2023-05-26
PoC: a .optional proxy that may or may not return a fake value

ligne:
### What does this change

This adds a `.optional` property, that should make all the usual functions return either a fake value or `None`, with a customisable probability.

This was a quick and dirty lunchtime project which could use some care and attention (in particular I'm not sure about the `prob=n` parameter, and the tests are a bit limited). I thought I'd get some feedback before spending too much time cleaning it up.

### What was wrong

Currently it can be a bit fiddly to create fake optional fake data. There are a few projects out there that extend `faker` to provide this functionality using custom providers, but they're limited in what they can generate. This provides access to everything that would normally be available.

### DIFF
--- a/faker/proxy.py
+++ b/faker/proxy.py
@@ -356,7 +356,11 @@ class OptionalProxy:
         @functools.wraps(function)
         def wrapper(*args: Any, prob: float = 0.5, **kwargs: Any) -> Optional[RetType]:
             if not 0 < prob <= 1.0:
-                raise ValueError()
-            return function(*args, **kwargs) if self._proxy.pybool() else None
+                raise ValueError("prob must be between 0 and 1")
+            return (
+                function(*args, **kwargs)
+                if self._proxy.boolean(chance_of_getting_true=int(prob * 100))
+                else None
+            )
 
         return wrapper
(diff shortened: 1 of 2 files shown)


ligne:
Sorry for the delay! I was out of the country and it took me a bit of time to catch up with things.

I've taken a slightly roundabout route because i assume you don't really want to have to pull in `typing_extensions` just for ParamSpec.

I've also fixed it so it the `prob` parameter is actually honoured!

ligne:
Done! I'd run black, but with the wrong line length. Sorry about that!

faker/proxy.py
```diff
@@ -322,3 +329,34 @@ def wrapper(*args, **kwargs):
             return retval
 
         return wrapper
+
+
+class OptionalProxy:
+    def __init__(self, proxy: Faker):
```
fcurella, line 335:
```suggestion
class OptionalProxy:
    """
    Return either a fake value or None, with a customizable probability.
    """
    def __init__(self, proxy: Faker):
```

fcurella (approved):
Thank you!

PerchunPak:
Haven't you considered also adding a `__call__` method for allowing passing custom callable or pure values? Similar to how it's done in my library https://github.com/PerchunPak/optional-faker#example.

PR #1869, opened 2023-05-31
Fix fr_FR postcode length

### LINKED ISSUE #1868: fr_FR address provider generates less than 5 digits postcode
* Faker version: master (18.9.0), bug introduced by https://github.com/joke2k/faker/pull/1852

The new provider method can generate 4 or less digits values. But french postcodes must be 5 digits long (cf https://en.wikipedia.org/wiki/Postal_codes_in_France).

Line to fix: https://github.com/joke2k/faker/blob/fa79c6b0c510274b08b46ad0f007fd1b517a9ab2/faker/providers/address/fr_FR/__init__.py#LL478C27-L478C27

### Steps to reproduce

1. generate a few postcodes with `fr_FR` address provider

### Expected behavior

5 digits values

### Actual behavior

From 3 to 5 digits values

vmttn:
### What does this change

Fix the length of the postcode values generated by the fr_FR address provider

### What was wrong

cf #1868 

### How this fixes it

Pass `fix_len` flag to the underlying `BaseProvider.random_number()` call

Fixes #1868

### DIFF
--- a/faker/providers/address/fr_FR/__init__.py
+++ b/faker/providers/address/fr_FR/__init__.py
@@ -469,7 +469,7 @@ class Provider(AddressProvider):
 
     def postcode(self) -> str:
         """
-        Randomly returns a postcode generated from existing french depertment number.
+        Randomly returns a postcode generated from existing french department number.
         exemple: '33260'
         """
         department = self.department_number()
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

PR #1875, opened 2023-06-12
New Identification Numbers for several locales.

Nikomixo:
### What does this change

Adds the following:

### ssn/de_AT
* TIN
* SSN

### ssn/de_DE
* TIN

### ssn.sl_SI
* TIN

Tests for all of the above.
Renamed `ssn/dk_DK` to `ssn/da_DK` to match with other locales.

### What was wrong

Some locale SSN providers were missing various forms of identification numbers. 

### How this fixes it

Adds a few missing providers.


fcurella (changes requested):
Can you fix the linting issues?

PR #1876, opened 2023-06-15
Update providers/phone_number/uk_UA/__init__.py

lexxai:
https://countrycode.org/ukraine

https://uk.wikipedia.org/wiki/%D0%A2%D0%B5%D0%BB%D0%B5%D1%84%D0%BE%D0%BD%D0%BD%D0%B8%D0%B9_%D0%BD%D0%BE%D0%BC%D0%B5%D1%80

Ukraine Country Code +380

### What does this change

Replaced the country code with the real value for this locale.

### What was wrong

The country code used is +38, which is not currently assigned to any country.

### How this fixes it

Replace the +38 code with +380.

Fixes #...

The +38 code has been replaced by +380.


fcurella (approved):
Thank you!

Issue #1849, opened 2023-04-22
Current US Phone number formatting allows for generation of non-real numbers.


dlwrnc:
* Faker version: 18.4.0
* OS: mac os

### Expected behavior

Per the current NANP (https://en.wikipedia.org/wiki/North_American_Numbering_Plan), US 10-digit phone numbers use the following format:

> The NANP numbering format may be summarized in the ten-digit notation NXX NXX-XXXX, where the Ns denote any of the digits 2–9, and the Xs denote any digit 0–9. 

### Actual behavior

Currently, the US phone number generation does not follow the convention of the Ns being 2-9, and can instead be 0-9. 

My proposed fix is to modify the [numerify](https://github.com/joke2k/faker/blob/master/faker/providers/__init__.py#L600) function to add a symbol to denote 2-9, and then modify the US phone number formatting as needed to use that symbol. I think the most difficult part of this modification is determining a good symbol candidate to serve as the 2-9 representation without messing up existing usages -- happy to take suggestions on this. 

EDIT: it looks like $ could work -- I'm going to make the modification and run it against the existing tests.

PR #1878, opened 2023-06-15
add digit above two type to numerify function

dlwrnc:
Issue ticket: https://github.com/joke2k/faker/issues/1849

### What does this change

Adds a digit type for values 2-9 to comply with US telephone number standards.

### What was wrong

US telephone number generation was creating impossible numbers. 

### How this fixes it

Adds a 2-9 type to numerify, adds testing to validate, ports the US phone numbers over to the new format.


dlwrnc:
Coverage is the same, not sure why coveralls is upset -- I think it's fine as-is, but happy to make modifications if needed.

fcurella (approved):
Thank you!

PR #1882, opened 2023-06-24
docs: update `pystr` docstring to include param descriptions, fix docstring formats

### LINKED ISSUE #1877: `fake.pystr()` returns unexpected length when using `max_chars` with `prefix` | `suffix`
* Faker version: 18.10.1
* OS: ubuntu 22.04

Love this library, thanks for the hard work!

I came across a situation today where strings returned by `fake.pystr(...)` are not what I would have expected.  [This line](https://github.com/joke2k/faker/blob/d193bba77917f28504a036d1c11e0c517cc61373/faker/providers/python/__init__.py#L126) adds the prefix and/or suffix to a string with length already established according to `max_chars`/`min_chars`.  Because of this, the resulting string is longer than expected when `max_chars` is used in conjunction with either `prefix` or `suffix`.

If you're open to a PR, I'd be willing to submit one to update this.  Just lemme know...thanks again!

### Steps to reproduce

```python
from faker import Faker

fake = Faker()
Faker.seed()

fake.pystr(max_chars=3, prefix="a")
fake.pystr(max_chars=3, suffix="z")
fake.pystr(max_chars=3, prefix="a", suffix="z")
```

### Expected behavior

```
aRu
TTz
aFz
```

### Actual behavior

```
aRuB
VTTz
aFmDz
```

themattmorris:
### What does this change

* Adds description of parameters to `pystr`.
* Adds newline to some docstrings that weren't rendering properly in docs.  Example:
image

### What was wrong

See #1877 

### How this fixes it

Should remove confusion around `max_chars` and `prefix`/`suffix` usage.  Docs should render appropriately when built with sphinx.

closes #1877

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -1809,6 +1809,7 @@ class Provider(BaseProvider):
         """
         Get a timestamp between January 1, 1970 and now, unless passed
         explicit start_datetime or end_datetime values.
+
         :example: 1061306726
         """
         start_datetime = self._parse_start_datetime(start_datetime)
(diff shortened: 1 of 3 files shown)


fcurella (approved):
Thank you!

PR #1883, opened 2023-06-26
Fix area codes for `fr_FR` mobile phone numbers

QuentinFchx:
### What does this change

Fixes the area code for `fr_FR` mobile phone numbers.

### What was wrong

Invalid `fr_FR` mobile phone numbers can be generated with the current implementation.
[Here is a list (in french)](https://fr.wikipedia.org/wiki/Liste_des_pr%C3%A9fixes_des_op%C3%A9rateurs_de_t%C3%A9l%C3%A9phonie_mobile_en_France) of valid French mobile phone number prefixes.
[The phonenumbers library implements a regex](https://github.com/daviddrysdale/python-phonenumbers/blob/dev/python/phonenumbers/data/region_FR.py) which _kind of_ match these rules.

### How this fixes it

It update the area codes for `fr_FR` mobile phone numbers according to this regex `(?:6(?:[0-24-8]\\d|3[0-8]|9[589])|7[3-9]\\d)\\d{6}` ([preview on debuggex](https://www.debuggex.com/r/Ke8L10UwlE5I94GI)).
As said earlier, this regex is not an exact representation of the french prefix rules, but I think it's safe to assume the `phonenumbers` library is popular enough to at least try to comply to their rules.

On a side note, would it be interesting to add tests validating generated phone numbers against the `phonenumbers` library?


fcurella (approved):
Thank you!

Issue #1879, opened 2023-06-18
Feature: Generate VIN


aliasgar55:
Can we add a feature to generate fake VIN for cars.

fcurella:
This would be a useful addition to the `automotive` provider! Feel free to submit a Pull Request!

fcurella:
Merged in https://github.com/joke2k/faker/pull/1884

PR #1884, opened 2023-06-27
Add fake automotive vin number function (#1879)

cyanghsieh:
### What does this change

base on https://en.wikipedia.org/wiki/Vehicle_identification_number to create fake vin number data function.

### What was wrong

new provider function

### How this fixes it
Note: I'm not quite familiar with pytest module, therefore I'm not quite confident on creating a new auto test flow for this vin function. Please forgive me.

### DIFF
--- a/faker/providers/automotive/__init__.py
+++ b/faker/providers/automotive/__init__.py
@@ -7,6 +7,36 @@ from .. import BaseProvider, ElementsType
 localized = True
 
 
+def calculate_str_weight(s: str, weight_factor: list) -> int:
+    """
+    multiply s(str) by weight_factor char by char
+    e.g.
+    input: s="ABCDE", weight_factor=[1, 2, 3, 4, 5]
+    return: A*1 + B*2 + C*3 + D*4 + E*5
+
+    will multiply 0 when len(weight_factor) less than len(s)
+    """
+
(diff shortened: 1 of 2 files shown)


themattmorris:
This is cool!  I am also interested in using this feature, so if needed, I'd be happy to collaborate on building the tests for the feature.

fcurella (changes requested):
I'll need tests before I can merge this.

themattmorris:
Looks like @cyanghsieh beat me to it anyways...thanks!

The only thing I wonder if is `test_vin` should be outside of class `_SimplyAutomotiveTestMixin`.  Looks like running as is will run the same exact test for every provider, as it does not depend on any attributes the same way that `test_license_plate` does.  But I'm also not familiar with the overall test strategy of this repo, and maybe this is what it should do?

fcurella:
> The only thing I wonder if is `test_vin` should be outside of class `_SimplyAutomotiveTestMixin`. Looks like running as is will run the same exact test for every provider, as it does not depend on any attributes the same way that `test_license_plate` does. But I'm also not familiar with the overall test strategy of this repo, and maybe this is what it should do?

I think that's ok for now. But we should think of a way to mark methods as locale-independent.

fcurella (approved):
Thank you!

PR #1887, opened 2023-07-07
Feature: Elector code for es_MX SSN provider

### LINKED ISSUE #1886: Feature: Elector code for es_MX ssn provider
* Faker version: 18.11.2
* OS: NA

The National Electoral Institute in Mexico issues a unique identifier code for electors, it would be great if the library could fake those.

### Steps to reproduce

NA

### Expected behavior

```python
>>> fake.elector_code()
"GLVLMR80070509M900"
```

### Actual behavior

It's not implemented 😅

edgarrmondragon:
### What does this change

Adds the elector code issued by the National Electoral Institute in Mexico.

### What was wrong

Wasn't implemented.

### How this fixes it

It implements it.

Fixes #1886

### DIFF
--- a/tests/providers/test_ssn.py
+++ b/tests/providers/test_ssn.py
@@ -631,7 +631,7 @@ class TestEsCA(TestEsES):
 
 
 class TestEsMX:
-    def setup_class(self):
+    def setup_method(self):
         self.fake = Faker("es_MX")
         Faker.seed(0)
 
--- a/tests/providers/test_ssn.py
+++ b/tests/providers/test_ssn.py
@@ -630,8 +630,8 @@ class TestEsCA(TestEsES):
         Faker.seed(0)
 
 
-class TestEsMX(unittest.TestCase):
-    def setUp(self):
+class TestEsMX:
+    def setup_class(self):
         self.fake = Faker("es_MX")
         Faker.seed(0)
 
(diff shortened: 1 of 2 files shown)


tests/providers/test_ssn.py
```diff
@@ -630,8 +630,8 @@ def setUp(self):
         Faker.seed(0)
 
 
-class TestEsMX(unittest.TestCase):
-    def setUp(self):
+class TestEsMX:
+    def setup_class(self):
```
fcurella, line 634: the reason why we are using `unittest.TestCase.setUp` is that we want to reset the `.seed` before each test. Can this be left as it is?

fcurella (changes requested):

tests/providers/test_ssn.py
edgarrmondragon: Hi @fcurella! I changed it to a pytest-style class because I wanted to use `pytest.mark.parametrize` to test the parameter options exhaustively in a slightly cleaner way.

It seems there are some options that would allow us to preserve that:

1) Using [`setup_method`](https://docs.pytest.org/en/7.4.x/how-to/xunit_setup.html#method-and-function-level-setup-teardown)
2) Using a pytest fixture. The default scope is `function`

I confirm that at least the latter works, so wdyt?

And I can certainly leave `setUp` as is and rewrite the parametrized test if you still prefer that approach.

fcurella (approved):
Thank you!

PR #1888, opened 2023-07-12
add ja_JP datetime function

cyanghsieh:
### What does this change
Add `MONTH_NAMES`, `DAY_NAMES` and 'TRADITIONAL_MONTH_NAMES` to ja_JP locale.

### What was wrong
Not implemented before.

### How this fixes it
Not implemented before.


fcurella (approved):
Thank you!

PR #1891, opened 2023-07-16
Add support for South African Zulu (zu_ZA) names and last names

iamkhaya:
### What does this change

Adds support for South African Zulu first names and last names

### What was wrong

This was currently missing.

### How this fixes it

This adds the missing person provider


fcurella (approved):
Thank you!

Issue #1892, opened 2023-07-31
`DynamicProvider` does not take weights into account.


pauldechorgnat:
* Faker version: 19.2
* OS: Windows

`DynamicProvider` does not take weights into account when we provide an `OrderedDict`.

### Steps to reproduce

```python
from faker import Faker
from faker.providers import DynamicProvider
from typing import OrderedDict
from collections import Counter

elements = ["a", "b", "c", "d", "e"]
probabilities = [.5, .2, .3, .0, .0]

weighted_elements = OrderedDict(zip(elements, probabilities))

provider = DynamicProvider(
    elements=weighted_elements,
    provider_name="nutriscore"
)

f = Faker()
f.add_provider(provider)

results = [f.nutriscore() for _ in range(100_000)]

Counter(results)
```

This will return something like:

```python
Counter({'d': 20128, 'b': 20071, 'a': 20042, 'e': 19938, 'c': 19821})
```

### Expected behavior

By providing an `OrderedDict`, we should be able to get data sampled using the weights of the `OrderedDict`

### Actual behavior

Actually, the `DynamicProvider` does not take weights into account: the above example shows an equal distribution between the elements.

### Proposed fix

To allow weighted `DynamicProvider`, we only need to change the definition of the `get_random_value` method:

```python
class DynamicProvider(BaseProvider):
    # ...
    def get_random_value(self) -> Any:
        if not self.elements or len(self.elements) == 0:
            raise ValueError("Elements should be a list of values the provider samples from")

        return self.random_elements(self.elements, length=1, use_weighting=True)[0]
```

It could be interesting to let the user choose if he wants to use the weights or not, even though he gave an `OrderedDict`. There are two ways to design this:

1. use an instance variable `use_weighting` defined in the constructor
2. use a parameter in the `get_random_value`

The first option would give us something like:

```python
class DynamicProvider(BaseProvider):
	def __init__(
        self,
        provider_name: str,
        elements: Optional[List] = None,
        generator: Optional[Any] = None,
		use_weighting: bool = False,
    ):
		"""..."""
		self.use_weighting = use_weigthing
		# ...

    def get_random_value(self) -> Any:
        if not self.elements or len(self.elements) == 0:
            raise ValueError("Elements should be a list of values the provider samples from")

        return self.random_elements(self.elements, length=1, use_weighting=self.use_weighting)[0]

```

The second option would give us:

```python
class DynamicProvider(BaseProvider):
    # ...
[...]

PR #1894, opened 2023-08-04
Add weigthing to dynamic providers

pauldechorgnat:
### What does this change

`DynamicProvider` now accept weighted elements if provided an `OrderedDict`

### What was wrong

`DynamicProvider` were not taking weights into account.

See Issue #1892 

Also fixes a C419 flake8 error on a test file.


fcurella (approved):
Thank you!

PR #1904, opened 2023-09-04
Additional pseudo-random color fakes

### LINKED ISSUE #1903: RGB/HSV/HSL color tuples from `color` provider
* Faker version: 19.3.1
* OS: any

With #104/#1055 the `color` provider learned how to generate colors with certain characteristics like "dark" or from a given hue. 

As noted by @malefice in the previous PR, the faked values are only available as strings. Often a numeric format is more suitable as input for other data-generation or computation. I am personally looking for this functionality to generate placeholder-images with similar colors and [pycairo](https://pycairo.readthedocs.io/en/latest/reference/patterns.html#class-solidpattern-pattern) works with RGB as float. 

With a little refactoring it should be easy to reuse the logic already implemented in `RandomColor` to provide int/float tuples for pseudo-random colors.

### Expected behavior

Instead of adding another `color_format` argument and `color()` having non-string return values, additional fakes could provide the color tuples using the same `hue` and `luminosity` kwargs as `color()`.

```
>>> from faker import Faker
>>> fake = Faker()
>>> fake.color(hue='red')
'#931929'
>>> fake.color_rgb(hue='red')
(247, 127, 118)
>>> fake.color_rgb_float(hue='red')
(0.98, 0.68355, 0.6566)
>>> fake.color_hsv(hue='red')
(334, 91, 78)
>>> fake.color_hsl(hue='red')
(340, 57, 60)
```

### Feature overlap and further changes...

The current color provider only generates strings.

There is however some overlap with `color()` and `rgb_css_color()`, that probably should be refactored/deprecated in a separate step, i.e.:

```
>>> fake.color(hue='blue', color_format='rgb')
'rgb(25, 211, 232)'
>>> fake.rgb_css_color()
'rgb(117,96,199)'
```

fdemmer:
### What does this change

Add `color_rgb`, `color_rgb_float`, `color_hsv` and `color_hsl` fakes

### What was wrong

Pseudo-random colors were only available as strings.

### How this fixes it

Additional fakes provide tuples of int/float.

Fixes #1903

### DIFF
--- a/faker/providers/color/color.py
+++ b/faker/providers/color/color.py
@@ -295,11 +295,11 @@ class RandomColor:
         """Return the hue range for a given ``color_input``."""
         if isinstance(color_input, (int, float)) and 0 <= color_input <= 360:
             color_input = int(color_input)
-            return (color_input, color_input)
+            return color_input, color_input
         elif isinstance(color_input, str) and color_input in self.colormap:
             return self.colormap[color_input]["hue_range"][0]
         elif color_input is None:
-            return (0, 360)
+            return 0, 360
 
         if isinstance(color_input, list):
             color_input = tuple(color_input)
@@ -315,7 +315,7 @@ class RandomColor:
                 v1, v2 = v2, v1
             v1 = max(v1, 0)
             v2 = min(v2, 360)
-            return (v1, v2)
+            return v1, v2
         raise TypeError("Hue must be a valid string, numeric type, or a tuple/list of 2 numeric types.")
 
     def get_saturation_range(self, hue: int) -> Tuple[int, int]:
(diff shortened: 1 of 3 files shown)


faker/providers/color/color.py
```diff
@@ -144,37 +144,64 @@ def __init__(self, generator: Optional["Generator"] = None, seed: Optional[Hasha
...
-            s_max, b_min = lower_bounds[-1]
-
-            self.colormap[color_name]["saturation_range"] = [(s_min, s_max)]
-            self.colormap[color_name]["brightness_range"] = [(b_min, b_max)]
-
     def generate(
         self,
         hue: Optional[HueType] = None,
         luminosity: Optional[str] = None,
         color_format: str = "hex",
```
fcurella, line 151: can we use the [`Literal` type hint](https://docs.python.org/3.9/library/typing.html#typing.Literal) here?

fcurella (changes requested):
Thank you! Could you also run `make lint` to apply our formatting?

faker/providers/color/color.py
fdemmer: good idea, done.

fdemmer:
> Thank you! Could you also run make lint to apply our formatting?

I did, but it did not find any issues; I suspect there was another issue with the pipeline before :shrug:

fcurella (approved):
Thank you! 🌈

Issue #1907, opened 2023-09-07
Pyfloat raises TypeError for None min or max value


HugoJP1:
* Faker version: 19.3.0
* OS: Mac OS 12.2.1 and Ubuntu 22.04.3
* Python 3.11

When calling `pyfloat` with a `max_value` of `None` and a `min_value`, then the method sometimes raises
`unsupported operand type(s) for -: 'NoneType' and 'float'`

### Steps to reproduce
Run the following code
```python
from faker import Faker


for i in range(3000):
    try:
        Faker().pyfloat(
                    left_digits=5,
                    right_digits=2,
                    min_value=99884.7,
                    positive=True,
                    max_value=None,
                )
    except Exception as e:
        print(e)
        print(i)
```

### Expected behavior

There should be no exception printed.

### Actual behavior

```
unsupported operand type(s) for -: 'NoneType' and 'float'
17
unsupported operand type(s) for -: 'NoneType' and 'float'
155
unsupported operand type(s) for -: 'NoneType' and 'float'
175
unsupported operand type(s) for -: 'NoneType' and 'float'
510
unsupported operand type(s) for -: 'NoneType' and 'float'
541
unsupported operand type(s) for -: 'NoneType' and 'float'
943
unsupported operand type(s) for -: 'NoneType' and 'float'
1005
unsupported operand type(s) for -: 'NoneType' and 'float'
1243
unsupported operand type(s) for -: 'NoneType' and 'float'
1255
unsupported operand type(s) for -: 'NoneType' and 'float'
1399
unsupported operand type(s) for -: 'NoneType' and 'float'
1677
unsupported operand type(s) for -: 'NoneType' and 'float'
1854
unsupported operand type(s) for -: 'NoneType' and 'float'
2036
unsupported operand type(s) for -: 'NoneType' and 'float'
2247
unsupported operand type(s) for -: 'NoneType' and 'float'
2448
unsupported operand type(s) for -: 'NoneType' and 'float'
2578
unsupported operand type(s) for -: 'NoneType' and 'float'
2812
unsupported operand type(s) for -: 'NoneType' and 'float'
2826

Process finished with exit code 0
```

-----

I dug into the code a little and it looks to me like the issue with the code is here:
```python
# It's possible for the result to end up > than max_value or < than min_value
        # When this happens we introduce some variance so we're not always the exactly the min_value or max_value.
        # Which can happen a lot depending on the difference of the values.
        # Ensure the variance is bound by the difference between the max and min
[...]

PR #1908, opened 2023-09-08
fix: pyfloat type error on none max value

HugoJP1:
### What does this change
Handle the combinations of specified digits and min and max values in a more advanced way for pyfloat.

### What was wrong

There were a few issues that this PR aims to address:
- This line was throwing a TypeError if max_value or min_value was None
```python
self.generator.random.uniform(0, max_value - min_value)
```

- This line would throw an error when `needed_left_digits == sys.float_info.dig` because the range was then invalid
- Giving in `-(10**sys.float_info.dig + 1)` as max value and `10**sys.float_info.dig` as min_value can never succeed because of the following code at the end:
```python
        if right_digits:
            result = min(result, 10**left_digits - float(f'0.{"0" * (right_digits - 1)}1'))
            result = max(result, -(10**left_digits + float(f'0.{"0" * (right_digits - 1)}1')))
        else:
            result = min(result, 10**left_digits - 1)
            result = max(result, -(10**left_digits + 1))
```
- The number of left and right digits were not always respected.

### How this fixes it
- `ValueError` is raised if max_value or min_value is out of bounds
- right digits is set to 0 if the inferred left digits is already equals to sys.float_info.dig
- Rather than choosing only the `left_number` which then caused issues in trying to figure out a "right number" which could then end up being lower/higher than the configured max/min value, an int is chosen randomly between min and max values which are scaled up to have the length of left + right digits so that afterwards a `.` can be inserted in the right place.
- The previous adjustment code containing the TypeError is removed.

Fixes https://github.com/joke2k/faker/issues/1907

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -161,11 +161,21 @@ class Provider(BaseProvider):
             raise ValueError("Max value must fit within left digits")
         if left_digits is not None and min_value and math.ceil(math.log10(abs(min_value))) > left_digits:
             raise ValueError("Min value must fit within left digits")
+        if max_value is not None and max_value < -(10**sys.float_info.dig + 1):
+            raise ValueError(f"Max value must be greater than {-10**sys.float_info.dig - 1}")
+        if min_value is not None and min_value > 10**sys.float_info.dig - 1:
+            raise ValueError(f"Min value must be less than {10**sys.float_info.dig - 1}")
 
+        left_digits_requested = left_digits is not None
+        right_digits_requested = right_digits is not None
         # Make sure at least either left or right is set
         if left_digits is None and right_digits is None:
             needed_left_digits = max(1, math.ceil(math.log10(max(abs(max_value or 1), abs(min_value or 1)))))
-            right_digits = self.random_int(1, sys.float_info.dig - needed_left_digits)
+            if (sys.float_info.dig - needed_left_digits) < 1:
+                right_digits = 0
+            else:
+                right_digits = self.random_int(1, sys.float_info.dig - needed_left_digits)
+
 
         # If only one side is set, choose #digits for other side
         if (left_digits is None) ^ (right_digits is None):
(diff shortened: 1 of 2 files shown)


faker/providers/python/__init__.py
```diff
@@ -181,32 +191,67 @@ def pyfloat(
...
+            if max_value is None:
+                temp_max_value = 10 ** (left_digits + right_digits)
+                if min_value is not None:
+                    temp_min_value = min_value * 10 ** right_digits
+            if min_value is None:
+                temp_min_value = -(10 ** (left_digits + right_digits))
+                if max_value is not None:
+                    if max_value < temp_min_value:
+                        temp_min_value = max_value - 1
+                    temp_max_value = max_value * 10 ** right_digits
```
HugoJP1, line 208: This section aims to infer boundaries in the case where we don't have max_value or min_value.

faker/providers/python/__init__.py
```diff
@@ -181,32 +191,67 @@ def pyfloat(
                 f"{sys.float_info.dig})",
             )
 
-        sign = ""
         if (min_value is not None) or (max_value is not None):
             # Copy values to ensure we're not modifying the original values and thus going out of bounds
-            left_min_value = min_value
-            left_max_value = max_value
+            temp_min_value = min_value
+            temp_max_value = max_value
```
HugoJP1, line 197: Renamed to `temp` rather than `left` because it is now doing both the right and left parts.

faker/providers/python/__init__.py
```diff
@@ -181,32 +191,67 @@ def pyfloat(
...
+                    10 ** (left_digits + right_digits - 1),
+                    temp_max_value,
+                    positive=True,
+                )
+                negative_number = self._safe_random_int(
+                    temp_min_value,
+                    -(10 ** (left_digits + right_digits - 1)),
+                    positive=False,
+                )
+                number = self.random_element((positive_number, negative_number))
```
HugoJP1, line 227: This part is so that we can generate a number which respects these two constraints:
1. It must have a fixed amount of left digits
2. It can be either positive or negative.

In this case we need to choose two numbers, one in the negative side, and one in the positive. We then choose randomly between the two.

fcurella (approved):
Thank you! <3

PR #1918, opened 2023-10-07
Feat/mac address multicast

d3vyce:
### What does this change

Add the options to create unicast or multicast MAC address (default: unicast).
As defined in IEEE 802.3, all MAC addresses that have the least significant bit of the most significant byte as "1" are multicast.
This is also mentioned in the RFC7042.

### What was wrong

Previously, the project did not differentiate between unicast and multicast mac addresses. 

### How this fixes it

This PR allows you to choose between unicast and multicast mac addresses. A test set and parameter documentation have also been added.

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -582,8 +582,18 @@ class Provider(BaseProvider):
             address = str(IPv6Network(address, strict=False))
         return address
 
-    def mac_address(self) -> str:
-        mac = [self.generator.random.randint(0x00, 0xFF) for _ in range(0, 6)]
+    def mac_address(self, multicast: bool = False) -> str:
+        """
+        Returns a random MAC address.
+
+        :param multicast: Multicast address
+        :returns: MAC Address
+        """
+        mac = [self.generator.random.randint(0x00, 0xFF) for _ in range(0, 5)]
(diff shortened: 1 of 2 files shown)


faker/providers/internet/__init__.py
```diff
@@ -582,8 +582,18 @@ def ipv6(self, network: bool = False) -> str:
...
-        mac = [self.generator.random.randint(0x00, 0xFF) for _ in range(0, 6)]
+    def mac_address(self, multicast: bool = False) -> str:
+        """
+        Returns a random MAC address.
+
+        :param multicast: Multicast address
+        :returns: MAC Address
+        """
+        mac = [self.generator.random.randint(0x00, 0xFF) for _ in range(0, 5)]
+        if multicast is True:
```
stefan6419846, line 593: Which advantages does this have over `if multicast:`?

d3vyce: None, I just tried to follow the existing code style, example of the `ipv4` function:
```python
if private is True:
    return self.ipv4_private(address_class=address_class, network=network)
elif private is False:
    return self.ipv4_public(address_class=address_class, network=network)
```

fcurella (approved):
Thank you!

PR #1920, opened 2023-10-08
fix spelling error for pyspark

NhSimple:
### What does this change
Changed "datafrane" to "dataframe"

### What was wrong

Dataframe was incorrectly spelled

### How this fixes it

Dataframe is now correctly spelled


fcurella:
Thank you!

PR #1922, opened 2023-10-09
Gender support for Latvian (lv_LV) names

OskarsPakers:
### What does this change

Separated male and female person first names and last names.


fcurella (approved):
thank you

Issue #1916, opened 2023-09-30
de_CH returns country names encoded in iso-8859-1 instead of utf-8


nchiapol:
* Faker version: 19.6.2
* OS: Debian, testing (But running faker installed with pip into a venv)

Faking country names returns strings encoded in iso-8859-1 instead of utf-8 when using the `de_CH` locale, when using the default locale, all utf-8 characters are converted to ASCII instead. `de_DE` and `fr_FR` seem to be doing the right thing.

### Steps to reproduce

* run the following script (with and without setting the locale):  
```
# coding: utf-8
import faker, random
faker.Faker.seed(0)
fkr = faker.Faker('de_CH')

for i in range(1000):
    res = fkr.country()
    if "and Pr" in res: #or "Ivoire" in res:
        print(f"Iteration: {i}")
        print("orig.: ", res)
        print("fixed: ", res.encode('iso-8859-1').decode())
        break
```

* observe output:
```
# with de_CH
Iteration: 68
orig.:  SÃ£o TomÃ© and PrÃ­ncipe
fixed:  São Tomé and Príncipe

# without de_CH
Iteration: 625
orig.:  Sao Tome and Principe
fixed:  Sao Tome and Principe
```

### Expected behavior

In all cases all returned strings are correctly encoded in utf-8

### Partial solution

Based on the fact, that `de_DE` works, the easiest solution for `de_CH` seems to be, to change the relative import `from .. import Provider as AddressProvider` to `from ..de import Provider as AddressProvider` in `faker/providers/address/de_CH/__init__.py`

nchiapol:
done, thanks for your work

PR #1924, opened 2023-10-09
use DE base for AddressProvider (#1916)

nchiapol:
### What does this change
Partial fix for #1916 

The original base (via `providers/date_time/__init__.py`) still loads strings with the wrong encoding.


fcurella (approved):
Thank you!

PR #1925, opened 2023-10-11
cs_CZ tune phone number (plus add official sources)

george0st:
### What does this change

Improve cs_CZ phone numbers

### What was wrong

Missing some ranges and add official sources


george0st:
Please merge

fcurella (approved):
Thank you!

PR #1929, opened 2023-10-13
Fix type hint of `providers.barcode.ean8`

Viicos:
I think this was a mistake when type hints were added.


fcurella (approved):
Thank you!

PR #1932, opened 2023-10-18
Color for cs_CZ

george0st:
### What does this change

Add provider for color for cs_CZ

### What was wrong

Missing color provider for cs_CZ


fcurella (changes requested):
Could you add tests?

PR #1935, opened 2023-10-18
Add global geo for Europe/Prague

george0st:
### What does this change

Add global geo for Europe/Prague

### What was wrong

Missing more geo location from Europe/Prague, CZ


fcurella (approved):
Thank you!

PR #1937, opened 2023-10-20
added handling for finnish ssn where min age and max age are the same

### LINKED ISSUE #1936: When attempting to generate a Finnish ssn of a specific age the provider crashes
* Faker version: 19.11.0
* OS: Windows 10

When attempting to generate a Finnish ssn of a certain age the provider crashes because ``datetime.timedelta(days=self.generator.random.randrange(min_age * 365, max_age * 365))`` is always trying to generate a time within range. The ssn provider should handle situations where min_age and max_age are the same.
For example:
```
if min_age == max_age:
    age = datetime.timedelta(days=min_age * 365)
else:
    age = datetime.timedelta(days=self.generator.random.randrange(min_age * 365, max_age * 365))
```

### Steps to reproduce

1. Create a faker with fi_FI locale: ``fake = Faker('fi_FI')``
2. Call the ssn provider with the same minimum and maximum age: ``fake.ssn(min_age=50, max_age=50, artificial=True)``

### Expected behavior

The faker generates an artificial Finnish ssn that is 50 years old. 

### Actual behavior

``ValueError: empty range for randrange() (18250, 18250, 0)``

Pakkanen1:
### What does this change

Allows the same min_age and max_age parameters for the Finnish ssn provider.
Added ``test_ssn_without_age_range()`` to test scenarios from the 3 supported centuries. 

### What was wrong

See issue #1936  

### How this fixes it

When ``max_age`` and ``min_age`` are the same, the days a person has been alive are calculated according to ``min_age`` instead of picking a random number.

Fixes #1936

### DIFF
--- a/faker/providers/ssn/fi_FI/__init__.py
+++ b/faker/providers/ssn/fi_FI/__init__.py
@@ -26,7 +26,10 @@ class Provider(SsnProvider):
             checksum_characters = "0123456789ABCDEFHJKLMNPRSTUVWXY"
             return checksum_characters[int(hetu) % 31]
 
-        age = datetime.timedelta(days=self.generator.random.randrange(min_age * 365, max_age * 365))
+        if min_age == max_age:
+            age = datetime.timedelta(days=min_age * 365)
+        else:
+            age = datetime.timedelta(days=self.generator.random.randrange(min_age * 365, max_age * 365))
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

PR #1938, opened 2023-10-23
Add geo for cs_CZ

george0st:
### What does this change

Add geo for cs_CZ

### What was wrong

Missing geo for cs_CZ


tests/providers/test_geo.py
```diff
@@ -79,6 +79,11 @@ def test_location_on_land_coords_only(self):
         assert Decimal(loc[0])  # Should be able to cast first two elements of tuple to Decimal
         assert Decimal(loc[1])
 
+class TestCsCz(unittest.TestCase):
+    def setUp(self):
+        self.fake = Faker("cs_CZ")
+        Faker.seed(0)
+
```
fcurella, line 86: could you add some test methods?

fcurella (changes requested):

fcurella (approved):
Thank you!

PR #1939, opened 2023-10-25
Code Completion: Adding stub generator utility (and generated stub)

KaylaHood:
### What does this change

The generate_stubs.py script creates a stub file, proxy.pyi, and saves it to the faker package's directory. This stub file contains extracted method declarations and docstrings (pulled from the providers) which are used by IDEs to provide code completion.

### What was wrong

Because the provider classes are "dynamically" linked at runtime, IDEs can't get a list of available faker methods to display as code completion suggestions.

### How this fixes it

Fixes #...
Generate a stub file (proxy.pyi) using the utility script (generate_stubs.py) to help IDEs provide code completion to developers.


Viicos:
Great! A couple things I saw:
- It seems some `typing` imports are missing, e.g. `Literal`
- It would be great to run `black` after the stub file is generated: we can use the API: `black.format_str(..., mode=black.Mode(is_pyi=True))`
- I don't think attributes (such as `credit_card_types`) are available on the `Faker` instance at runtime
- Do we want all the localized providers to be available, or only the methods from the provider in the `__init__.py`?

This can of course be discussed

gryznar:
Is it going to happen something with this pr? Missing code completion is a big drawback :(

KaylaHood:
> @KaylaHood @Viicos Is it going to happen something with this pr? Missing code completion is a big drawback :(

Sorry, I totally forgot to revisit this PR after I got the comment from Viicos.
@Viicos I have some questions:

- Where are the `typing` imports missing? In the generate_stubs.py script?
- How should I tie in the black API to the stub file generation? Would you like me to add a `generate-stub` recipe to the Makefile that has the `black` recipe as a prereq?
- Oops, you're right about the attributes not being accessible at runtime. I can fix that.
- I think all of the localized providers should be available in the stub, because that's the most common usage of the faker library (most users don't select a subset of locales, they just use the `Faker` class as it comes by default). However, it is true that this can be confusing for situations where a specific locale/set of locales was given to the `Faker` initializer - the auto-complete will include methods that don't exist in the selected locale(s). I don't really see a way to avoid this without major refactoring of the Faker library (the locale selection of the `Faker` instance would need to be determinable prior to runtime).

Viicos:
> Where are the `typing` imports missing? In the generate_stubs.py script?

Sorry, I meant in the generated stub file.

> How should I tie in the black API to the stub file generation? Would you like me to add a `generate-stub` recipe to the Makefile with the `black` recipe as a prereq?

I can look into that.

> I think all of the localized providers should be available in the stub, because that's the most common usage of the faker library.

Ah in that case it makes sense, I thought at first that only the non locale-specific methods were available. Maybe worth double checking this.

---

Other than that, the script looks good, but feels a bit hacky and could potentially lead to invalid syntax. I think a more robust implementation could be done using ast/cst, but that can be time consuming to implement. If I have time some day, I can come up with an implementation (but for later, sticking with your implementation is fine for now!)

KaylaHood:
I removed the attributes from the stub file and I added a recipe for `generate-stubs` to the Makefile. I also added it as a prereq to the `release` recipe. I didn't add `black` as a prereq because it wasn't already a prereq for the `release` recipe which implies it was intended as a "use-when-desired" recipe rather than an "always-run-prior-to-release" recipe.

> Other than that, the script looks good, but feels a bit hacky and could potentially lead to invalid syntax. I think a more robust implementation could be done using ast/cst, but that can be time consuming to implement. If I have time some day, I can come up with an implementation (but for later, sticking with your implementation is fine for now!)

Yes, this was intended as a quick hack to get the autocomplete working for me while I wrote unit tests. I figured it'd be useful to the others on the issue thread who were having the same frustration. If you'd like to improve it, please be my guest!

KaylaHood:
> Ah in that case it makes sense, I thought at first that only the non locale-specific methods were available. Maybe worth double checking this.

I made sure to include all the available locales in my stub generator ([see lines 85-91](https://github.com/joke2k/faker/pull/1939/files#diff-5d30fdeb43b3adab06bd4dc48f33062f55bd87c3a0a838493f4c70ba19902ae8R85-R91))

fcurella:
If the stubs are to be committed to the repo, then I would add `generate-stubs` to the `lint` recipe, rather than `release`.

We should alse update our [contributing docs](https://github.com/joke2k/faker/blob/master/CONTRIBUTING.rst) to mention running the task when submitting new PRs

KaylaHood:
> If the stubs are to be committed to the repo, then I would add `generate-stubs` to the `lint` recipe, rather than `release`.
> 
> We should alse update our [contributing docs](https://github.com/joke2k/faker/blob/master/CONTRIBUTING.rst) to mention running the task when submitting new PRs

Sorry for another delay, I was away on holiday. I've made the changes you suggested, can you check them out?

fcurella (approved):
Nice work! Thank you!

MarcelWilson:
Does it make sense to try to add a unittest which verifies all the expected methods are present in the stub?

faker/proxy.pyi
```diff
@@ -0,0 +1,3415 @@
...
+        """
+        ...
+    def postcode_in_state(self, state_abbr: Optional[str] = ...) -> str:
+        """
+        :returns: A random postcode within the provided state abbreviation
+        
+        :param state_abbr: A state abbreviation
+        """
+        ...
+    def random_choices(self, elements: Union[Collection[str], Collection[T], collections.OrderedDict[T, float]] = ..., length: Optional[int] = ...) -> Sequence[T]:
```
MarcelWilson, line 147: Neither `Collection` nor `collections` appear in the imports

faker/proxy.pyi
```diff
@@ -0,0 +1,3415 @@
...
+        :sample: base='123456', number_system_digit=1
+        :sample: base='120000', number_system_digit=0
+        :sample: base='120003', number_system_digit=0
+        :sample: base='120004', number_system_digit=0
+        :sample: base='120000', number_system_digit=0, safe_mode=False
+        :sample: base='120003', number_system_digit=0, safe_mode=False
+        :sample: base='120004', number_system_digit=0, safe_mode=False
+        """
+        ...
+    def color(self, hue: Optional[HueType] = ..., luminosity: Optional[str] = ..., color_format: str = ...) -> str:
```
MarcelWilson, line 667: `HueType` isn't imported

faker/proxy.pyi
```diff
@@ -0,0 +1,3415 @@
...
+        Refer to https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html for details.
+        
+        The arguments ``hue`` and ``luminosity`` are the same as in the color provider and are simply forwarded to
+        it to generate both the background and the shape colors. Therefore, you can ask for a "dark blue" image, etc.
+        
+        :sample: size=(2, 2), hue='purple', luminosity='bright', image_format='pdf'
+        :sample: size=(16, 16), hue=[90,270], image_format='ico'
+        """
+        ...
+    def json(self, data_columns: Optional[List] = ..., num_rows: int = ..., indent: Optional[int] = ..., cls: Optional[Type[json.encoder.JSONEncoder]] = ...) -> str:
```
MarcelWilson, line 1664: `json` isn't imported

faker/proxy.pyi
```diff
@@ -0,0 +1,3415 @@
...
+        Generate a credit card expiry date.
+        
+        This method uses |date_time_between| under the hood to generate the
+        expiry date, so the ``start`` and ``end`` arguments work in the same way
+        here as it would in that method. For the actual formatting of the expiry
+        date, |strftime| is used and ``date_format`` is simply passed
+        to that method.
+        """
+        ...
+    def credit_card_full(self, card_type: Optional[CardType] = ...) -> str:
```
MarcelWilson, line 785: `CardType` isn't imported

faker/proxy.pyi
```diff
@@ -0,0 +1,3415 @@
...
+        """
+        ...
+    def time_object(self, end_datetime: Union[datetime.date, datetime.datetime, datetime.timedelta, str, int, None] = ...) -> datetime.time:
+        """
+        Get a time object
+        
+        :example: datetime.time(15, 56, 56, 772876)
+        """
+        ...
+    def time_series(self, start_date: Union[datetime.date, datetime.datetime, datetime.timedelta, str, int] = ..., end_date: Union[datetime.date, datetime.datetime, datetime.timedelta, str, int] = ..., precision: Optional[float] = ..., distrib: Optional[Callable[[datetime.datetime], float]] = ..., tzinfo: Optional[datetime.tzinfo] = ...) -> Iterator[Tuple[datetime.datetime, Any]]:
```
MarcelWilson, line 1093: `Callable` isn't imported

faker/proxy.pyi
```diff
@@ -0,0 +1,3415 @@
+
+from typing import Any, Iterable, Sequence, Union, Type, Dict, List, Iterator, Tuple, Optional, TypeVar, Set
+from enum import Enum
+from faker.providers import T
+import datetime
+from decimal import Decimal
+from faker.providers.python import TEnum
+from faker.generator import Generator
+
+T = TypeVar("T")
```
MarcelWilson, line 10: `T` is redeclared after it was imported.

generate_stubs.py
```diff
@@ -0,0 +1,180 @@
...
+from faker import Factory, Faker
+
+import faker.proxy
+
+BUILTIN_MODULES_TO_IGNORE = ["builtins"]
+GENERIC_MANGLE_TYPES_TO_IGNORE = ["builtin_function_or_method", "mappingproxy"]
+MODULES_TO_FULLY_QUALIFY = ["datetime"]
+
+
+imports: Dict[str, Optional[Set[str]]] = defaultdict(lambda: None)
```
MarcelWilson, line 16: Annotation should use `DefaultDict` instead of `Dict`

generate_stubs.py
```diff
@@ -0,0 +1,180 @@
...
+
+classes_and_locales_to_use_for_stub: List[Tuple[object, str]] = []
+for locale in AVAILABLE_LOCALES:
+    for provider in PROVIDERS:
+        if provider == "faker.providers":
+            continue
+        prov_cls, _, _ = Factory._find_provider_class(provider, locale)
+        classes_and_locales_to_use_for_stub.append((prov_cls, locale))
+
+all_members: List[Tuple[UniqueMemberFunctionsAndVariables, str]] = \
```
MarcelWilson, line 93: Should be `List[Tuple[UniqueMemberFunctionsAndVariables, Optional[str]]]` since the last item in the list has `None` in it.

generate_stubs.py
```diff
@@ -0,0 +1,180 @@
...
+        for module in imports.keys():
+            if module in MODULES_TO_FULLY_QUALIFY:
+                continue
+            sig_str = sig_str.replace(f"{module}.", "")
+
+        comment = inspect.getdoc(func_value)
+        signatures_with_comments.append((f"def {func_name}{sig_str}: ...", None if comment == "" else comment, False))
+
+signatures_with_comments_as_str = []
+for sig, comment, is_preceding_comment in signatures_with_comments:
```
MarcelWilson, line 147: mypy errors because it infers `sig` is `Signature` from line 110 instead of `str`.  Easiest solution would be to rename this variable.

generate_stubs.py
```diff
@@ -0,0 +1,180 @@
...
+                    imports[module].add(member)
+        
+        new_parms = []
+        for key, parm_val in sig.parameters.items():
+            new_parm = parm_val
+            if parm_val.default is not inspect.Parameter.empty:
+                new_parm = parm_val.replace(default=...)
+            if (new_parm.annotation is not inspect.Parameter.empty 
+                and not new_parm.annotation.__module__ in BUILTIN_MODULES_TO_IGNORE):
+                module, member = get_module_and_member_to_import(new_parm.annotation, locale)
```
MarcelWilson, line 128: It looks like the reason a bunch of imports are missing is this function doesn't look inside annotations.  So whenever `Union[x, y]` occurs, it doesn't grab x & y.

faker/proxy.pyi
```diff
@@ -0,0 +1,3415 @@
...
+        ...
+    def __getitem__(self, locale: str) -> Generator: ...
+    def __init__(self, locale: Union[str, Sequence[str], Dict[str, Union[int, float]], None] = ..., providers: Optional[List[str]] = ..., generator: Optional[Generator] = ..., includes: Optional[List[str]] = ..., use_weighting: bool = ..., **config: Any) -> None:
+        """
+        Initialize self.  See help(type(self)) for accurate signature.
+        """
+        ...
+    def __setstate__(self, state: Any) -> None: ...
+    def items(self) -> List[Tuple[str, Generator]]: ...
+    def seed(seed: Union[int, float, str, bytes, bytearray, None] = ...) -> None:
```
MarcelWilson, line 3394: signature for this doesn't match.  It's supposed to be a classmethod.

MarcelWilson:
I tweaked generate_stub.py to fix the issues with missing imports, staticmethods, and classmethods.
https://gist.github.com/MarcelWilson/bf830e853b54a59603013d50a801d948#file-generate_stub-py
Here is the stub output: 
https://gist.github.com/MarcelWilson/ca6eaa52f91015c3bc2abbe5a442d6de#file-proxy-pyi

I took the liberty of removing the docstrings as they aren't needed in a stub.  They can easily be re-enabled by flipping these lines:
https://gist.github.com/MarcelWilson/bf830e853b54a59603013d50a801d948#file-generate_stub-py-L161-L162

I'm not 100% it's covering all the methods without a unittest.

bandophahita:
I went and made a branch with the changes I suggested.   I even got the black and isort formatting included in the generate_stubs.py.
https://github.com/bandophahita/faker/tree/faker-stubs-rebased

If you'd like I can create a new PR with the updates.

agusdmb:
i've been following this PR, i would really like for it to make into master that's why im leaving this message.. having these stubs would be really helpful

fcurella:
@KaylaHood do you want to ingrate @bandophahita's changes into your branch?

fcurella:
@bandophahita ball's in your court! 🙂

bandophahita:
@KaylaHood  While I appreciate the gesture, I would say if it is easier feel free to integrate.

RanaRana10:
i dont understand what did you say how i can active the type hint showing in vs code any tutorials or articles

Issue #1940, opened 2023-10-25
Italian SSN does not take into account the latest days of the months


TommasoLencioni:
* Faker version: 19.12.0
* OS: macOS 14.0

When generating an italian SSN the days 29, 30 and 31 (of suitable months) are not considered.
This simplify the logic but exclude realistic cases.
This is due to the range of `random_int` in [SSN generation method](https://github.com/joke2k/faker/blob/master/faker/providers/ssn/it_IT/__init__.py#L8022).

### Steps to reproduce

1. The issue can be statically observed from the code. There is no way to generate to generate a SSN `{surname}{name}{year}{month}{29|30|31}{municipality}{checksum}` from `Faker('it_IT').ssn()`.

### Expected behavior

A SSN generated with `Faker('it_IT').ssn()` can be referred to a citizen born on 29, 30 and 31 of a suitable month.

### Actual behavior

No SSN generated with `Faker('it_IT').ssn()` can be referred to a citizen born on 29, 30 and 31 of a suitable month.

stefan6419846:
Do you want to submit a corresponding PR which correctly uses a month-based limit?

PR #1941, opened 2023-10-25
Fix latest days of the month as birth day in italian SSN

TommasoLencioni:
### What does this change

This PR refers to [this issue](https://github.com/joke2k/faker/issues/1940).

### What was wrong

Random interval in Italian SSN generation was not broad enough to consider all possible birth dates.

### How this fixes it

A new method check for leap year and another new method returns the maximum day for the generated month-year couple.
This maximum day is used as upper bound for random_int day generation.


fcurella (approved):
Thank you!

PR #1942, opened 2023-10-25
Addition geo

george0st:
### What does this change

Add other geo for cs_CZ

### What was wrong

Add missing geo


fcurella (approved):
Thank you!

PR #1946, opened 2023-11-02
Support platform specific second precision

### LINKED ISSUE #1945: Subsecond precision support for fake time
* Faker version: 19.10.0
* OS: Arch Linux

Randomized datetimes produced by a faker instance only ever have second precision. This is rarely a problem but when combined with systems that send datetimes as unix timestamps (because it is easy to deal with in json), it can become a problem that `when.timestamp() == int(when.timestamp())`. This is caused by pythons somewhat loose equality check on numbers where floats and integers can compared and would actually be equal each other, when the fractional part of a float is zero.

I have skimmed the code generating datetime objects and it seems that in most (if not all) cases the randomness comes from a simpler random generation in the form of a second interval generated by `random.randint`, which as the name implies only ever returns integers and not floats as my needs would require.

If this has any interest then I could probably give it a go replacing the innermost calls to `random.randint` in the datetime provider, with something like `random.uniform`. But since am not very familiar with this code, there might be more suitable solution that I am just unaware of.

### Steps to reproduce

```python
from datetime import datetime, timezone, timedelta
from faker import Faker

faker = Faker()
point_in_time = datetime.now(timezone.utc)

randomized_datetime = faker.date_time_between(
	point_in_time,
	point_in_time + timedelta(days=5),
)

# technically this could be able to pass,
# but I have never been so lucky:
assert randomized_datetime.microsecond != 0

# or alternatively from the description above:
unix_ts = randomized_datetime.timestamp()
assert unix_ts != int(unix_ts)
```

### Expected behavior

When producing randomized datetime objects the microsecond value should be non-zero, for sub-second precision.

### Actual behavior

Randomized datetime objects only ever have second precision.

cknv:
### What does this change

This inspects the platform and defines a platform specific property for generating random seconds.
For windows this uses `random.randint` as used before and for everything else `random.uniform`, which returns floats instead of ints.

Then later on when trying to generate a random number of seconds this proper is then used to get the function that can return values with the correct behaviour.

I tried setting the function directly on `self` in an `__init__`, but that worked poorly when seeding the faker after creation, as seeding recreates the internal random instance, so it requires some lazy evaluation to be able to point to the correct random instance in all cases.

Internally the points in time is still just ints. Changing those to floats is not strictly important, and would just add noise to diff - making it harder to review.

### What was wrong

Using `random.randint` everywhere when generating seconds can only ever give seconds precision.

### How this fixes it

Fixes #1945 by allowing sub-second on platforms expected to support it.

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -1,3 +1,4 @@
+import platform
 import re
 
 from calendar import timegm
@@ -73,6 +74,21 @@ for name, sym in [
 
 
 class Provider(BaseProvider):
+    # NOTE: Windows only guarantee second precision, in order to emulate that
+    #       we need to inspect the platform to determine which function is most
+    #       appropriate to generate random seconds with.
+    if platform.system() == "Windows":
+
+        @property
+        def _rand_seconds(self):
+            return self.generator.random.randint
+
+    else:
+
+        @property
+        def _rand_seconds(self):
+            return self.generator.random.uniform
+
     centuries: ElementsType[str] = [
         "I",
         "II",
@@ -1814,7 +1830,7 @@ class Provider(BaseProvider):
         """
         start_datetime = self._parse_start_datetime(start_datetime)
         end_datetime = self._parse_end_datetime(end_datetime)
-        return self.generator.random.randint(start_datetime, end_datetime)
+        return self._rand_seconds(start_datetime, end_datetime)
 
     def time_delta(self, end_datetime: Optional[DateParseType] = None) -> timedelta:
         """
@@ -1824,7 +1840,7 @@ class Provider(BaseProvider):
         end_datetime = self._parse_end_datetime(end_datetime)
         seconds = end_datetime - start_datetime
 
-        ts = self.generator.random.randint(*sorted([0, seconds]))
+        ts = self._rand_seconds(*sorted([0, seconds]))
         return timedelta(seconds=ts)
 
     def date_time(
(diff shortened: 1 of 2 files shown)


faker/providers/date_time/__init__.py
```diff
@@ -73,6 +74,21 @@ class ParseError(ValueError):
...
+
+        @property
+        def _rand_seconds(self):
+            return self.generator.random.randint
+
+    else:
+
+        @property
+        def _rand_seconds(self):
+            return self.generator.random.uniform
```
cknv, line 90: This could be done with a function that internally uses `platform.system()` to determine which one is the best function, but I do not expect the platform to change at any time when using faker. Making a small effort to avoid checking the same condition over and over seemed worth it - but the formatting turned out a bit wonky somehow.

fcurella (approved):
Thank you!

mattinbits:
This change caused a problem for some of our tests, since we combine `faker` with `pandas` to create dataframes of fake data. `pandas` uses a different type for a `datetime` depending on it's precision (`datetime64[ms]` vs `datetime64[ns]`). Maybe it's too late since this is merged, but it might be better for this to be `date_time_ns()` or similar rather than changing the behaviour of the existing function?

cknv:
Oh that's not great @mattinbits, I didn't mean to break anything.

My focus was to get something that more closely resembled `datetime.now`, and I guess I didn't think about the case of wanting less precision.

While implementing this I did toy with the idea of being able to pass an argument to define the precision of the timestamp, at the time that was mostly limited to passing either `float` or `int`, but that argument could just as well be `s`, `ms` or `ns` (or something even better that I just haven't thought of).

Ultimately I can perhaps refine this, but some input on what ideas are worth pursuing would probably best.

fcurella:
I think we should what Python does: always return a `float`. On Windows, the decimal pert will always be `0`, eg `1061306726.0`

PR #1947, opened 2023-11-02
Fix pydecimal crash on float min_value or max_value

s-weigand:
### What does this change

This fixes crashes of `Faker().pydecimal` when used with `float` `min_value` and/or `max_value` on python >= 3.10.

```py
>>> From faker import Faker

>>> Faker().pydecimal(min_value=0.2, max_value=1.5)

ValueError: non-integer arg 1 for randrange()
```

### What was wrong

An internal change of `random.randrange` introduced in Python 3.10 prohibits float values (see #1824) 

### How this fixes it

Casting values to `int` before passing them to `random_int` ensures that `random.randrange` gets integers.

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -258,21 +258,21 @@ class Provider(BaseProvider):
 
     def pydecimal(
         self,
-        left_digits=None,
-        right_digits=None,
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

PR #1950, opened 2023-11-14
Ensure unix_time provider returns an integer

karl-potisepp:
Based on the common understanding that a UNIX timestamp is the number of seconds since **00:00:00 UTC on 1 January 1970**, and also the typing information provided in the function signature, this change forces the returned timestamp to integer precision.

### What does this change

This casts the result of **unix_time** into an **integer**.

### What was wrong

unix_time provider was returning **float** values, which is in conflict with the specified return type of **int**. For reference, [here is the Wikipedia page for UNIX time](https://en.wikipedia.org/wiki/Unix_time).

### How this fixes it

This casts the result of **unix_time** into an **integer**.

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -1830,7 +1830,7 @@ class Provider(BaseProvider):
         """
         start_datetime = self._parse_start_datetime(start_datetime)
         end_datetime = self._parse_end_datetime(end_datetime)
-        return self._rand_seconds(start_datetime, end_datetime)
+        return int(self._rand_seconds(start_datetime, end_datetime))
 
     def time_delta(self, end_datetime: Optional[DateParseType] = None) -> timedelta:
         """


stefan6419846:
I am not sure whether changing the type hint to float is the more appropriate solution here, as `time.time` does return a float as well: https://docs.python.org/3/library/time.html#time.time (The stubs state a float as well: https://github.com/python/typeshed/blob/main/stdlib/datetime.pyi)

karl-potisepp:
> I am not sure whether changing the type hint to float is the more appropriate solution here, as `time.time` does return a float as well: https://docs.python.org/3/library/time.html#time.time (The stubs state a float as well: https://github.com/python/typeshed/blob/main/stdlib/datetime.pyi)

I agree that this is also a solution to the issue. Although I would say - from personal experience - that having the UNIX timestamp as an integer type is a very widely spread practice, and this can break a lot of users code (and lead them to come here and make pull requests like this :) ). I don't think I am wrong when I say that calling it **unix_time** creates an expectation of getting back a 32 bit integer.

In any case, I will defer to the decision of the maintainers. Thank you for your support!

fcurella:
My preference is to go with what Python does and be consistent with the language.

PR #1953, opened 2023-11-18
fix: random sign when +ve is false

### LINKED ISSUE #1928: pyfloat ignores the positive kwarg
* Faker version: Faker==19.10.0
* OS: MacOs 13.3.1
* python version: 3.11

[pyfloat](https://faker.readthedocs.io/en/master/providers/faker.providers.python.html#faker.providers.python.Provider.pyfloat) ignores the positive kwarg. When setting the flag to `False` I expect that the fake results will be no `positive`

### Steps to reproduce

```python
import faker

f = faker.Faker()

fake.pyfloat(positive=False)

In [9]: f.pyfloat(positive=False)
Out[9]: 989.306591673515

In [10]: f.pyfloat(positive=False)
Out[10]: 2202.72694698392

In [11]: f.pyfloat(positive=False)
Out[11]: 3824832.83198619

In [12]: f.pyfloat(positive=False)
Out[12]: 25312785206442.4

In [13]: f.pyfloat(positive=False)
Out[13]: 4906363.78374671

In [14]: f.pyfloat(positive=False)
Out[14]: -593.480996281136

In [15]: f.pyfloat(positive=False)
Out[15]: -15794267.4477139
```

### Expected behavior

If I use `positive=False` then I expect all the results should be negative

### Actual behavior

Results are positive and negative

viraj-s15:
### What does this change

When using `print(f.pyfloat(positive=False))` , the number could be positive or negative.
Now when positive is false, only negative numbers would be returned. 

### What was wrong

Both negative and positive being returned. The code also seemed to not be format by isort (or maybe something wrong on my end). I did run isort though.

### How this fixes it

Removed randomly picking a sign if positive is false

Fixes #1928 

Now this could be intended, if it was, maintainers, please let me know, I would love to implement the random sign functionality but that would require changing the function arguments and tests.

#### All the tests pass on python 3.11

image

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -478,9 +478,9 @@ Credits
     :target: https://coveralls.io/r/joke2k/faker?branch=master
     :alt: Test coverage
 
-.. |build| image:: https://github.com/joke2k/faker/workflows/Python%20Tests/badge.svg?branch=master&event=push
-    :target: https://github.com/joke2k/faker/actions?query=workflow%3A%22Python+Tests%22+branch%3Amaster+event%3Apush
-    :alt: Build status of the master branch on Mac/Linux
+.. |build| image:: https://github.com/joke2k/faker/actions/workflows/ci.yml/badge.svg
+    :target: https://github.com/joke2k/faker/actions/workflows/ci.yml
+    :alt: Build status of the master branch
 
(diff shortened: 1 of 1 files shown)


stefan6419846:
I would argue against mixing this lot of formatting changes with your actual fix, as this makes it hard to actually spot the relevant changes.

PR #1954, opened 2023-11-18
fix: random sign when +ve is false(re)

viraj-s15:
### What does this change

When using print(f.pyfloat(positive=False)) , the number could be positive or negative.
Now when positive is false, only negative numbers would be returned.

### What was wrong

Both negative and positive being returned.

### How this fixes it

Removed randomly picking a sign if positive is false

Fixes: #1928 

Now this could be intended, if it was, maintainers, please let me know, I would love to implement the random sign functionality but that would require changing the function arguments and tests.

I did not want to ruin the commit history by making changes to the messed up pr.
Apologies to the maintainers. heres the actual pr.

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -203,7 +203,7 @@ class Provider(BaseProvider):
                 positive,
             )
         else:
-            sign = "+" if positive else self.random_element(("+", "-"))
+            sign = "+" if positive else "-"
             left_number = self.random_number(left_digits)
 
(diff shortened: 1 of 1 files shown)


fcurella:
I think the correct behavior should be:

`positive=True` always return a positive number
`positive=False` always return a negative
`positive=None` (or not specified at all) return random sign

viraj-s15:
> `positive=True` always return a positive number `positive=False` always return a negative `positive=None` (or not specified at all) return random sign

Agreed, will add that.

fcurella (approved):
Thank you!

PR #1962, opened 2023-12-14
feat: Enable generation of windows file paths

michaelCTS:
### What does this change

`fake.file_path()` now has a `window: Optional[bool]` that allows generating windows file paths.

### What was wrong

It was not possible to fake windows paths.

### How this fixes it

Windows paths (both relative and absolute) can be faked.

Fixes #1961

PR #1963, opened 2023-12-16
add support for multiple file systems path rules

parsariyahi:
### What does this change

This solution let the function ``file_path()`` to generate path in any file system rules.

### What was wrong

``file_path()`` was just generating ``unix`` like paths

### How this fixes it

I used a flag that get the ``OS``, includes: ``linux``, ``windows``,
With this solution other file system rules can be implemented.
Fixes #...


stefan6419846:
This seems to be an alternative implementation of #1962.

parsariyahi:
yes, I think this is more flexible and can implement more standards

faker/providers/file/__init__.py
```diff
@@ -269,16 +282,28 @@ def file_path(
...
 
+        if ``file_system`` is set (default="linux"), the generated path uses
+        specified file system path standard, the list of valid file systems include:
+        ``'windows'``, ``'linux'``.
+
         :sample: size=10
         :sample: depth=3
         :sample: depth=5, category='video'
         :sample: depth=5, category='video', extension='abcdef'
+        :sample: depth=5, category='video', extension='abcdef', file_system='windows
```
fcurella, line 293: typo (missing closing quote):

```suggestion
        :sample: depth=5, category='video', extension='abcdef', file_system='windows'
```

faker/providers/file/__init__.py
```diff
@@ -269,16 +282,28 @@ def file_path(
...
         """
+        fs_rule = self.file_systems_path_rules.get(file_system_rule, None)
+        if not fs_rule:
+            raise TypeError("Specified file system is invalid.")
```
fcurella, line 297: Whhy do we declare the argument to be `Optional`, if then we eeror when it's `None`? Should we remove `Optional` from the type hint? The default value would still make it backward compatible.

fcurella:
@parsariyahi In the future, could you please communicate with the PR's original author before submitting an alternative?

faker/providers/file/__init__.py
parsariyahi: > Whhy do we declare the argument to be `Optional`, if then we eeror when it's `None`? Should we remove `Optional` from the type hint? The default value would still make it backward compatible.

for this my idea was something like user missple the ``windows`` word or set an invalid value.
if we don't support or missple the value we throw that error.
and ``Optional`` is for user to know that this param is not required.

parsariyahi:
> @parsariyahi In the future, could you please communicate with the PR's original author before submitting an alternative?

sure, that was my bad, I saw the issue and checked the PRs but at the time there was'nt any, so I coded and didn't check again and submit the PR.

faker/providers/file/__init__.py
fcurella: If we tell users that this defaults to `"linux"`, I don't think it owuld be necessary to have it declared as `Optional`

faker/providers/file/__init__.py
```diff
@@ -279,25 +292,38 @@ def file_path(
...
                 extension = ""
 
-        file: str = self.file_name(category, extension)
-        path: str = f"/{file}"
+        fs_rule = self.file_systems_path_rules.get(file_system_rule, None)
+        if not fs_rule:
+            raise TypeError("Specified file system is invalid.")
+
+        root: str = fs_rule.get("root")
+        seperator: str = fs_rule.get("seperator")
```
fcurella, line 320: Since we're defining the values of the dictionary ourselves, I dont think we need to worry about non-existent keys.

```suggestion
        root = fs_rule["root"]
        seperator = fs_rule["separator"]
```

faker/providers/file/__init__.py
```diff
@@ -262,6 +274,7 @@ def file_path(
         category: Optional[str] = None,
         extension: Optional[Union[str, Sequence[str]]] = None,
         absolute: Optional[bool] = True,
+        file_system_rule: Optional[str] = "linux",
```
fcurella, line 277: Can we specify this as `Literal`?

```suggestion
        file_system_rule: Literal["linux", "windows"] = "linux",
```

faker/providers/file/__init__.py
```diff
@@ -204,6 +204,18 @@ class Provider(BaseProvider):
...
+
+    file_systems_path_rules: Dict[str, Dict] = {
+        "windows": {
+            "root": "C:\\",
+            "seperator": "\\",
+        },
+        "linux": {
+            "root": "/",
+            "seperator": "/",
+        },
```
fcurella, line 216: The correct spelling is `separator`:

```suggestion
        "windows": {
            "root": "C:\\",
            "separator": "\\",
        },
        "linux": {
            "root": "/",
            "separator": "/",
        },
```

fcurella (changes requested):

faker/providers/file/__init__.py
parsariyahi: yes I think this is better type hinting

faker/providers/file/__init__.py
```diff
@@ -262,6 +274,7 @@ def file_path(
         category: Optional[str] = None,
         extension: Optional[Union[str, Sequence[str]]] = None,
         absolute: Optional[bool] = True,
+        file_system_rule: Literal["linux", "windows"] = "linux",
```
fcurella, line 277: Can you import `Literal` at the top of the file?

fcurella (approved):
Thank you!

PR #1966, opened 2023-12-20
Allow extension in file_path to take a list of allowed extensions, or empty for "no extension"

### LINKED ISSUE #1964: Feature request: file_path extension=False to generate a path without extension
```
faker.file_path(extension=None)
```
 will generate something like `/event/amount.js`

### Actual behavior

There is no way to generate a path without an extension. 

### Expected behavior

This is counter-intuitive - `exrtension? None` should mean that there is no extension. 
And it would be useful to have a way to generate a path without an extension, regardless of how it is done.

### Suggestion

In order not to break backward compatibility, can I suggest changing extension to be either None, bool or a string, with the following behaviour

* True -> same as None
* False ->  no extension
* None ->  as it is now 
* a string -> the string

Thanks

gotofritz:
### What does this change

Change to file_path behaviour: allow extension to take a list of allowed extensions, or empty for "no extension". Non-breaking change.

### What was wrong

There was no way to have a path without extension

### How this fixes it

As proposed by @joke2k, passing a list allows to have no extension is list is empty, but it also allows the user to provide a list of candidate extensions from which to pick one.

Fixes #1964

### DIFF
--- a/faker/providers/file/__init__.py
+++ b/faker/providers/file/__init__.py
@@ -291,7 +291,7 @@ class Provider(BaseProvider):
 
         file: str = self.file_name(category, extension)
         if no_extension:
-            file = re.sub(r"\.[^/]+$", "", file)
+            file = file.rsplit(".", 1)[0]
         path: str = f"/{file}"
         for _ in range(0, depth):
             path = f"/{self.generator.word()}{path}"
--- a/faker/providers/file/__init__.py
+++ b/faker/providers/file/__init__.py
@@ -1,8 +1,7 @@
-import re
 import string
 
 from collections import OrderedDict
-from typing import Dict, List, Optional, Union
+from typing import Dict, Optional, Sequence, Union
 
 from .. import BaseProvider, ElementsType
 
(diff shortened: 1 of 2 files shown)


faker/providers/file/__init__.py
```diff
@@ -269,12 +270,28 @@ def file_path(
...
+
         :sample: size=10
         :sample: depth=3
         :sample: depth=5, category='video'
         :sample: depth=5, category='video', extension='abcdef'
+        :sample: extension=[]
+        :sample: extension=["a", "bc", "def"]
         """
+        no_extension = False
+        if isinstance(extension, list):
```
fcurella, line 285: I'd like for `extension` to also possibly a tuple.
We could type the argument with `extension: Optional[Sequence[str]]`, and then check for `not isinstance(extension, str)` instead:

```suggestion
        if not isinstance(extension, str):
```

I would also change the behavior of `file_name`, so that if the empty string is passed, no extension and no dot are added. This would allow us to pass `""` to `file_name` and avoid stripping it out later.

faker/providers/file/__init__.py
```diff
@@ -269,12 +270,28 @@ def file_path(
...
+            if extension:
+                extension = self.random_element(extension)
+            else:
+                extension = None
+                no_extension = True
+
         file: str = self.file_name(category, extension)
+        if no_extension:
+            file = re.sub(r"\.[^/]+$", "", file)
```
fcurella, line 294: If we have to strip, I'd rather use `rsplit()`:

```suggestion
            file = file.rsplit(".", 1)[0]
```

faker/providers/file/__init__.py
gotofritz: Done - also checking for None

fcurella (approved):
Thank you!

Issue #1970, opened 2024-01-09
Add fa_IR for currency


parsariyahi:
In ``providers.currency`` we don't have ``IRR`` the persian currency.
if you are interested I would be happy to submit a PR for it

stefan6419846:
Feel free to open a PR if it is missing.

fcurella:
Merged in #1971

PR #1971, opened 2024-01-10
Add fa_IR localization for currency provider

parsariyahi:
### Add ``fa_IR`` localization in ``currency`` provider

``IRR`` is Iran main currency and looks something like ``1,000,000,000 IRR``,
This PR will add better localize faker for Iran currency.


parsariyahi:
This PR is related to #1970

fcurella (approved):
Thank you!

PR #1973, opened 2024-01-12
add bank uk_UA

SanderFtn:
add provider.bank for uk_UA localization


stefan6419846:
Could you please add corresponding tests as well?

SanderFtn:
Yes.
Add some tests with new commit.

tests/providers/test_bank.py
```diff
@@ -136,6 +137,21 @@ def test_iban(self, faker, num_samples):
             assert re.fullmatch(r"\d{2}\d{24}", iban[2:])
 
 
+class TestUkUA:
```
fcurella, line 140: The capitalization is actually meaningful here, because of how we run tests.

```suggestion
class TestUkUa:
```

fcurella (changes requested):

fcurella (approved):
Thank you!

SanderFtn:
Thank you too!

PR #1976, opened 2024-01-18
Add providers for ms_MY

ngbk1993:
### What does this change
Add providers for ms_MY for the address,
person, phone_number and ssn.

This is porting from the faker php
https://github.com/fzaninotto/Faker

Tested with git diff --check and the code is formatted with black --line-length 120

### What was wrong

The ms_MY provider is missing.

### How this fixes it

Add the ms_MY provider for address, person, phone_number and ssn


faker/providers/person/ms_MY/__init__.py
```diff
@@ -0,0 +1,5482 @@
...
+        "{{initialIndian}}{{firstNameFemaleIndian}}",
+        "{{initialIndian}}{{lastNameIndian}}",
+        "{{firstNameFemaleIndian}} a/l {{firstNameMaleIndian}}",
+        "{{firstNameFemaleIndian}} a/l {{firstNameMaleIndian}} {{lastNameIndian}}",
+        "{{firstNameFemaleIndian}} {{lastNameIndian}} a/l {{firstNameMaleIndian}}",
+        "{{firstNameFemaleIndian}} {{lastNameIndian}} a/l {{firstNameMaleIndian}} {{lastNameIndian}}",
+        "{{firstNameFemaleIndian}} {{lastNameIndian}}",
+    )
+
+    firstNameMaleMalay_tuple = (
```
fcurella, line 74: Can you add sources for this data?

fcurella (changes requested):
Can you add sources for the dat, and tests?

faker/providers/person/ms_MY/__init__.py
ngbk1993: Hi @fcurella, they are ported from the php faker https://github.com/fzaninotto/Faker/blob/master/src/Faker/Provider/ms_MY/Person.php

can i just include the github link where it is ported it from?

ngbk1993:
> Can you add sources for the dat, and tests?

I can add some tests to it, but how do i run the test locally? can share the step?

faker/providers/person/ms_MY/__init__.py
fcurella: You can follow the instructions [here](https://github.com/joke2k/faker/blob/master/README.rst#tests)

dancergraham:
> > Can you add sources for the dat, and tests?
> 
> I can add some tests to it, but how do i run the test locally? can share the step?

I had to `python -m pip install pytest freezegun validators` and then `python -m pip install .` after each change before running pytest

stefan6419846:
In theory it should be easier as documented in the docs at https://github.com/joke2k/faker/blob/master/CONTRIBUTING.rst: Just install `tox` via *pip* and then just use the `tox` command which should take care of the remaining stuff.

faker/providers/person/ms_MY/__init__.py
fcurella: Yes, that would work 👍🏼

ngbk1993: sure, let me add the faker php link as the source. will update at the next version with the tests file.

stefan6419846:
> Was trying to dry run the tox but somehow i'm getting this output not sure if it is expected

Yes, this is expected. What *tox* does here basically is to iterate over the configured Python versions and if it finds the corresponding interpreter, will create an isolated environment, install the required packages there and run the necessary steps like tests there.

> yes, the pip install step was somehow missing from the CONTRIBUTING.rst, we cant assume that everyone has the tox module installed.

Feel free to submit a corresponding PR to improve the docs if you miss something there or something tends to be unclear and could be improved.

ngbk1993:
Hi @fcurella, was busy with my work, will try to get it in by this month, can help to reopen this PR?

PR #1978, opened 2024-01-19
Area code, add 830000 and parameterized

antik-x:
### What does this change

In the `zh_CN` provider for SSN , the area code is enhanced by adding `830000` and making it parameterized.

### What was wrong

The area code `830000` has been officially released by www.gov.cn. During actual testing, it is necessary to generate SSN in the specified region.

### How this fixes it

Firstly, `830000` is added to the `area_codes` list.
Then, in the `ssn()` method, a new parameter `area_code` is added, with the default value being `None`. 
If the passed `area_code` value exists in the `area_codes` list, it will be used to generate an SSN that meets the requirements. Otherwise, one of the codes from `area_codes` will be randomly selected to generate the SSN.


fcurella (changes requested):
Thank you! Could you run `make lint` to format the code according to our code style?

fcurella (approved):
Thank you!

PR #1982, opened 2024-01-24
"adding height and weight for person provider"

boopathinmc7217:
### What does this change

Added Height and Weight for the person provider

### What was wrong

-

### How this fixes it

-

### DIFF
--- a/faker/providers/person/__init__.py
+++ b/faker/providers/person/__init__.py
@@ -329,19 +329,30 @@ class Provider(BaseProvider):
         """Generate a random i18n language name (e.g. English)."""
         return self.random_element(self.language_names)
 
-    def height(self) -> str:
+    def height(self, fmt: str = "ft") -> str:
         """
         Generate random height for person in feet and inches
 
-        Example :
-            4'5"
-        """
-        return f"{self.random_number(digits=1)}'{self.random_number(digits=2)}\""
+        Arguments:
+        ----------
+            fmt = format of the data, by default it will be taken as ft and inch
 
-    def weight(self) -> str:
+        """
+        if fmt == "cm":
+            return f"{self.random_number(digits=2)} cms"
+        else:
+            return f"{self.random_number(digits=1)}'{self.random_number(digits=2)}\""
+
+    def weight(self, fmt: str = "lbs") -> str:
         """
         Generate random weight for the person
-        Example :
-        140 lbs
+
+        Arguments:
+        ---------
+            fmt = format of the data, by default it will be taken as lbs
         """
-        return f"{self.random_number(digits=3)} lbs"
+        weight = self.random_number(digits=3)
+        if fmt == "kg":
+            return f"{weight} kg"
+        else:
+            return f"{weight} lbs"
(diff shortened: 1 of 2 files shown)


faker/providers/person/__init__.py
```diff
@@ -328,3 +328,31 @@ def suffix_female(self) -> str:
...
+        """
+        Generate random height for person in feet and inches
+
+        Arguments:
+        ----------
+            fmt = format of the data, by default it will be taken as ft and inch
+
+        """
+        if fmt == "cm":
+            return f"{self.random_number(digits=2)} cms"
```
stefan6419846, line 342: Does the trailing *s* really make sense here?

faker/providers/person/__init__.py
```diff
@@ -328,3 +328,31 @@ def suffix_female(self) -> str:
...
+        else:
+            return f"{self.random_number(digits=1)}'{self.random_number(digits=2)}\""
+
+    def weight(self, fmt: str = "lbs") -> str:
```
stefan6419846, line 346: I would argue the same as for the height regarding parameter handling here.

tests/providers/test_person.py
```diff
@@ -1331,5 +1332,28 @@ def test_full_name(self):
...
+        """
+        Test  height generation for person
+        """
+        self.assertTrue(hasattr(self.provider, "height"))
+        self.assertRegex(self.fake.height(), r"\d'\d+\"")
+        self.assertRegex(self.fake.height("cm"), r"\d{1,3}\scms")
+
+    def test_weight(self):
+        """
+        Test  weight generation for person
```
stefan6419846, line 1351:
```suggestion
        Test weight generation for person
```

faker/providers/person/__init__.py
```diff
@@ -328,3 +328,31 @@ def suffix_female(self) -> str:
     def language_name(self) -> str:
         """Generate a random i18n language name (e.g. English)."""
         return self.random_element(self.language_names)
+
+    def height(self, fmt: str = "ft") -> str:
```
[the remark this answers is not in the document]
fcurella: the unit should be according to the locale. If a locale only uses one unit (eg: Italy only uses Metric), I wouldn't even have the parameter, and reserved it for those locales that use both.

boopathinmc7217:
> I'm not sure if it's appropriate to have this in `person`. Could it be a more generic "measures" provider with `weight(min, max)` and `length(min, max)`?

Agreed

faker/providers/person/__init__.py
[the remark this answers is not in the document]
boopathir5626: will look into a way to use the locale

boopathir5626:
> I'm not sure if it's appropriate to have this in `person`. Could it be a more generic "measures" provider with `weight(min, max)` and `length(min, max)`?

Agreed, then we can support the measures

faker/providers/person/__init__.py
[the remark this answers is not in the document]
boopathinmc7217: will look into way for using current locale, for using the locale i hope we need to maintain a ref

boopathinmc7217:
> I'm not sure if it's appropriate to have this in `person`. Could it be a more generic "measures" provider with `weight(min, max)` and `length(min, max)`?

Agreed , we can support the measurements ,

faker/providers/person/__init__.py
dclobato: > Does the trailing _s_ really make sense here?

No.  There is no plural for it...

PR #1983, opened 2024-01-28
feat(ssn)Add support uk_Ua ssn method birthday and gender optional pa…

lozik4:
### Now the ssn for the uk_Ua locale can generate values for a specific gender and date of birth. 

Add optional params birthdate and gender. 
Customize logic generation and write tests.

### What was wrong

Many validators in Ukraine check the ssn together with the date of birth, the previous implementation does not allow to pass validation, because the method takes a random date of birth. Gender was not taken into account at all.
Now it is possible to generate the desired value more accurately. If this is not necessary, the old logic will work.

### DIFF
--- a/tests/providers/test_ssn.py
+++ b/tests/providers/test_ssn.py
@@ -1367,7 +1367,7 @@ class TestZhTW(unittest.TestCase):
             assert tw_checksum(sample) % 10 == 0
 
 
-class TestUkUa(unittest.TestCase):
+class TestUkUA(unittest.TestCase):
     def setUp(self):
         self.fake = Faker("uk_Ua")
         Faker.seed(0)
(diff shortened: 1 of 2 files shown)


tests/providers/test_ssn.py
```diff
@@ -1364,3 +1365,30 @@ def test_gender(self):
     def test_checksum(self):
         for sample in self.samples:
             assert tw_checksum(sample) % 10 == 0
+
+
+class TestUkUa(unittest.TestCase):
```
fcurella, line 1370: please keep the capitalization of test cases consistent:

```suggestion
class TestUkUA(unittest.TestCase):
```

fcurella (approved):
Thank you!

PR #1985, opened 2024-01-31
sphinx conf: keep copyright year up-to-date

thijstriemstra:
### What does this change

Add current year to generated docs, useful to determine if the docs are recent.

uses py2.6 compatible syntax (if needed)

ps. congrats on 10 years!


fcurella (approved):
Thank you!

thijstriemstra:
I'd expected the change to show up on rtd by now but it seems it didn't update, any idea @fcurella?

PR #1986, opened 2024-02-04
feat(uk_UA-automotive) Add uk_UA automotive provider

lozik4:
### What does this change

Add Uk_UA Automotive provider

### What was wrong

This provider was absent  in  uk_UA locale


fcurella (approved):
Thank you!

PR #1988, opened 2024-02-06
Consistent error handling for mis-ordered start-dates and end-dates

prescod:
### What does this change

In the past, Faker would throw an error if the start of a daterange was after the end. This is still the behaviour on Windows but not Unix.

### What was wrong

Python's random module is itself inconsistent about mis-ordered arguments to random.randint versus random.uniform. This bubbled up to a different behaviour on Windows and Unix because those different functions are used depending on the platform.

### How this fixes it

This PR ensures that on both Windows and Unix, an exception is thrown. This is "backwards compatible" to older versions of Faker and compatible across platforms. It is not backwards compatible to the very most recent versions of Faker, when running on Unix, because it re-introduces the exception.

On all platforms, the exception is a ValueError that contains the string "empty range" and this can be used by callers to detect the problem. Snowfakery, for example, has code to detect this problem and report it nicely to users.

Fixes #...

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -79,15 +79,17 @@ class Provider(BaseProvider):
     #       appropriate to generate random seconds with.
     if platform.system() == "Windows":
 
-        @property
-        def _rand_seconds(self) -> Callable[[int, int], int]:
-            return self.generator.random.randint
+        def _rand_seconds(self, start_datetime: int, end_datetime: int) -> float:
+            return self.generator.random.randint(start_datetime, end_datetime)
 
     else:
 
-        @property
-        def _rand_seconds(self) -> Callable[[float, float], float]:
-            return self.generator.random.uniform
+        def _rand_seconds(self, start_datetime: int, end_datetime: int) -> float:
+            if start_datetime > end_datetime:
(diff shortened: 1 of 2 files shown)


fcurella:
Do we know the reason for inconsistency in Python?

prescod:
Maybe people coming from a statistical/simulation background are more accustomed to the arguments being interchanger. And uniform is designed for them. Whereas people coming from a Python "range" behaviour background are more accustomed to the arguments being strictly ordered.

That's roughly what a couple of LLMs said but of course they are just guessing as I am.

https://chat.openai.com/share/1f311605-ce61-4b5b-b702-ba8c41e6833a

For the record, I find it a bit weird that the faker library behaves differently on different platforms by design. If I'm generating a datettime, that datetime is usually representing data that came from a database or JSON and the database or JSON could have been populated from code running on any operating system.

I'm having trouble imagining the situation where I'm running Faker on Windows and it breaks my code that the datettimes are "too precise.". But maybe you really saw such a case.

fcurella (approved):
The way I see, `faker` is making the "precision switch" according to the platform, and so `faker` should be responsible for providing consistent behavior of its own abstractions.

PR #1994, opened 2024-02-14
fix: update `pydecimal` algorithm to ensure left part is not generated with a leading 0

### LINKED ISSUE #1992: `pydecimal(left_digits=14)` produced value with 13 left digits
* Faker version: 22.6.0
* OS: macOS, Linux

I've encountered a situation where `pydecimal` appears to not respect `left_digits`:

```python
from faker import Faker

fake = Faker()
Faker.seed(0)

fake.pydecimal(left_digits=14, right_digits=2)  # Decimal("5550397399547.96")
```

"5550397399547.96" has 13 left digits. How is this possible?

Note I found this while running a suite of a couple of thousands of tests. One of the tests failed because the value produced did not have the expectated number of digits. If I run the failing test in isolation, it passes i.e. that specific value doesn't get generated. I'm only getting this when I run the whole suite.

alexei:
### What does this change

This changes the algorithm of `pydecimal` to ensure the integer part if of the specified length.

### What was wrong

As suggested in https://github.com/joke2k/faker/issues/1992, it is possible for `pydecimal` to generate a value with an integer part that's shorter than the specified length.

### How this fixes it

The original algorithm generated a list of digits and joined it into an integer. However, with that approach it is possible that one or more leading digits are 0, thus ending up with a shorter value.

For example,

```python
from faker import Faker
fake = Faker()
fake.seed_instance(15)
fake.pydecimal(left_digits=8, right_digits=3)
```

Generates `08023002` for left number, thus resulting in `Decimal("8023002.531")`. Since `8023002` is only 7 digits, it does not satisfy the requirement that the resulting value has 8 left digits.

Fixes #1992

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -262,6 +262,21 @@ class Provider(BaseProvider):
     def pyint(self, min_value: int = 0, max_value: int = 9999, step: int = 1) -> int:
         return self.generator.random_int(min_value, max_value, step=step)
 
+    def _random_int_of_length(self, length: int) -> int:
+        """Generate a random integer of a given length
+
+        If length is 0, so is the number. Otherwise the first digit must not be 0.
+        """
+
+        if length < 0:
+            raise ValueError("Length must be a non-negative integer.")
+        elif length == 0:
+            return 0
+        else:
+            min_value = 10 ** (length - 1)
+            max_value = (10**length) - 1
+            return self.pyint(min_value=min_value, max_value=max_value)
+
     def pydecimal(
         self,
         left_digits: Optional[int] = None,
@@ -310,7 +325,7 @@ class Provider(BaseProvider):
                 min_left_digits = math.ceil(math.log10(max(min_value or 1, 1)))
                 if left_digits is None:
                     left_digits = self.random_int(min_left_digits, max_left_random_digits)
-                left_number = "".join([str(self.random_digit()) for i in range(0, left_digits)]) or "0"
+                left_number = str(self._random_int_of_length(left_digits))
         else:
             if min_value is not None:
                 left_number = str(self.random_int(int(max(max_value or 0, 0)), int(abs(min_value))))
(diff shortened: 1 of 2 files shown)


alexei:
I fear the tests I added are only relevant to the original algorithm. Let me know how to go about this.

fcurella:
The issues with hardcoding a specific seed in tests is that different versions of Python may return a different value. What we usually do is set the seed to `0`, and then run the same test for some number of iteration, and  hope for the best 🤞🏼

alexei:
I did try that initially and I gave up after a couple of million iterations. @stefan6419846 [suggested](https://github.com/joke2k/faker/issues/1992#issuecomment-1943828856) to try to find a seed and did the trick. I also noticed there were other tests that seeded the generator so I figured it was harmless.

However, as I pointed out in my previous comment, the tests are only relevant with the previous technique where digits are generated individually. With this change an integer is generated as a whole, so this bug would no longer exist. I would say it's not worth keeping them. Besides, the number of digits is already checked in an earlier test. These two are just to reproduce the bug I reported. What do you think?

stefan6419846:
My initial proposal with a fixed seed is mostly helpful for doing local debugging on your setup as once you have found a suitable seed, digging through the implementation and analyzing where the issue stems from makes the most sense. This does not necessarily need to hold true as mentioned in the previous comments in this PR for actual unit/integration tests.

alexei:
There are other tests doing the same: https://github.com/joke2k/faker/blob/1dc888b36cbdcd91c31bbcbe71d34267acdd5898/tests/providers/test_python.py#L444-L467

In any case, this is focusing on the wrong question. I think I should drop the tests, as they're not necessary. If you think the tests should stay, then please either guide me so I can rewrite them in a way that fits the codebase, or change them yourselves.

fcurella:
I wonder if we could mock the random generator on the instance to force it
to return a fixed value. I can take a look into it later today

fcurella:
ok, I've looked into this and now I have a better understadning of what you're saying. I'd say let's drop the tests

alexei:
@fcurella thanks for taking a look; I dropped the tests

fcurella (approved):
Thank you!

PR #1997, opened 2024-02-24
Fix for #1993: Locale et_EE (Estoniana) with seed_instance returns random values instead of fixed upon restart

### LINKED ISSUE #1993: Locale et_EE (Estoniana) with seed_instance returns random instead of fixed
* Faker version: 23.1.0
* OS: Windows, Mac

Brief summary of the issue:
When using Faker with the locale et_EE and seed_instance the class returns are random. When using a different locale like nl_NL the returns are fixed. When running the script below twice the output names are Marina Sanne, when running a second time the names should be the same instead, it returns Reet Sanne.

### Run script below two times
```
from faker import Faker
locale = ["et_EE"]

fakers = Faker(locale)
fakers.seed_instance(0)

print(fakers.first_name_female()) 

locale = ["nl_NL"]

fakers = Faker(locale)
fakers.seed_instance(0)

print(fakers.first_name_female())

```
### Expected behavior

Marina
Sanne

Marina
Sanne

### Actual behavior

Marina
Sanne

Reet
Sanne

AlexLitvino:
### What does this change

Fix makes Faker return data repeatedly the same when seed is set to the same value on several runs.
Lists for first_name_female, first_name_male, first_name, prefix, last_name for et_EE locale are made from tuples by converting them to sets, uniting them and converting them to list.
Conversion to sets are required because some initial tuples contains common elements (like "Anna" are in both first_names_female_est and first_names_female_rus tuples).
But when sets union converted to list (like `first_names_female = list(set(first_names_female_est + first_names_female_rus))`) there's no guarantee of order elements in list (try to run several times `list({'a', 'b', 'c'})` resulting list will be different).
That's why with every new run, even seed was set to the same value, initial lists were different and that's why different values were produced.
To fix this issue changed `list` to `sorted` to guarantee that lists will have the same elements in the same order on every run.

On current master the code below returns different values when re-run script though seed set to 0. On fix branch it returns the same values:
```
from faker import Faker
fake = Faker("et_EE")
fake.seed_instance(0)
print(fake.first_name_female())
print(fake.first_name_male())
print(fake.first_name())
print(fake.prefix())
print(fake.last_name())
```

Also changed line `first_names = list(set(first_names_male) | set(first_names_female))` to `first_names = sorted(set(first_names_male + first_names_female))` to be consistent with other list creation.

### What was wrong

For locale et_EE, when seed_instance(0) was called, with each rerun, Faker returns different data for first_name_female, first_name_male, first_name, prefix, last_name (described in [#1993](https://github.com/joke2k/faker/issues/1993))

### How this fixes it

Fixes #1993 by preserving order in lists: prefixes, first_names_male, first_names_female, first_names, last_names.
Checked that it was the only locale where set to list conversion was used.

### DIFF
--- a/faker/providers/person/et_EE/__init__.py
+++ b/faker/providers/person/et_EE/__init__.py
@@ -33,7 +33,7 @@ class Provider(PersonProvider):
     prefixes_neutral = ("doktor", "dr", "prof")
     prefixes_male = ("härra", "hr") + prefixes_neutral
     prefixes_female = ("proua", "pr") + prefixes_neutral
-    prefixes = list(set(prefixes_male + prefixes_female))
+    prefixes = sorted(set(prefixes_male + prefixes_female))
 
     suffixes = ("PhD", "MSc", "BSc")
 
@@ -162,9 +162,9 @@ class Provider(PersonProvider):
 
     first_names_rus = first_names_male_rus + first_names_female_rus
 
-    first_names_male = list(set(first_names_male_est + first_names_male_rus))
-    first_names_female = list(set(first_names_female_est + first_names_female_rus))
-    first_names = list(set(first_names_male) | set(first_names_female))
+    first_names_male = sorted(set(first_names_male_est + first_names_male_rus))
+    first_names_female = sorted(set(first_names_female_est + first_names_female_rus))
+    first_names = sorted(set(first_names_male + first_names_female))
 
     # http://ekspress.delfi.ee/kuum/\
     # top-500-eesti-koige-levinumad-perekonnanimed?id=27677149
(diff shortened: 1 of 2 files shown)


fcurella (changes requested):
Could you add some tests to prevent regressions?

AlexLitvino:
@fcurella , added tests for methods exposed from Person provider for et_EE locale.
Tried add test to cover initial issue, but it wasn't reproduced using auto test (on code before fix, I got female name, then reload faker.providers.person.et_EE module and then got female name again, in this case names were the same).

fcurella (approved):
Thank you!

PR #1998, opened 2024-02-26
Add Sk_SK provideres for Person & Bank

mhandl:
### What does this change

Add sk_SK providers for Person & Bank

### What was wrong

 The sk_SK providers were missing for Person & Bank

### How this fixes it

New providers were written.


fcurella (approved):
Thank you!

PR #1999, opened 2024-02-26
feat(uk_UA-credit-cards) Add Uk-UA credit card provider

lozik4:
### Add UA Credit card provider
Add tests
Add bank list
Add translit function


fcurella (approved):
Thank you!

PR #2003, opened 2024-03-02
feat(geo) Add Grenada geo #2001

### LINKED ISSUE #2001: Updating land_coords used in local_latlng() to cover every supported country code
* Faker version: 23.3.0
* OS: macOS Ventura 13.6.4

When using faker.local_latlng method I thought that it has exported locations from geonames.org for every possible country code (which can be generated by faker itself), but it turned out that we have a list of possible locations in 
`faker/faker/providers/geo/__init__.py::Provider:land_coords` 

Hard-coded values are not the problem though, but the lack of locations for some countries (like mentioned below Grenada (GD)). 

Is it possible to update land_coords with additional locations?

### Steps to reproduce

x= faker.local_latlng(country_code="GD")

### Expected behavior

Something similar to:
X is ('12.722', '-61.3729', 'Grenville', 'GD', 'Grenada')

### Actual behavior

X is None

lozik4:
Add Grenville
change Kiev to Kyiv (the correct name capital of Ukraine)

### How this fixes it

add ("12.12278", "-61.62498", "Grenville", "GD", "America/Grenada"),

Fixes #2001


fcurella (approved):
Thank you!

ghostishev:
Thank you for fixing Kyiv!
I'd like to fix Zaporizhzhia as well, how do you validate it?

lozik4:
> Thank you for fixing Kyiv! I'd like to fix Zaporizhzhia as well, how do you validate it?
@ghostishev 

You will create PR, but in reference [wiki](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones)  Europe/Zaporozhye
General All Timezones library uses IANA timezone db
in this DB Zaporizhzhia is not fixed now.

```
import pytz

if __name__ == '__main__':
    for timezone in pytz.all_timezones:
        print(timezone)
```

PR #2004, opened 2024-03-07
Typo fix - prefix for male BG names

DimitarVanguelov:
The Bulgarian person provider used `prefix_female` instead of `prefix_male` within `formats_male`.

### DIFF
--- a/faker/providers/person/bg_BG/__init__.py
+++ b/faker/providers/person/bg_BG/__init__.py
@@ -1759,7 +1759,7 @@ class Provider(PersonProvider):
 
     formats_male = (
         "{{first_name_male}} {{last_name_male}}",
-        "{{prefix_female}} {{first_name_male}} {{last_name_male}}",
+        "{{prefix_male}} {{first_name_male}} {{last_name_male}}",
     )
 
     formats = formats_male + formats_female


lozik4 (approved):

lozik4:
@DimitarVanguelov maybe you will add some tests?

lozik4:
> Hi @lozik4, I'd be happy to contribute but could use some guidance as to how to test for this sort of thing :)

@DimitarVanguelov
For example
```
class TestBgBg(unittest.TestCase):
    def setUp(self):
        self.fake = Faker("bg-BG")
        Faker.seed(0)

    def test_name(self):
        for _ in range(100):
            name = self.fake.name_male()
```
name will be used template, prefix must be in prefix_male
Your provider doesn't have anything tests. You may write some tests in test_person.py file. 
In this file lot of test in another localization, can be examples for you.

DimitarVanguelov:
Thanks, I will look into this. 

> Your provider doesn't have anything tests.

Just in case there is any confusion, I did not write this provider, I was just perusing the code and found the typo.

In any case, thanks again and Слава Україні!

fcurella (changes requested):
Thank you!

PR #2005, opened 2024-03-13
Add more accepted formats

maximegmd:
### What does this change

Adds a couple formats for belgians numbers.

### What was wrong

Missing commonly used formats.


fcurella (approved):
Thank you!

PR #2010, opened 2024-03-16
Add surname de_DE

george0st:
Add addition surname for de_DE


fcurella:
Since the list of last names does not have a source, where are you getting your last names from?

george0st:
@fcurella , I added sources to the code

fcurella:
I don't doubt those last names exist, but I need a source for the _entire_ list. That we won't get a Pull Request every time someone thinks of a last name!

george0st:
@fcurella , I added list of all DE surnames (it contains also these two lasts)

fcurella (approved):
Thank you!

PR #2013, opened 2024-03-21
Add providers

reichie020212:
### What does this change

Add providers for the locales si_LK, ta_LK, and lo_LA.

### What was wrong

Faker doesn't have configuration for si_LK, ta_LK, and lo_LA yet.

### How this fixes it

Add providers for si_LK, ta_LK, and lo_LA.


stefan6419846:
Could you please add the list sources as well?

reichie020212:
> Could you please add the list sources as well?

we wrote it ourself since we have colleagues in Laos and Sri Lanka

fcurella:
@reichie020212 Could you do some research and see if Laos and/or Sri Lanka have any government source we could use instead (eg a Census)?

Issue #1604, opened 2022-02-03
Tab completion in interactive shell not working since Python version 3.8


CelestialGuru:
Tab autocomplete in python interactive shell unavailable when using Python versions 3.9, 3.10 or 3.11.

Python documents this feature: https://docs.python.org/3.11/tutorial/interactive.html#tab-completion-and-history-editing. (Docs here are the same in all versions of Python)

## Setup

* Faker version: 17.3.0
* OS: Debian GNU/Linux 11 (bullseye)
<details>
<summary>OS versions</summary>

    PS C:\Users\User> docker run -it --rm python:3.8 cat /etc/os-release
    PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
    NAME="Debian GNU/Linux"
    VERSION_ID="11"
    VERSION="11 (bullseye)"
    VERSION_CODENAME=bullseye
    ID=debian
    HOME_URL="https://www.debian.org/"
    SUPPORT_URL="https://www.debian.org/support"
    BUG_REPORT_URL="https://bugs.debian.org/"

    PS C:\Users\User> docker run -it --rm python:3.9 cat /etc/os-release
    PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
    NAME="Debian GNU/Linux"
    VERSION_ID="11"
    VERSION="11 (bullseye)"
    VERSION_CODENAME=bullseye
    ID=debian
    HOME_URL="https://www.debian.org/"
    SUPPORT_URL="https://www.debian.org/support"
    BUG_REPORT_URL="https://bugs.debian.org/"

    PS C:\Users\User> docker run -it --rm python:3.10 cat /etc/os-release
    PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
    NAME="Debian GNU/Linux"
    VERSION_ID="11"
    VERSION="11 (bullseye)"
    VERSION_CODENAME=bullseye
    ID=debian
    HOME_URL="https://www.debian.org/"
    SUPPORT_URL="https://www.debian.org/support"
    BUG_REPORT_URL="https://bugs.debian.org/"

    PS C:\Users\User> docker run -it --rm python:3.11 cat /etc/os-release
    PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
    NAME="Debian GNU/Linux"
    VERSION_ID="11"
    VERSION="11 (bullseye)"
    VERSION_CODENAME=bullseye
    ID=debian
    HOME_URL="https://www.debian.org/"
[...]

PR #2015, opened 2024-03-27
Typing-only Mixin Class for Type Checking

H00N24:
### What does this change

While debugging a typing-related bug in my tests I created a small mixin of some Faker providers for mypy type checking, unfortunately when I added all of them for the future checks I noticed a type/name conflict between address and bank/date_time providers. 

### What was wrong

As Faker proxy is dynamically gathering methods from locals, mypy cannot determine the correct types.

### How this fixes it

This PR adds a typing-only class that is a mixture of all available providers thus giving mypy and IDEs enough information for type checking/completion (potentially fixing https://github.com/joke2k/faker/issues/1604).

### Problems

As some providers have common attributes with different datatypes direct mixing was not possible which makes this change backward incompatible. I would love to hear your thoughts on this approach. Thanks.


fcurella:
Hi @H00N24 

With https://github.com/joke2k/faker/pull/1939, is this still an issue?

PR #2017, opened 2024-04-04
Remove duplicate assigned HTTP Status codes

153957:
### What does this change

Remove duplicate entries from `http_assigned_codes`.

### What was wrong

Two values appeared twice in the list skewing the probabilities and making the list harder to maintain.

### How this fixes it

Removed duplicate entries.

### Remark

Would it be acceptable to replace this hard coded list by values from the `http.HTTPStatus` enum included in the Python standard library? (I understand if this repo prefers hard coded lists)
Also; that enum has changed in recent Python version, Python 3.9 added 3 additional status codes, including `418 – IM A TEAPOT` which is not part of the current list because it is considered [Unused](https://www.rfc-editor.org/rfc/rfc9110.html#name-418-unused) by the RFC.

### DIFF
--- a/faker/providers/internet/__init__.py
+++ b/faker/providers/internet/__init__.py
@@ -134,8 +134,6 @@ class Provider(BaseProvider):
     http_assigned_codes: ElementsType[int] = (
         100,
         101,
-        100,
-        101,
         102,
         103,
         200,


fcurella:
I'd rather have a static list, so our dataset doesn't change across Python versions.

fcurella (changes requested):
Thank you! Could you rebase against the current `master` branch?

fcurella (approved):
Thank you!

Issue #2019, opened 2024-04-05
`tests/providers/test_phone_number.py::TestSkSk::test_phone_number` test fails


mgorny:
* Faker version: 24.7.1
* OS: Gentoo Linux amd64

The test `tests/providers/test_phone_number.py::TestSkSk::test_phone_number` fails. Apparently, it fails since it was introduced in #2012 but CI never run tests because of formatting issues, and it was merged and then released without actually testing.

CC @fcurella, @george0st 

### Steps to reproduce

1. `tox`

### Expected behavior

Tests passing.

### Actual behavior

```pytb
============================================================== FAILURES ===============================================================
_____________________________________________________ TestSkSk.test_phone_number ______________________________________________________

self = <tests.providers.test_phone_number.TestSkSk object at 0x7fd74a8beb10>, faker = <faker.proxy.Faker object at 0x7fd74a78ad10>
num_samples = 100

    def test_phone_number(self, faker, num_samples):
        pattern: Pattern = re.compile(r"^(00421|\+421)? ?[2-9][0-9]{2} ?[0-9]{3} ?[0-9]{3}$")
        for _ in range(num_samples):
            phone_number = faker.phone_number()
>           assert pattern.fullmatch(phone_number)
E           AssertionError: assert None
E            +  where None = <built-in method fullmatch of re.Pattern object at 0x55805ba0e7a0>('+421 46 048 7647')
E            +    where <built-in method fullmatch of re.Pattern object at 0x55805ba0e7a0> = re.compile('^(00421|\\+421)? ?[2-9][0-9]{2} ?[0-9]{3} ?[0-9]{3}$').fullmatch

tests/providers/test_phone_number.py:83: AssertionError
```

PR #2020, opened 2024-04-05
Repair unit test for phone number sk_SK

george0st:
Repair unit test for phone number sk_SK see #2019


george0st:
NOTE: I repaired the code, but it seems that tests are executed with old code

fcurella (approved):
I'll figure it out, thank you!

PR #2021, opened 2024-04-10
Update list of first_name_* and last_name on the PersonProvider for p…

dclobato:
### What does this change

Update list of first_name_* and last_name on the PersonProvider for pt_BR

### What was wrong

List of names from 2013/2014

### How this fixes it

Add names from a lista of 2023


fcurella (approved):
Thank you!

PR #2025, opened 2024-04-12
Remove a few mistakes (duplicities also) in job sk_SK

george0st:
Remove a few mistakes (duplicities also) in job sk_SK


fcurella (approved):
Thank you!

PR #2026, opened 2024-04-12
Added avoid_int feature

benjfield:
### What does this change

I have added an argument called avoid_int to pyint that will generate numbers in the range that are different from avoid_int.

### What was wrong

When running tests, it can be useful to have data generated other than a specific value for negative tests. I believe this was not generally possible with faker.

### How this fixes it

If avoid_int is not none, pyint will not generate the value specified.


fcurella:
This looks like a _very_ specific use case. I'm not convinced it's common enough to be worth the additional complexity.

PR #2030, opened 2024-04-25
drop python<=3.7 support

kloczek:
### What does this change
According to https://endoflife.date/python python 3.7 has been EOSed at 7 Jun 2023.
- filter all code over `pyupgrade --py38-plus`.
- manual cleanups removing some older python version code
- filter all code over ruff to remove unused imports


fcurella (approved):
Thank you! ✨

PR #2047, opened 2024-05-15
Overload hash generation functions to make return types more specific

### LINKED ISSUE #2046: Type annotations on `md5`, `sha1`, and `sha256` should be overloaded
* Faker version: 25.0.1
* OS: all

The docstrings for these functions state:

> If raw_output is False (default), a hexadecimal string representation of the MD5 hash
> will be returned. If True, a bytes object representation will be returned instead.

But the type annotations do not reflect this:

```console
$ cat t.py
import faker

fake = faker.Faker()

x = fake.md5(raw_output=False)
reveal_type(x)

y = fake.md5(raw_output=True)
reveal_type(y)

$ mypy t.py
$ mypy t.py
t.py:6: note: Revealed type is "Union[builtins.bytes, builtins.str]"
t.py:9: note: Revealed type is "Union[builtins.bytes, builtins.str]"
Success: no issues found in 1 source file
```

The revealed types ought to be `builtins.str` and `builtins.bytes` respectively.

I believe this can be achieved by overloading the type annotations on those methods (using `typing.overload`.

I would be happy to open a PR to make those changes, if that would be welcome?

samueljsb:
Before this change, the return type for these functions was always `bytes| str`. However, the type can be narrowed based on the value passed to `raw_output`.

This change adds overload annotations for the two allowable values to allow type checkers to narrow the type and correctly infer the exact return type.

Resolves #2046

---

As opposed to the example in the issue, this now shows the correct types:

```console
$ cat t.py
import faker

fake = faker.Faker()

x = fake.md5(raw_output=False)
reveal_type(x)

y = fake.md5(raw_output=True)
reveal_type(y)

x = fake.sha1(raw_output=False)
reveal_type(x)

y = fake.sha1(raw_output=True)
reveal_type(y)

x = fake.sha256(raw_output=False)
reveal_type(x)

y = fake.sha256(raw_output=True)
reveal_type(y)

$ mypy t.py
t.py:6:13: note: Revealed type is "builtins.str"
t.py:9:13: note: Revealed type is "builtins.bytes"
t.py:12:13: note: Revealed type is "builtins.str"
t.py:15:13: note: Revealed type is "builtins.bytes"
t.py:18:13: note: Revealed type is "builtins.str"
t.py:21:13: note: Revealed type is "builtins.bytes"
Success: no issues found in 1 source file
```


fcurella:
Hi! Could you push something to your branch, so that workflow can run again?

fcurella:
the stub file (`faker/proxy.pyi`) is meant to be autogenerated by [`generate_stubs.py`](https://github.com/joke2k/faker/blob/master/generate_stubs.py), which currently does not recognize `@overload`ed signatures. Could you look into adding support for `@overload` to `generate_stubs.py`?

fcurella (approved):
Thank you!

PR #2048, opened 2024-05-20
Add addition geo for pl_PL

george0st:
Add addition geo locations for pl_PL


fcurella (approved):
Thank you

faker/providers/geo/pl_PL/__init__.py
```diff
@@ -26,4 +26,25 @@ class Provider(GeoProvider):
         ("50.29761", "18.67658", "Gliwice", "PL", "Europe/Warsaw"),
         ("50.32492", "18.78576", "Zabrze", "PL", "Europe/Warsaw"),
         ("50.34802", "18.93282", "Bytom", "PL", "Europe/Warsaw"),
+        ("52.24424", "21.08545", "Praga Południe", "PL", "Europe/Warsaw"),
```
mgorny, line 29: Is it intentional that the list includes both cities all across Poland and districts of Warsaw?

george0st: Hi @mgorny, the "Europe/Warsaw" is only about continent and capital city (I saw the same logic e.g. for Turkey). If you see exact definition, please let me know.

mgorny: The last field is a timezone. I'm referring to entries such as "Praga Południe" or "Śródmieście" that aren't cities but districts (i.e. should be "Warszawa Praga Południe" and "Warszawa Śródmieście" to be consistent.

That said, I don't see the main list in `geo/__init__.py` to list any districts, so I think only actual cities should be listed.

Also, I've just noticed that "Warsaw" uses the English name while all other cities use their Polish names.

george0st: I see, you are right, the "Praga Południe" and "Śródmieście" are part of Warsaw. I will repair it, I would like to focus on cities (not part of cities).

mgorny: Thanks. I can make a PR, if you prefer, as (obviously) I'm familiar with cities in PL :-).

george0st: Hi @mgorny , the repair see #2056

PR #2050, opened 2024-05-28
Add handshake emoji with different color variations to emoji provider…

tamkc:
### What does this change

This commit adds support for handshake emoji with different color variations to the emoji provider list. The new emoji includes support for various skin tones and genders.

The new emoji list now includes the following handshake emoji variations:

- 🤝🏻
- 🤝🏼
- 🤝🏽
- 🤝🏾
- 🤝🏿


fcurella (approved):
Thank you!

PR #2054, opened 2024-05-29
Added support for Nigerian Yoruba names and surnames (yo_NG)

5uru:
### What does this change

Add Added support for Nigerian Yoruba names and surnames.


fcurella (approved):
Thank you!

PR #2057, opened 2024-06-05
More fixes for the `geo/pl_PL` provider

mgorny:
### What does this change

Fix more errors in `geo/pl_PL` provider data.

### What was wrong

1. The timezone was specified as `Europe/Warszawa` which is not a standard iANA timezone name.
2. The capital city was using English name, inconsistently with other cities.
3. Some city names were missing diacritics.
4. A few districts of Warsaw remained on the list.

### How this fixes it

1. Restored `Europe/Warsaw` as the timezone name.
2. Renamed the capital to `Warszawa`.
3. Added missing diacritics.
4. Removed the remaining districts.

I've also reported the issues to `latitude.to`, but given it's autogenerated, I doubts whether they'll be able to fix the data. Overall, it looks like an unreliable source of information.


fcurella (approved):
Thank you

mgorny:
Thanks!

Issue #2041, opened 2024-05-05
pydecimal() - `min_value` and `max_value` should support `Decimal`, along with `float`


sshishov:
* Faker version: 25.0.1
* OS: MacOS 14.4.1 (23E224)

After migrating to version 25+, profiding default value as `Decimal` for `min` or `max` values causes violation

### Steps to reproduce

```python
amount=fake.pydecimal(min_value=Decimal(1), max_value=Decimal(100000)),
```

### Expected behavior

The should not be any error observed. The rationale - a lot of software using Decimal from the box for all calculations, in the constants etc to rely on precise double calculation. Therefore I think that it should support Decimal from the box, especially if it is returning Decimal itself.

### Actual behavior

Produces this mypy error:
```
 error: Argument "min_value" to "pydecimal" of "Faker" has incompatible type "Decimal"; expected "float | None"  [arg-type]
```

stefan6419846:
This is a typing issue, no "real" bug. Feel free to provide a PR to update the corresponding type hint and regenerate the typing stubs.

sshishov:
@stefan6419846 how I can report typing issues? Because we are havily using faker, and we are requiring `strict` typing, that's why I have noticed these errors...

If you will give me a way to report typing issues, then I will report there.

stefan6419846:
It is indeed correct to report this here, but your initial report sounded like an implementation bug and not a type hint issue.

parsariyahi:
If we want to fix that, we can add more annotation in code 
```python
    def pydecimal(
        self,
        left_digits: Optional[int] = None,
        right_digits: Optional[int] = None,
        positive: bool = False,
        min_value: Optional[Union[float, int, Decimal]] = None,
        max_value: Optional[Union[float, int, Decimal]] = None,
    ) -> Decimal:
```
but I think a problem is, we can have another types of numbers that are subclass of int or something like this,
how we can handle that in here?

sshishov:
Hi @parsariyahi , any subclass should be okay, I guess. We need to check though...

parsariyahi:
After some research, I found something, the built-in `numbers` module, this module has a class `Number` which is a abstract base class for all numbers I think.
you can read it here: https://docs.python.org/3/library/numbers.html
as the docs said:
[...]

PR #2059, opened 2024-06-10
Change `pydecimal` type hint for `min_value`, `max_value` to `numbers.Number`

parsariyahi:
### What does this change
in`faker\providers\python\__init__.py` method `pydecimal` I changed the type hints for `min_value`, `max_value` to `numbers.Number`

### What was wrong
#2041 issue in `mypy` for `type hints`

### How this fixes it
This PR will use `numbers.Number` instead of `float` for type hint

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -5,6 +5,7 @@ import warnings
 
 from decimal import Decimal
 from enum import Enum
+from numbers import Number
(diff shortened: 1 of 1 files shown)


fcurella (approved):
Thank you!

fcurella:
@sshishov I see [the types updated](https://github.com/joke2k/faker/blob/master/faker/proxy.pyi#L2453-L2454). What version of `Faker` are you running?

Issue #2061, opened 2024-06-17
passport_owner() generates different gender names in wrong order


Bikbai:
* Faker version: "25.8.0"
* OS: windows 11

Trying to generate female names in cyrilic. First component should be family, last name according to ICAO DOC 9303

### Steps to reproduce

`from faker import Faker
fake = Faker('ru-RU')
print(fake.passport_owner(gender='F'))`

### Expected behavior

('Белоусова', 'Лариса')

### Actual behavior

('Лариса', 'Белоусов')

stefan6419846:
The corresponding provider is only implemented for the US: https://github.com/joke2k/faker/tree/master/faker/providers/passport Feel free to submit a localized version for Russia.

PR #2066, opened 2024-06-26
Added passport provider for ru_RU language

Abdujabbar:
### What does this change

Added new passport provider for ru_RU language

### What was wrong

It looks like based on an issue https://github.com/joke2k/faker/issues/2061 for passport_owners we need to return first surname and then given_name

### How this fixes it

I created a new provider class in the path: faker/providers/passport/ru_RU/__init__.py and overloaded the function passport_owner

### DIFF
--- a/faker/providers/passport/ru_RU/__init__.py
+++ b/faker/providers/passport/ru_RU/__init__.py
@@ -1,0 +1,11 @@
+from typing import Tuple
+
+from faker.typing import SexLiteral
+
+from .. import Provider as PassportProvider
+
+
+class Provider(PassportProvider):
+    def passport_owner(self, gender: SexLiteral = "M") -> Tuple[str, str]:
+        given_name, surname = super().passport_owner(gender)
+        return surname, given_name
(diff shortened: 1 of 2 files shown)


faker/providers/passport/ru_RU/__init__.py
```diff
@@ -0,0 +1,9 @@
+from typing import Tuple
+
+from .. import Provider as PassportProvider
+
+
+class Provider(PassportProvider):
+    def passport_owner(self, gender: str = "M") -> Tuple[str, str]:
```
fcurella, line 7: you can use `faker.typing.SexLiteral` as type for the `gender` argument.

fcurella (changes requested):

faker/providers/passport/ru_RU/__init__.py
Abdujabbar: Agree, let I will fix it! 

Thanks for feedback.

fcurella (changes requested):
Can you add a test for the new provider?

Abdujabbar:
> Can you add a test for the new provider?

@fcurella  I have added a simple use case for testing, let's try to check. 

But for now it looks like some pipelines failed, I guess it's not related into my changes, am I right?

fcurella (approved):
Thank you!

Issue #2045, opened 2024-05-13
Provider `pydict` is flaky 


bverhoeve:
* Faker version: 25.0.1
* OS: macOS Sonoma

Brief summary of the issue goes here.
When using the `pydict` provider, this sometimes results in a `ValueError` in the underlying `random_elements` provider:
```
ValueError: Sample length cannot be longer than the number of unique elements to pick from.
```
This error does not pop up all the time but it seems to occur for some combination. Chances of it happening seem to increase with a higher value for the `nb_elements` parameter

### Steps to reproduce

1. Run `faker.pydict(nb_elements=1000)` a couple of times. After a while the error pops up

### Expected behavior
A random `dict` is generated

### Actual behavior
A `ValueError` is raised.

mileswatsonbjss:
We have a potential fix, will post the PR here when it is ready for review

bverhoeve:
Great news! What was the issue?

mileswatsonbjss:
Here is the PR, If you could both please review when you have a chance (https://github.com/joke2k/faker/pull/2067)

Details re. what caused the issue etc.. are all in the PR.

fcurella:
Released in v26.0.0

bverhoeve:
Awesome, thanks guys!

PR #2067, opened 2024-06-26
Fix upper limit of nb_elements

mileswatsonbjss:
### What does this change

Fixes the upper limit of nb_elements

### What was wrong

see https://github.com/joke2k/faker/issues/2045

### How this fixes it

When requesting a high number of nb_elements when calling `pydict`, unless you specified `variable_nb_elements` as `False` it randomises the number so that it is then between 60% and 140% of the original requested `nb_elements`.  
This would be fine in cases where `nb_elements` is a number far below the length of the `lorem` (locale) list but if 140% of the number (`nb_elements`) is greater than the `lorem` (locale) list length you may see the error in the issue linked above.

It was also possible see this error when `variable_nb_elements` is `False`, if the number of `nb_elements` is greater than the length of the lorem (locale) list length.

The fix we have added compares the length of the `nb_elements` to the length of the lorem (locale) list after the call to `randomize_nb_elements` in `pydict` to cater for both scenarios above.  If `nb_elements` is greater than the length of the `lorem` (locale) list we set `nb_elements` to the length of the `lorem` (locale) list, also making the user aware of the fact that this has happened through a warning, else we use the number provided by the user (`nb_elements`) or by the randomisation function in cases where `variable_nb_elements` is `True`.

### DIFF
--- a/faker/providers/lorem/__init__.py
+++ b/faker/providers/lorem/__init__.py
@@ -23,23 +23,16 @@ class Provider(BaseProvider):
     word_connector = " "
     sentence_punctuation = "."
 
-    def words(
+    def get_words_list(
         self,
-        nb: int = 3,
         part_of_speech: Optional[str] = None,
         ext_word_list: Optional[Sequence[str]] = None,
-        unique: bool = False,
     ) -> List[str]:
-        """Generate a tuple of words.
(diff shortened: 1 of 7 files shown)


fcurella (approved):
Thank you!

PR #2069, opened 2024-07-07
Extend ro_RO company localization with prefixes

DDSNA:
### What does this change

Added prefixed to company ro_RO

### What was wrong

No prefixes function

### How this fixes it

Added the most common prefixes for companies


fcurella (approved):
Thank you!

PR #2071, opened 2024-07-09
Declare support for Python 3.13

edgarrmondragon:
### What does this change

The first release candidate of Python 3.13 is around the corner ([2024-07-30](https://peps.python.org/pep-0719/)), this adds it to the test matrix and declares support in the project metadata.

### What was wrong

Python 3.13 seems to be supported without any code changes, but this makes sure it's tested.

### How this fixes it

* Adds PyPI trove classifier
* Adds `3.13` to the list of versions for `actions/setup-python`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,7 +160,6 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python }}
-        allow-prereleases: true
     - name: Install dependencies
       run: |
         python -m pip install mypy
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -245,7 +245,6 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python }}
-          allow-prereleases: true
       - name: Install Tox and any other packages
         run: python -m pip install tox coveralls
       - name: Run Tox
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -197,7 +197,6 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python }}
-          allow-prereleases: true
       - name: Install Tox and any other packages
         run: python -m pip install tox coveralls
       - name: pytest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,7 +138,13 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python:
+        - "3.8"
+        - "3.9"
+        - "3.10"
+        - "3.11"
+        - "3.12"
+        - "3.13"
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


edgarrmondragon:
Tests are 🟢 over on my fork: https://github.com/edgarrmondragon/faker/actions/runs/9849708765

.github/workflows/ci.yml
```diff
@@ -154,6 +160,7 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python }}
+        allow-prereleases: true
```
stefan6419846, line 163: Why not using `"3.13-dev"` instead to better indicate that Python 3.13 is not yet available as a stable release?

tox.ini
```diff
@@ -1,11 +1,11 @@
 [tox]
-envlist=py{38,39,310,311,312,py3},alpine,flake8,checkmanifest,isort,mypy,doc8
+envlist=py{38,39,310,311,312,313,py3},alpine,flake8,checkmanifest,isort,mypy,doc8
 skip_missing_interpreters = true
 
 [testenv]
 deps =
     coverage>=5.2
-    freezegun<0.4
+    freezegun<2
```
stefan6419846, line 8: How is this related to the Python 3.13 change?

edgarrmondragon: Ah, forgot to comment. freezegun 1.5.0 was made compatible with Python 3.13 via https://github.com/spulec/freezegun/commit/d75458f3f691d0c2c35559872716d0ebab738817.

.github/workflows/ci.yml
edgarrmondragon: Feels easier to let it stay as `3.13` after the final release rather than coming back and changing it from `3.13-dev` to `3.13`. I don't have a strong preference, though.

tox.ini
edgarrmondragon: Thoughts @stefan6419846?

stefan6419846: I am usually against needless lower pinning.

stefan6419846: Is there a specific reason for the upper limit? We tend to not have it for the other dependencies as well.

edgarrmondragon: Not really. I'd actually prefer `freezegun>=1.5.0` since that's really the hard constraint to make it work on 3.13.

edgarrmondragon: I've removed the upper constraint: 6c915fe0fc1505b65a247bb4c106cf9dbaf332f5

edgarrmondragon:
Rebased and confirmed with `tox -e py313` that tests pass.

.github/workflows/ci.yml
edgarrmondragon, line 163:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -184,6 +197,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python }}
+          allow-prereleases: true
```
edgarrmondragon, line 200:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -226,6 +245,7 @@ jobs:
...
```
edgarrmondragon, line 248:
```suggestion
```

edgarrmondragon:
ping @fcurella 🙏

fcurella:
Thank you!

PR #2080, opened 2024-08-01
Fix `pydecimal` handling of `positive` keyword

### LINKED ISSUE #2051: `pydecimal` does not properly handle `positive` keyword
The same issue is happening for `pydecimal` as it was reported for `pyfloat`: https://github.com/joke2k/faker/issues/1928

I think that it should work more straightforward and more reliable, like: we should have a `min` and `max` values being updated and only at the end we generate the value.

MIN:
  - `min` unset, `positive` unset, result = minimum calculated from places
  - `min` set, `positive` unset, result = maximum of `min` and minimum calculated from places
  - `min` unset, `positive` set to True, result = maximum between 0 and minimum calculated from places
  - `min` unset, `positive` set to False, result = minimum calculated from places
  - `min` set and `positive` set to True, result = maximum between 0 and `min`
  - `min` set and `positive` set to False, result = maximum of `min` and minimum calculated from places

MAX:
  - `max` unset, `positive` unset, result = maximum calculated from places
  - `max` set, `positive` unset, result = minimum of `max` and maximum calculated from places
  - `max` unset, `positive` set to True, result = maximum calculated from places
  - `max` unset, `positive` set to False, result = minimum between 0 and maximum calculated from places
  - `max` set and `positive` set to True, result = minimum of `max` and maximum calculated from places
  - `max` set and `positive` set to False, result = minimum between 0 and `max`
  
Afterwards we should check `min` <= `max` and then return the random value from this range

I guess this approach is safe and reliable and will indicate about issues in usage when you try to build pydecimal from constrains which produce invalid range

tahzeer:
### What does this change

- `pydecimal` now handles `positive` keyword as expected 
- `random.Random` now takes in expected type `int` for parameter

### What was wrong

- Same as #1954 
- `random.Random` no longer accepts any hashable object as param [(link)](https://docs.python.org/3/library/random.html#alternative-generator)

### How this fixes it

- Now it is checked weather `positive` is None by default (in which case sign assigning is random) or it has been explicitly passed a s `True`/`Flase` (handled in `else` block)
- Explicit typecasting from `Hashable` to `int`

Fixes #2051

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -320,7 +320,7 @@ class Provider(BaseProvider):
         elif max_value is not None and max_value <= 0:
             sign = "-"
         else:
-            if positive == None:
+            if positive is None:
                 sign = self.random_element(("+", "-"))
             else:
                 sign = "+" if positive else "-"
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -285,7 +285,7 @@ class Provider(BaseProvider):
         self,
         left_digits: Optional[int] = None,
         right_digits: Optional[int] = None,
-        positive: bool = False,
+        positive: Optional[bool] = None,
         min_value: Optional[BasicNumber] = None,
         max_value: Optional[BasicNumber] = None,
     ) -> Decimal:
@@ -320,7 +320,10 @@ class Provider(BaseProvider):
         elif max_value is not None and max_value <= 0:
             sign = "-"
         else:
-            sign = "+" if positive else self.random_element(("+", "-"))
+            if positive == None:
+                sign = self.random_element(("+", "-"))
+            else:
+                sign = "+" if positive else "-"
 
         if sign == "+":
             if max_value is not None:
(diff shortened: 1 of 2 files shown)


faker/providers/python/__init__.py
```diff
@@ -320,7 +320,10 @@ def pydecimal(
         elif max_value is not None and max_value <= 0:
             sign = "-"
         else:
-            sign = "+" if positive else self.random_element(("+", "-"))
+            if positive == None:
```
fcurella, line 323: Just a stylistical note: could we use `is` to compare to `None`?

```suggestion
            if positive is None:
```

fcurella (changes requested):

faker/providers/python/__init__.py
tahzeer: added the changes in [bf1055a](https://github.com/joke2k/faker/pull/2080/commits/bf1055ace76df99e7a403275cf997c165ed325a8)

PR #2081, opened 2024-08-07
Add en_MS Malaysian provider

carlosfunk:
### What does this change

Add the en_MS provider just for addresses.  This is pretty much a straight port from [PHP Faker](https://github.com/fzaninotto/Faker/blob/master/src/Faker/Provider/ms_MY/Address.php) with some improved state abbreviations.


fcurella (approved):
Thank you!

PR #2083, opened 2024-08-08
Pakistani names Class

c2-tlhah:
### What does this change
I have added names of pakistani people
Brief summary of the changes.
Created a new file having class for pakistani names
### What was wrong
NA
NA
### How this fixes it
NA
NA
Fixes #...
NA


fcurella (changes requested):
Can you add a test for the new provider?

c2-tlhah:
> Can you add a test for the new provider?
< Done

fcurella (approved):

PR #2086, opened 2024-08-22
Add a few names to the `en_in` Person provider

Vikrambgs:
**added some Indian male names**


fcurella (changes requested):
Can you please add the source for these names?

fcurella:
I'm sorry but we can't accept names that are unsourced. This is both to avoid sampling bias and to reduce list updates.

PR #2088, opened 2024-08-22
Add es_ES isbn provider

mondeja:
### What does this change

Add es_ES isbn provider and refactor the base provider to simplify localized provider creation.

### What was wrong

The base provider was not inheritable and the code assumptions there were confusing.


fcurella (approved):
Thank you!

PR #2094, opened 2024-08-28
Add it_IT lorem provider

gianni-di-noia:
**What does this change**
Extend the Italian language coverage

**What was wrong**
while setting the "it_IT" language, it was returning Latin instead of Italian sentences 

**How this fixes it**
Add it_IT LoremProvider

fixes: https://github.com/joke2k/faker/issues/2095


fcurella (approved):
Thank you!

PR #2114, opened 2024-10-08
Uk-ua updates

lexxai:
### What does this change

Renamed city and street names when translating the Ukrainian parts of providers.
In accordance with changes in national legislation, and text optimization and replacement of duplicates.

But the text is not sorted.

- https://zakon.rada.gov.ua/laws/show/z1534-23
- https://zakon.rada.gov.ua/laws/show/3984-IX
- https://omr.gov.ua/ua/acts/council/203797
- https://maps.visicom.ua
- https://uk.wikipedia.org


fcurella (approved):
Thank you!

PR #2120, opened 2024-10-15
Fix Greek SSN format pattern.

### LINKED ISSUE #2119: Wrong pattern for Greek police IDs
https://github.com/joke2k/faker/blob/8cee450e10463c0360e50e2b56da4ecfea1dd513/faker/providers/ssn/el_GR/__init__.py#L36

* Faker version: v30.3.0
* OS: Linux

### Expected behavior

Proper format / pattern should be "???####".

### Actual behavior

Current format / pattern is "??######".

### LINKED ISSUE #2121: Wrong pattern for Greek SSNs
https://github.com/joke2k/faker/blob/8cee450e10463c0360e50e2b56da4ecfea1dd513/faker/providers/ssn/el_GR/__init__.py#L70

* Faker version: v30.3.0
* OS: Linux

### Expected behavior

Proper format / pattern should be "%d%m%y#####".

### Actual behavior

Current format / pattern is "%d%m%y####".

nepiskopos:
Fixes #2121.

### DIFF
--- a/faker/providers/ssn/el_GR/__init__.py
+++ b/faker/providers/ssn/el_GR/__init__.py
@@ -60,7 +60,7 @@ class Provider(BaseProvider):
         return vat_id + str(tin_checksum(vat_id))
 
     # Uses Luhn checksum according to this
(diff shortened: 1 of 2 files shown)


Abdujabbar:
I think that would be great if we could add some tests to validate changes.

fcurella (changes requested):
Could add a check for the `ssn` length to the [existing tests](https://github.com/joke2k/faker/blob/master/tests/providers/test_ssn.py#L235-L240)?

nepiskopos:
I apologize, but the issue I was trying to fix is invalid. I misunderstood the way Luhn algorithm creates the check digit. The current behavior is correct.

Issue #2125, opened 2024-10-23
faker.providers: is this really needful to keep in this package


RanaUniverse:
* Faker version: faker             30.8.0
* OS: Ubuntu 24.04

Brief summary of the issue goes here.

![image](https://github.com/user-attachments/assets/5124dfca-450e-4cb3-9104-6a82ce032935)
I was trying to learn how to use faker library in python, so first went to the pypi page and trying checking copy paste the code example given there, so when i come to this example and then i copy paste this in my vs code and i find no problem, but see the below image, when i exactly copy paste the code from the website into my main.py then when i run it shows a random ip address, which i understand as the docs says it will generate the fake ip address.

![image](https://github.com/user-attachments/assets/c31cfecc-c91f-4a8a-b64e-ef5ab814e24e)

And after i see the code i didn't understand why the **from faker.providers import internet**

and i didnot understand what is the use case of this line, and why i need to add it liek thsi, **fake.add_provider(internet)**

so i just comment this two line form the example code and then copy paste this in a new main2.py file and when i run , i found the same type result, i am also getting the fake ip address from this library as seen in the below image:

![image](https://github.com/user-attachments/assets/7ee554a4-c5ec-4d95-82a0-85ccacba8358)

😢😢😢😢😢😢😢😢😢😢😢😢
😢😢😢😢😢😢😢😢😢😢😢😢
😢😢😢😢😢😢😢😢😢😢😢😢

Please friends say me why i need to use this 

*** 
from faker.providers import internet
fake.add_provider(internet)
***

Please say me when i will use this and when i will not use this, i am just confused in the beginning of this library use.
Please say if you can explain in little words,

stefan6419846:
I do not understand the title, but as you saw, the README seems to be a bit outdated and the providers usually are registered automatically.

Feel free to submit a corresponding PR.

stefan6419846:
You should not have to use the `add_provider` calls unless you are using external providers - the bundled providers are already available out of the box.

The rendered representation on PyPI usually is the README of the package and not the actual documentation, although it might include some basic information. You should usually have a look at the linked documentation at https://faker.readthedocs.io/en/stable/index.html (although this uses the same README/text on the index page).

I have been talking about submitting a pull request here on GitHub. Nevertheless, the README seems just a big misleading here and I am not sure whether it is worth it to make this more explicit.

RanaUniverse:
[...]

PR #2127, opened 2024-10-26
Updated README docs on providers part

laziest-coder:
### What does this change
Fixes https://github.com/joke2k/faker/issues/2125

### What was wrong

New programmers were confused by outdated docs on Providers.

### How this fixes it

* Delete call to `add_provider` from example code
* Added note section stating that common providers are already available

### Checklist

- [x] I have run `make lint`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -104,12 +104,13 @@ packaged in "providers".
 .. code:: python
 
     from faker import Faker
-    from faker.providers import internet
 
     fake = Faker()
-    fake.add_provider(internet)
 
     print(fake.ipv4_private())
+    print(fake.address())
+
+**Note:** The `add_provider`_ method is not necessary unless you are using external providers - the bundled providers are already available out of the box.
 
 
 Check the `extended docs`_ for a list of `bundled providers`_ and a list of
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -104,12 +104,14 @@ packaged in "providers".
...
-    from faker.providers import internet
 
     fake = Faker()
-    fake.add_provider(internet)
 
     print(fake.ipv4_private())
+    print(fake.address())
+
+.. note::
+    The `add_provider` method is not necessary unless you are using `external providers`_ - the bundled providers are already available out of the box.
```
stefan6419846, line 114: This render incorrectly at GitHub and referencing `add_provider` without it being shown anywhere does not make much sense IMHO.

laziest-coder: @stefan6419846 yes you are right, I've updated the rendering part as well as added reference to `add_provider` method

README.rst
```diff
@@ -104,12 +104,13 @@ packaged in "providers".
...
     from faker import Faker
-    from faker.providers import internet
 
     fake = Faker()
-    fake.add_provider(internet)
 
     print(fake.ipv4_private())
+    print(fake.address())
+
+**Note:** The `add_provider`_ method is not necessary unless you are using external providers - the bundled providers are already available out of the box.
```
fcurella, line 113: To me it makes it a little puzzling to mention the `The add_provider method` in this paragraph, when it's not in the code example above.

fcurella (changes requested):
We already have a quickstart section that does not mention `add_provider` specifically to keep things simple for new users.

While I agree that the wording in this section could be improved, I dont think we should remove the mention of `add_provider`

laziest-coder:
Agree, the changes don't seem to change much, closing the PR.

PR #2130, opened 2024-10-29
fix typing in date_time provider for maybe_format_as_string functions

### LINKED ISSUE #2116: Optional format as string of dates broke existing typed uses
* Faker version: 30.3.0
* OS: Linux

We have some uses of `date_between` and similar that expect the target variable to be of type `datetime.date`.
The new decorator breaks the type annotations because the output of these functions is now assumed to be `Union[datetime.date, str]`.

### Steps to reproduce

The following code passes type checking in 30.2.0:
```py
foo: datetime.date = Faker().date_between()
```

It does not anymore in 30.3.0:
```
example.py:1: error: Incompatible types in assignment (expression has type "date | str", variable has type "date")  [assignment]
```

### Expected behavior

As the `pattern` argument is `None`, the return type should not have changed.

### Actual behavior

The `pattern` argument is not taken into account to type the return value.

### Possible fix

Using [`typing.overload`](https://docs.python.org/3/library/typing.html#typing.overload) on the affected functions would allow for the return type to be dependent on the parameter types.

This would unfortunately also add quite a lot of repetition to the function definitions, as they need to be repeated multiple times.

nils-van-zuijlen:
### What does this change

Add overload typing definitions to functions using the `maybe_format_as_string` decorator.

### What was wrong

The functions return values typing was not marked as depending on the arguments.

### How this fixes it

Fixes #2116 

### Checklist

- [x] I have run `make lint`

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -9,7 +9,7 @@ from datetime import datetime
 from datetime import time as dttime
 from datetime import timedelta
 from datetime import tzinfo as TzInfo
-from typing import Any, Callable, Dict, Iterator, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Iterator, Optional, Tuple, Union, overload
 
 from dateutil import relativedelta
 from dateutil.tz import gettz, tzlocal, tzutc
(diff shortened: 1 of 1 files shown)


nils-van-zuijlen:
As t6he original issue was fixed, this does not have a reason to exist anymore.

PR #2134, opened 2024-11-04
Remove Collection[str] type from ElementTypes

tvuotila:
### What does this change

This removes `Collection[str]` type from `ElementsType` union type.

### What was wrong

`ElementsType` union type already has `Collection[T]` which can resolve to `Collection[str]` when type is `ElementsType[str]`. New mypy versions could not narrow `ElementsType` type when it was used as argument type.

### How this fixes it

Remove `Collection[str]` type from `ElementsType` union type and expect everyone to use `ElementsType[str]` when they want `Collection[str]`.

### Checklist

- [x] I have run `make lint` (I did not include changes to `faker/proxy.pyi` or changes to unrelated files.)

### DIFF
--- a/faker/providers/__init__.py
+++ b/faker/providers/__init__.py
@@ -17,7 +17,7 @@ _re_qm = re.compile(r"\?")
 _re_cir = re.compile(r"\^")
 
 T = TypeVar("T")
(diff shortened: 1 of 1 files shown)


fcurella (approved):
Thank you!

PR #2133, opened 2024-11-02
Fix korean road address

semi-yu:
### What does this change

Adds building-number(건물 번호) to a road address in 'ko_KR' locale. Without building-numbers, faker-generated addresses do not reflect real addressing system in Republic of Korea(South korea). 

**before:**
address form: `"{{road_name}}{{road_suffix}}"`
대전광역시 강서구 석촌호수거리

**after:** 
*below shows one of possible outputs*

- **대전광역시 강서구 석촌호수거리 110** (Basic building number format)
- **대전광역시 강서구 석촌호수거리 지하110** (The building's main entrance is underground)
- **대전광역시 강서구 석촌호수거리 110-31** (The building is cramped together, distinguished by sub-building number; exact format is (*main-building-number*)-(*sub-building-number*))

(FYI, 석촌호수거리 is a road name.)
 
Change above also applies to the other form of an address, which is defined as : `"{{road_name}}{{road_number}}{{road_suffix}}"`.  

### What was wrong

The original source code did not include any logic to generate building numbers. Although it is explained on comments written on source code.

### How this fixes it

Added additional strings to describe building numbers. 

Fixes #2132 

### Checklist

- [x] I have read the documentation about [CONTRIBUTING](https://github.com/joke2k/faker/blob/master/CONTRIBUTING.rst)
- [x] I have read the documentation about [Coding style](https://github.com/joke2k/faker/blob/master/docs/coding_style.rst)
- [x] I have run `make lint`

PR #2138, opened 2024-11-14
Resolve address number output issue and refactor road_formats for clarity

semi-yu:
# What does this change
This change addresses the issue in the previous implementation (#2133), where `#` was mistakenly printed instead of the actual address numbers. Additionally, the `road_formats` structure has been simplified by incorporating `{{building_number}}` directly for more concise formatting.

# What was wrong
The original code had a formatting error where `#` was printed instead of the intended numeric address numbers. This caused unexpected outputs in the feature responsible for generating ROK (🇰🇷) road addresses (도로명 주소), leading to inaccuracies.

# How this fixes it
The building_number method now correctly handles the generation of building numbers. It produces a standard building number or, with a certain probability, generates variations such as underground entrances using `building_number_underground()` or `includes sub-building numbers via building_number_segregated()`. This probability was determined arbitrarily, as there are no official government statistics available for this data.

### Checklist

- [x] I have run `make lint`

### DIFF
--- a/faker/providers/address/ko_KR/__init__.py
+++ b/faker/providers/address/ko_KR/__init__.py
@@ -527,6 +527,44 @@ class Provider(AddressProvider):
         """
         return self.random_element(self.road_suffixes)
 
+    def building_number(self) -> str:
+        """
+        :returns: A random building number
+
+        Generates building number(건물 번호). There are 3 types of building number with current ROK addressing system.
+            (1) 19: A typical format. Only marks one building.
+            (2) 지하11: The building entrance is underground.
+            (3) 132-1: Several buildings are distinguished with sub-building-number(가지 번호).
+
+        Generating probability is arbitrarily.
+
+        :example: 19, 지하11, 143-1
+        """
+        if self.random_int() % 9 < 1:
+            return self.building_number_underground()
+        elif self.random_int() % 9 < 4:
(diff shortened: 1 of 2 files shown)


stefan6419846:
As this is a bug of the previous variant, shouldn't there be a test which checks for the previously broken behavior as well?

semi-yu:
Hello,
You mentioned that it was a bug. However, there are a few points to consider:

1. It's not so much a bug as it was my oversight. I forgot to add the `bothify()` function. 
2. This implementation fixes the part that was done clumsily. Instead of relying on adding strings to formats, I changed it so that the function generates the string.
3. For string returns, I ensured that only `str` types are returned and modified the tests to check only for that.

Please review these changes. Since I'm using ChatGPT for communication, there may be some nuances lost in translation. Thank you.

fcurella:
If addresses are supposed to never contain `#`, you could write a test like `assert "#" not in building_number`

semi-yu:
Good morning from Korea.

Thank you for clarifying the task.
I have added two test cases for the newly implemented building number functions.

- Any string generated by `building_number_underground()` must begin with '지하' (which means "underground").
- Any string generated by `building_number_segregated()` must include a dash ('-').

Please let me know if anything else is needed!

fcurella (approved):
Thank you!

PR #2141, opened 2024-11-19
feat: add Ukrainian currency provider

SaulTigh:
### What does this change

Added Ukrainian currency provider.

### What was wrong

--

### How this fixes it

--

### Checklist

- [x] I have run `make lint`


SaulTigh:
@fcurella FYI: i've run make lint, but generate_stubs didn't work as expected on my py3.12 virtualenv, so excluded result of that script from changeset

fcurella (approved):
Thank you!

Issue #2152, opened 2025-01-22
Faker 34.0 is breaking change


sww314:
Our build/tests broke when upgrading from 33.1 to 34. There was nothing in the release note that indicated the change in behavior was intentional.

This is field that causes the test failure:

`manufactured_on = factory.Faker("date_between", start_date="-30y", end_date="today")`

Logs from out test:
```
Step #8 - "test": 
Step #8 - "test": cls = <class 'faker.providers.date_time.en_US.Provider'>, value = 'today'
Step #8 - "test": 
Step #8 - "test":     @classmethod
Step #8 - "test":     def _parse_date_string(cls, value: str) -> Dict[str, float]:
Step #8 - "test":         parts = cls.regex.match(value)
Step #8 - "test":         if not parts:
Step #8 - "test":             raise ParseError(f"Can't parse date string `{value}`")
Step #8 - "test":         parts = parts.groupdict()
Step #8 - "test":         time_params: Dict[str, float] = {}
Step #8 - "test":         for name_, param_ in parts.items():
Step #8 - "test":             if param_:
Step #8 - "test":                 time_params[name_] = int(param_)
Step #8 - "test":     
Step #8 - "test":         if "years" in time_params:
Step #8 - "test":             if "days" not in time_params:
Step #8 - "test":                 time_params["days"] = 0
Step #8 - "test":             time_params["days"] += 365.24 * time_params.pop("years")
Step #8 - "test":         if "months" in time_params:
Step #8 - "test":             if "days" not in time_params:
Step #8 - "test":                 time_params["days"] = 0
Step #8 - "test":             time_params["days"] += 30.42 * time_params.pop("months")
Step #8 - "test":     
Step #8 - "test":         if not time_params:
Step #8 - "test": >           raise ParseError(f"Can't parse date string `{value}`")
Step #8 - "test": E           faker.providers.date_time.ParseError: Can't parse date string `today`
Step #8 - "test": 
Step #8 - "test": /usr/local/lib/python3.12/site-packages/faker/providers/date_time/__init__.py:1924: ParseError
```

clarin-ebtio800090:
Hello, we tried to solve the issue.

This is what we did:

Update the _parse_date_string method in the date_time provider to properly handle the 'today' string. This will fix the ParseError that occurs when using 'today' as an end_date in the date_between method.

[...]

PR #2153, opened 2025-01-22
fix “today” parsing

fcurella:


sww314:
Thanks for the quick fix. 
#2152

PR #2158, opened 2025-01-27
Add `ru_RU` SNILS generator

### LINKED ISSUE #2150: Add `ru_RU` SNILS generator
`Faker version: 33.3.1`

Using faker for personal data generation, I discovered that there is no method to generate [russian individual insurance account number (SNILS)](https://en.wikipedia.org/wiki/SNILS_(Russia)). Thus I would like to request and develop this feature. The algorithm is pretty simple: 9 random digits + 2 digits representing a checksum.

**The checksum algorithm:**
```python
def calculate_snils_checksum(value: str) -> str:
    base_snils_length = 9
    check_sum = sum(int(value[i]) * (base_snils_length - i) for i in range(base_snils_length))

    if check_sum > 101:
        check_sum = check_sum % 101

    if check_sum < 100:
        check_sum = str(check_sum).zfill(2)
    else:
        check_sum = "00"

    return check_sum
```
, assuming you pass 9-digits-string as a `value` parameter.

Pandede:
### What does this change
Resolves #2150.
Add `company.snils` function of locale `ru_RU`, and here is [the implementation](https://mimesis.name/master/_modules/mimesis/builtins/ru.html#RussiaSpecProvider.snils) in `mimesis`.
### Checklist
- [x] I have run `make lint`

### DIFF
--- a/faker/providers/company/ru_RU/__init__.py
+++ b/faker/providers/company/ru_RU/__init__.py
@@ -12,6 +12,11 @@ def calculate_checksum(value: str) -> str:
     return str((check_sum % 11) % 10)
 
 
+def calculate_snils_checksum(numbers: str) -> str:
+    checksum = sum(int(v) * (9 - i) for i, v in enumerate(numbers)) % 101
+    return "%02d" % (checksum if checksum < 100 else 0)
+
+
 class Provider(CompanyProvider):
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

Issue #2042, opened 2024-05-05
uuid4() - become unusuable for safe typing without casting or ignoring


sshishov:
* Faker version: 25.0.1
* OS: MacOS 14.4.1 (23E224)

When using `fake.uuid4()`, the returning type is `bytes | str | UUID` which is impossible to safely use later on in the code.
I guess one way to solve it to use `literals` instead of `cast_to` callable, as there are limited ways to cast it to. Only 3 types, therefore better to use `Literal` as `cast_to` and provide `@typing.overload` to properly infer return type from the input param. Also by default it should return `str`

### Steps to reproduce

Nypy will produce error as for type checker the value can be either bytes or str or UUID which cannot be joined together. But in runtime here will no be any errors

```python
value1 = fake.uuid4()
value2 = fake.uuid4()

joined = ','.join([value1, value2])  # <-- errors will be here
```

### Expected behavior

No typing errors should be observed

### Actual behavior

Using `fake.uuid4()` without `typing.cast` or `type: ignore` appears impossible after adding the types.

stefan6419846:
Feel free to provide a corresponding PR which improves this.

sshishov:
@stefan6419846 it will require breaking changes, is it okay? I definitely can provide PR for this one

fcurella:
Can't this be fixed by using some kind of `TypeVar` in the type hint of `uuid4`?

steve-mavens:
Based on the type in the docs:

```
def uuid4(cast_to: ~typing.Callable[[~uuid.UUID], str] | ~typing.Callable[[~uuid.UUID], bytes] | None = <class 'str'>) → bytes | str | UUID
```

Like fcurella says, I think it should be possible to overload it like:

```
T = TypeVar('T', str, bytes)

@overload
def uuid4(cast_to: Callable[[UUID], T]) -> T: ...
@overload
def uuid4(cast_to: None) -> UUID: ...
```

Failing that, get rid of T by explicitly list the possibilities:

```
@overload
def uuid4(cast_to: Callable[[UUID], str]) -> str: ...
@overload
def uuid4(cast_to: Callable[[UUID], bytes]) -> bytes: ...
@overload
def uuid4(cast_to: None) -> UUID: ...
```

I don't think that should be a breaking change. 

Or potentially you could make the function more general, allowing `cast_to` to convert to any type at all instead of just `str` or `bytes`:

```
T = TypeVar('T')

@overload
[...]

PR #2164, opened 2025-02-04
Fix typing for uuid4()

evenicoulddoit:
### What does this change

* Adds proper type support for `uuid4()`, so that static type checkers understand the return type based on the inputs

### What was wrong

* Working with Faker in an environment using a static type checker such as MyPy / Pyright is frustrating when using `uuid4()`, because these tools rightly point out that the return type is one of `UUID | str | bytes`

### How this fixes it

* This adds overload typing so that the output is deterministic

### Checklist

- [X] I have read the documentation about [CONTRIBUTING](https://github.com/joke2k/faker/blob/master/CONTRIBUTING.rst)
- [X] I have read the documentation about [Coding style](https://github.com/joke2k/faker/blob/master/docs/coding_style.rst)
- [X] I have run `make lint`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,13 +8,16 @@ mypy:
 	mypy --install-types --non-interactive --config mypy.ini faker
 
 black:
-	black --line-length 120 .
+	black .
 
 isort:
 	isort --atomic .
 
-generate-stubs: 
-	python3.10 generate_stubs.py
+generate-stubs:
+	python3.11 generate_stubs.py
+	black faker/proxy.pyi
+	isort faker/proxy.pyi
+
 
 lint: generate-stubs isort black mypy flake8
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,15 +159,15 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: ${{ matrix.python }}
+        python-version: '3.12'
     - name: Install dependencies
       run: |
-        python -m pip install mypy
-    - name: Static type checking with mypy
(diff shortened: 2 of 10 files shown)


fcurella (changes requested):
Thank you!

We must also update the [`generate_stubs.py`](https://github.com/joke2k/faker/blob/master/generate_stubs.py) script to support the `@overload` decorator.

See also https://github.com/joke2k/faker/issues/2046, https://github.com/joke2k/faker/issues/2042

evenicoulddoit:
@fcurella done. Also updated the dependency management to make it easier to consistently develop

Makefile
```diff
@@ -8,13 +8,16 @@ mypy:
 	mypy --install-types --non-interactive --config mypy.ini faker
 
 black:
-	black --line-length 120 .
```
evenicoulddoit, line 11: This is now defined in `pyproject.toml` and needn't be tracked elsewhere.

Makefile
```diff
@@ -8,13 +8,16 @@ mypy:
...
 
 isort:
 	isort --atomic .
 
-generate-stubs: 
-	python3.10 generate_stubs.py
+generate-stubs:
+	python3.11 generate_stubs.py
+	black faker/proxy.pyi
+	isort faker/proxy.pyi
```
evenicoulddoit, line 19: Unless these are applied you don't actually see your changes properly.

evenicoulddoit:
@fcurella the CI process uses the master branch to determine the checks. I can't fix the typing until #2171 is merged in.

fcurella:
Done in 9c8b132424b57be6e4b3fb8ea3d129495c0fd4cd

PR #2167, opened 2025-02-05
fix: remove invalid word from ja_JP lorem provider

kymckay:
### What does this change

I'm removing the `〜` character from the word list of the `ja_JP` lorem provider.

### What was wrong

As far as I can tell (I'm not a Japanese speaker and may be wrong) this is not a word. We hit this unexpectedly in one of our unit tests.

### Checklist

- [ ] I have run `make lint`


fcurella (approved):
Thank you!


From a little googling, my understanding is that `〜` is punctuation, and therefore not a word.

PR #2172, opened 2025-02-07
Drop support for Python 3.8

knyghty:
### What does this change

Drops support for Python 3.8

### What was wrong

EOL Python version still supported.

### How this fixes it

Dropped various references to Python 3.8.

I explicitly didn't check if there's any code that can be cleaned up. Maybe good to run pyupgrade on it and see what happens?

### Checklist

- [x] I have run `make lint`


knyghty:
I think I found a bit more. In fact I think it's possible to get rid of typing-extensions now as well. I've done this here, but it could be moved to another PR if you'd prefer.

fcurella (changes requested):
Could you also remove 3.8 from tox.ini and from the testing matrix?

fcurella (approved):
Never mind my previous comment!

Thank you! 🌈

PR #2185, opened 2025-03-07
Add hash functions overloads with no params

### LINKED ISSUE #2184: Hash function's new typing, such as faker.sha1, is not backward compatible
Hi, thank you very much for this library

The version [36.2.1](https://github.com/joke2k/faker/commit/2f20f76db8921cc97b257d04eca75916485028a0) added new overloads so that sha1 is correctly typed related to the raw_output... but the function itself has a default so calling it with no params should be supported.

```
value = faker.sha1()
````
produces
```
error: All overload variants of "sha1" of "Faker" require at least one argument  [call-overload]
note: Possible overload variants:
note:     def sha1(self, raw_output: Literal[True]) -> bytes
note:     def sha1(self, raw_output: Literal[False]) -> str
```

Versions / config
Python: 3.12.8
faker: 36.2.2
mypy: 1.15.0
mypy config:
```
[tool.mypy]
namespace_packages=true
explicit_package_bases=true
strict=true
show_error_codes=true
```

lindycoder:
### What does this change

Added an extra over load to all sha functions that takes no param and return a string.

### What was wrong

The original functions all have a default `False`, which means it should be callable without a parameter, returning a string.

### How this fixes it

By providing the missing overload

Fixes #2184

### Checklist

- [ ] I have run `make lint` 

^ running make lint is changing things unrelated to this change

### DIFF
--- a/faker/providers/misc/__init__.py
+++ b/faker/providers/misc/__init__.py
@@ -61,6 +61,9 @@ class Provider(BaseProvider):
         return os.urandom(length)
 
     @overload
+    def md5(self) -> str: ...
+
+    @overload
     def md5(self, raw_output: Literal[True]) -> bytes: ...
 
     @overload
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

PR #2186, opened 2025-03-10
Fix timezone import name clash

### LINKED ISSUE #2183: TypeError: timezone() missing required argument 'offset' when using date_time provider
Hej all,
With faker v36.2.2 on Python 3.12.0 I have an issue that might originate from the recent datetime refactoring. 

The following code works
```python
from faker import Faker, providers
from datetime import timezone

fk = Faker()
Faker.seed(4321)
# fk.add_provider(providers.date_time)
# print(timezone is fk.timezone)
print(fk.timezone())
```

But when umcommenting the line with add_provider, the error indicated in the title happens. 

As a workaround I can drop the date_time provider but it still might be confusing to users.

As a hint I found that `timezone is fk.timezone` weirdly is True when the provider is added, maybe something mixed up in the imports.

pylipp:
### What does this change

Fixes #2183

### What was wrong

I couldn't use `add_provider()` with the `date_time` provider and the `timezone()` method.

I found out that the built-in timezone module is identical to the object returned from the `timezone()` method.

git-bisect showed me that 604e1ed6 is the culprit commit.

### How this fixes it

Use a renaming of the built-in timezone module at import time.

Fixes #2183

### Checklist

- [x] I have run `make lint`

### DIFF
--- a/faker/providers/date_time/__init__.py
+++ b/faker/providers/date_time/__init__.py
@@ -7,7 +7,8 @@ from datetime import MAXYEAR
 from datetime import date as dtdate
 from datetime import datetime
 from datetime import time as dttime
-from datetime import timedelta, timezone
+from datetime import timedelta
+from datetime import timezone as dttimezone
 from datetime import tzinfo as TzInfo
 from typing import Any, Callable, Dict, Iterator, Optional, Tuple, Union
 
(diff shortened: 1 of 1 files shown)


fcurella (approved):
Thank you!

PR #2200, opened 2025-04-08
Fix remove duplicated cities in  `DE_de` address provider

KarelZe:
### What does this change

This PR removes duplicates for German cities.

### What was wrong

Cities `Eichstätt`, `Soltau` and `Stade` where duplicated, so that they were sampled more frequently than any other city (see reproducer). The official German city register lists only one exact match for these cities. See here: https://www.statistikportal.de/de/gemeindeverzeichnis

Reproducer:

```python
from collections import Counter

from faker import Faker

fake = Faker("de_DE")

city = [fake.city() for _ in range(100_000)]
counts = Counter(city)

for city, freq in counts.most_common(15):
    print(f"{city}: {freq}")
```
Output:
```
Stade: 539
Soltau: 514
Eichstätt: 501
Ribnitz-Damgarten: 300
Dachau: 299
Lübben: 297
...
```

### How this fixes it

Duplicates are removed. Uniform sampling is ensured.

### Checklist

- [x] I have run `make lint`


fcurella (approved):
Thank you!

PR #2210, opened 2025-05-04
Update README.rst

zachpinto:
Added Web App link

### What does this change

Added Streamlit Web App link to user GUI version of Faker

### What was wrong

Nothing was wrong. But a GUI is fun and cool

### How this fixes it

GUI Web App

- [x] I have run `make lint`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,6 +88,13 @@ to ``faker.Generator.format(method_name)``.
     # 'Wellington Koelpin II'
     # 'Ms. Karley Kiehn V'
 
+Web App
+-----------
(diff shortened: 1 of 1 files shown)


fcurella:
I’m worried this could give users the false impression that we support a
GUI for Faker.

zachpinto:
Do you support a GUI? lol

 I guess I could put it into an examples directory.

fcurella:
We could make it clear by having a ‘related projects’ section with the
disclaimer that those projects are not directly supported by us

PR #2214, opened 2025-05-15
Changed the function for generating vin, now the last 4 characters are digits

nesb1:
### What does this change

VIN must ends with 4 digits, https://www.studocu.com/es-ar/document/universidad-nacional-de-misiones/legislacion-y-ejercicio-profesional/iso-3779-2009-vin-decoder/70130592 in 4.4

```
The VIS shall be the third and final section of the VIN and shall consist of eight characters, the last four being numerical. 
```

### What was wrong

Before this change VIN can ends with ascii letters.

### How this fixes it

change args in `bothify` function call

### Checklist

- [x] I have run `make lint`


fcurella (approved):
Thank you!

PR #2235, opened 2025-07-05
Fix ISBN13 generation for pt_BR locale (closes #2217)

### LINKED ISSUE #2217: isbn13() ignores locale and generates non-local ISBNs even with pt_BR
When using Faker('pt_BR').isbn13(), the generated ISBNs always contain group codes such as 978-0 or 978-1, which belong to English-speaking countries (e.g., US/UK). They do not reflect Brazilian ISBN group codes such as 978-85 or 979-85, despite the locale being set to pt_BR. This behavior contradicts expectations, as the Faker documentation shows that locale='pt_BR' is supported across many providers, including: address(), bairro(), postcode(), estado(), current_country(). These providers generate culturally appropriate data for Brazil. Therefore, users reasonably expect that ISBNs generated with pt_BR would also reflect Brazilian standards.

```
from faker import Faker

fake = Faker('pt_BR')
print(fake.isbn13())
# Output: '978-1-79230-225-1'
# Group "1" = English-speaking region (US/UK)
```

### Expected Behavior
ISBNs generated with Faker('pt_BR') should contain Brazilian group codes, such as: `978-65-89999-01-3`
Alternatively, this limitation should be clearly documented to avoid misleading users.

### Actual Behavior
ISBNs generated ignore the locale and always use group codes from other regions (e.g., 0, 1, 978-1), regardless of the configured locale.

angelicaccampos:
### What does this change

Adds a localized `isbn13()` implementation for the `pt_BR` locale, generating ISBNs using valid Brazilian group codes (`978-85`, `978-65`, and `979-85`). It includes a new provider at `faker.providers.isbn.pt_BR`.

### What was wrong

When using `Faker('pt_BR').isbn13()`, the returned ISBNs always include group codes like `978-0` or `978-1`, which correspond to English-speaking countries. This behavior disregards the configured locale and is inconsistent with expectations for Brazilian data.

### How this fixes it

This change introduces a custom provider for `pt_BR` that:
- Selects a Brazilian EAN and group code.
- Randomly generates publisher and title codes.
- Computes a valid ISBN-13 check digit.
- Returns the final ISBN using correct formatting.

This ensures that `Faker('pt_BR').isbn13()` produces culturally appropriate and structurally valid ISBNs for Brazil.

Fixes #2217

### DIFF
--- a/faker/providers/isbn/pt_BR/__init__.py
+++ b/faker/providers/isbn/pt_BR/__init__.py
@@ -1,0 +1,73 @@
+# faker/providers/isbn/pt_BR/__init__.py
+from faker.providers.isbn import Provider as ISBNProvider
+import random
+
+class Provider(ISBNProvider):
+    """ISBN provider para locale pt_BR com grupos brasileiros oficiais."""
+    
+    BRAZILIAN_ISBN_GROUPS = [
+        {
+            'ean': '978', 'group': '65', 'name': 'Brasil (new allocation)',
+            'publisher_ranges': [(10, 99), (100, 999), (1000, 9999)]
+        },
+        {
+            'ean': '978', 'group': '85', 'name': 'Brasil (traditional)',
+            'publisher_ranges': [(10, 99), (100, 999), (1000, 9999)]
+        },
+        {
+            'ean': '979', 'group': '85', 'name': 'Brasil (extended)',
+            'publisher_ranges': [(10, 99), (100, 999), (1000, 9999)]
+        }
+    ]
+    
+    def isbn13(self, separator="-"):
+        """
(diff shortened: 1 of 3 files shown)


faker/providers/isbn/pt_BR/__init__.py
```diff
@@ -0,0 +1,73 @@
+# faker/providers/isbn/pt_BR/__init__.py
+from faker.providers.isbn import Provider as ISBNProvider
+import random
+
+class Provider(ISBNProvider):
+    """ISBN provider para locale pt_BR com grupos brasileiros oficiais."""
```
fcurella, line 6: Can you use English for docstrings and comments?

fcurella (changes requested):

PR #2238, opened 2025-07-06
Plate.test

Fernandavazgit1:
### What does this change

This PR adds a new method called `latam_plate()` to the `automotive` provider for generating vehicle license plates from different Latin American countries.

Supported formats:
- 🇧🇷 Brazil: `ABC-1234`
- 🇦🇷 Argentina: `AB 123 CD`
- 🇨🇱 Chile: `AB-CD-12`

The method allows the developer to specify a country using the `pais` parameter (default: random choice). This is useful for testing systems involving vehicle registration, logistics, transportation apps, etc.

### What was wrong

Currently, Faker does not support license plate formats specific to Latin American countries. Developers working on apps for this region need custom generators to simulate realistic vehicle data for Brazil, Argentina, Chile, and others.

### How this fixes it

This adds a custom method `latam_plate()` which generates localized plate formats depending on the selected country. It uses internal logic for formatting and random generation based on common standards in those regions.

Fixes #12345  

### Checklist

- [x] I have run `make lint` and `make test`
- [x] I have added unit tests in `test_plate.py`
- [x] I have documented the new method

### DIFF
--- a/MétodoSerTestado.py
+++ b/MétodoSerTestado.py
@@ -1,0 +1,10 @@
+from faker.providers import BaseProvider
+import random
+import string
+
+class Provider(BaseProvider):
+    def passport_number(self):
+    
+        letters = ''.join(random.choices(string.ascii_uppercase, k=2))
+        numbers = ''.join(random.choices(string.digits, k=7))
+        return letters + numbers
--- a/codigotest.py
+++ b/codigotest.py
@@ -1,0 +1,38 @@
+import unittest
+from faker import Faker
+import re
+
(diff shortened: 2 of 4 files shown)


MétodoSerTestado.py
```diff
@@ -0,0 +1,10 @@
+from faker.providers import BaseProvider
+import random
+import string
+
+class Provider(BaseProvider):
+    def passport_number(self):
```
fcurella, line 6: did you push the right code? Looks like this is generating a passport number, not a Latin American license plate

PR #2241, opened 2025-07-12
feat: add StrongPasswordProvider for secure password generation

Prg-maker:
### What does this change

Brief summary of the changes.

### What was wrong

### How this fixes it

Fixes #...
- However, it does not guarantee that there is at least one of the specified characters.

### Checklist

- [ ] I have read the documentation about [CONTRIBUTING](https://github.com/joke2k/faker/blob/master/CONTRIBUTING.rst)
- [ ] I have read the documentation about [Coding style](https://github.com/joke2k/faker/blob/master/docs/coding_style.rst)
- [ ] I have run `make lint`

### DIFF
--- a/faker/providers/password/__init__.py
+++ b/faker/providers/password/__init__.py
@@ -1,0 +1,14 @@
+import string
+import random
+import string
+import numbers
+from .. import BaseProvider
+
+class Provider(BaseProvider):
+    """Password generation provider"""
+    password = []
+    def generate_password(self, size: int) -> str:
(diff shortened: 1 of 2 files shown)


fcurella (changes requested):
We already have a `password` method in the [`misc` provider](https://github.com/joke2k/faker/blob/master/faker/providers/misc/__init__.py#L164). I'd rather improve that then add a new one.

Would you also please fill the PR template?

PR #2243, opened 2025-07-17
fix(pydecimal): allow Decimal type for min_value and max_value

### LINKED ISSUE #2242: Add Decimal arg type to the pydecimal() provider
Hi, I just noticed that the `pydecimal()` provider https://github.com/joke2k/faker/blob/fff4c0d858724799720b8bce96e5e7381325f3be/faker/providers/python/__init__.py#L284-L291 does not support [Decimal](https://docs.python.org/3/library/decimal.html#decimal-objects) objects for its `min_value` and `max_value` arguments. I think it should, considering this provider _produces_ Decimals.

There’s a subtle issue here regarding these lower/upper bound values, because
```python
>>> decimal.Decimal('0.1') == float('0.1')
False
```
See also [Floating-Point Arithmetic: Issues and Limitations](https://docs.python.org/3/tutorial/floatingpoint.html).

Happy to provide a PR if you want.

sshishov:
### What does this change

Allow `Decimal` type to be provided for `min_value` and `max_value` in type-safe manner.

### What was wrong

The type checkers `mypy` and `pyright` were complaining that the type `Decimal` is not supported by the function type declaration.

### How this fixes it

Add `Decimal` type to the allowed union

Fixes #2242, #2041

### Checklist

- [+] I have read the documentation about [CONTRIBUTING](https://github.com/joke2k/faker/blob/master/CONTRIBUTING.rst)
- [+] I have read the documentation about [Coding style](https://github.com/joke2k/faker/blob/master/docs/coding_style.rst)
- [+] I have run `make lint`

### DIFF
--- a/faker/proxy.pyi
+++ b/faker/proxy.pyi
@@ -2666,8 +2666,8 @@ class Faker:
         left_digits: Optional[int] = ...,
         right_digits: Optional[int] = ...,
         positive: Optional[bool] = ...,
-        min_value: Union[float, int, None] = ...,
-        max_value: Union[float, int, None] = ...,
+        min_value: Union[float, int, Decimal, None] = ...,
+        max_value: Union[float, int, Decimal, None] = ...,
     ) -> Decimal: ...
     def pydict(
         self,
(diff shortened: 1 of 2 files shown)


sshishov:
I have tested with `numbers.Number` and it does not work as you have suggested earlier @fcurella 

Regarding other comments, like `__abs__` it should work, if it is not workin, then it is typing issues from other places, because providing `Decimal` value to the function does work without any issues and exceptions.

PR #2245, opened 2025-07-27
Recreate the documentation

writedev:
# What does this change

This completely changes the documentation.

### What was wrong

The main problem with Faker's documentation was its readability and the fact that not all functions were listed.

### How this fixes it
I decided to use mkdocs material because it makes documentation easier to create and more attractive.

Fixes #...

### Checklist

- [x] list all fonctions
- [ ] Create a space to learn how to create Providers
- [ ] Trad page


stefan6419846:
Shouldn't the focus be on adding the missing bits to the actual code instead of having a separate file to maintain for each method?

writedev:
> Shouldn't the focus be on adding the missing bits to the actual code instead of having a separate file to maintain for each method?

I created a file for each function because it seemed more logical that way.

And what do you mean by adding missing elements from the actual code?

Because normally there are only three descriptions missing.

fcurella:
Having to maintain a file for each method is a non-starter.
We currently autogenerate `.rst` files based on code. Could you provide a way to do that for the `md` files in your PR?

writedev:
@fcurella

To do this, I listed all of Faker's functions, their results, and their descriptions, then I asked the ChatGPT API to give me a reformulated description for each function.

But you say that it will be complicated to maintain a file for each function, but the functions that are already present will not change or will change very little, so will it remain the same concept? And for a new function, will it be enough to follow the same “template”?

Issue #2253, opened 2025-08-23
First 8 digits of result of `ssn()` in 'nl_NL' provider are (needlessly) unique


Dutcho:
# First 8 digits of result of `ssn()` in 'nl_NL' provider are (needlessly) unique
## Issue
The `ssn()` code generates the first 8 digits by calling `random.sample(range(10), k=8)`, so only the 9th digit can duplicate one of the first 8 digits.
The first 8 digits are all different.
```
>>> assert all(len(set(fake.ssn()[:-1])) == 8 for _ in range(100_000))
```
## Fix
The code *should* call `random.choices(range(10), k=8)` instead.
## Requirement
Uniqueness of the first 8 digits is **not** a requirement for Dutch BSN's (see example at [Wikipedia](https://nl.wikipedia.org/wiki/Burgerservicenummer#11-proef)).
It reduces the range from BSN's >80 million to only 1.6 million, which is less than the population of the Netherlands. That's how I found the issue, trying to generate a test file of 2 million unique BSN's.
## Secondary issue
Although incorrect, the current version avoids `ssn()` results with 2 (or more) leading zeroes.
That's a happy accident as BSN's must be 8 (+ leading zero) or 9 digits long.

Therefore, when fixing `ssn()`, the code should also filter results with leading '00'. That can be accomplished
- either by brute force (i.e. filtering out generated `[0, 0, ...]` `digits` lists)
- or by choosing digit 2 from `range(1 if digit1 == 0 else 0, 10)` (instead of `range(10)`).

PR #2254, opened 2025-08-23
Nl nl ssn issue #2253

Dutcho:
### What does this change

Fix [issue #2253](https://github.com/joke2k/faker/issues/2253) on `ssn()` in 'nl_NL' provider.

### What was wrong

Generated `ssn()` values had unique leading 8 digits, which is *not* a BSN (Dutch ssn) requirement.
As a consequence the range (number of possible values) was limited to 10! / 2! * 9 / 10.

### How this fixes it

Replace `random.sample()` by `random.choices()` and `continue` in case `digits[:2] == [0, 0]` in `ssn()` code.
Now the range is 10 ** 8 * 99 / 100 * 9 / 10, and I generated 20 million unique BSN's without issue (other than 2.6 million collisions).

### Checklist

- [X] I have read the documentation about [CONTRIBUTING](https://github.com/joke2k/faker/blob/master/CONTRIBUTING.rst)
- [X] I have read the documentation about [Coding style](https://github.com/joke2k/faker/blob/master/docs/coding_style.rst)
- [ ] I have run `make lint`
(I didn't install locally)

### DIFF
--- a/faker/providers/ssn/nl_NL/__init__.py
+++ b/faker/providers/ssn/nl_NL/__init__.py
@@ -21,10 +21,10 @@ class Provider(SsnProvider):
 
         while True:
             # create an array of first 8 elements initialized randomly
-            digits = self.generator.random.choices(range(10), 8)
-            # discard if not 8 or 9 digits long
-            if digits[0] == 0 and digits[1] == 0:
-                continue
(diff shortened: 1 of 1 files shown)


fcurella (changes requested):
can you add a unit test for regression purposes?

PR #2268, opened 2025-09-18
Add Scientific Provider

AlTisl:
### What does this change

This pull request introduces a new community provider, `ScientificProvider`, with methods which allows users to generate fake author identifiers used in scientific databases and publication tracking systems like Scopus, ORCID etc.

### What was wrong

The Faker library previously did not have a provider with such functionality.

### Checklist


fcurella (approved):
Thank you!

PR #2299, opened 2025-12-27
feat: Add selective uniqueness with exclude_types()

### LINKED ISSUE #2204: Question: uniqueness except bools
Sometimes, Faker generates duplicate values.

Faker supports a `unique` mode, which guarantees unique values: https://faker.readthedocs.io/en/master/#unique-values

I could implement this by changing `get_faker`s return value to `fake.unique`, but that would also apply to `bool`s, which causes unique values (true/false) to be exhausted **real** quick. And I don't care about duplicate booleans.

**Does Faker support excluding specific types from uniqueness (`bool`s in this case)?**

To decrease the chance of uniques, I now add `unique` to all Faker invocations (on instances returned by `get_faker`). However, this is not foolproof (one could forget to add `unique`), and inevitably still causes duplicates incidentally, albeit less often.

rodrigobnogueira:
# Add selective uniqueness with `exclude_types()`

Implements feature requested in #2204: Allow excluding specific types (like `bool`) from uniqueness checks to prevent exhaustion.

## Problem

Boolean values only have 2 possible values (True/False), so `fake.unique.pybool()` exhausts after 2 calls and raises `UniquenessException`. Users want to use `unique` globally but exclude types with small value sets.

## Solution

Added `exclude_types()` method to `UniqueProxy` that returns a new proxy excluding specified types from uniqueness enforcement.

### Usage

```python
from faker import Faker

fake = Faker()

# Exclude bools from uniqueness checks
proxy = fake.unique.exclude_types([bool])

# Bools can now return duplicates without exhaustion
for _ in range(100):
    proxy.pybool()  # Works fine! No UniquenessException

# Other types still enforce uniqueness
names = [proxy.first_name() for _ in range(5)]
# All names are unique
```

### Multiple type exclusions

```python
# Exclude multiple types at once
proxy = fake.unique.exclude_types([bool, int])
```

## Implementation Details

- Added `_excluded_types` attribute to `UniqueProxy.__init__`
- New `exclude_types([type, ...])` method returns new proxy with exclusions
- Modified `_wrap()` to skip uniqueness check for excluded types
- Excluded proxies share `_seen` dictionary for consistency
- Works with locale selection: `fake.unique.exclude_types([bool])['en_US']`

## Testing

**New tests**: 8 comprehensive tests in `test_unique_exclude.py`
- Excludes specified types from uniqueness
- Still enforces uniqueness for non-excluded types  
- Prevents exhaustion for types with limited values
- Shares seen dictionary across proxies
- Works with locale selection

**Full test suite**: ✅ 2209 passed, 13 skipped

## Backward Compatibility

✅ **Fully backward compatible** - no breaking changes
- Existing `fake.unique` behavior unchanged
- New `exclude_types()` is opt-in

Fixes #2204

### Checklist

- [x] I have run `make lint`

### DIFF
--- a/docs/fakerclass.rst
+++ b/docs/fakerclass.rst
@@ -354,6 +354,49 @@ be raised.
         fake.unique.boolean()  # UniquenessException!
 
 
+For types with limited value sets (like booleans with only 2 values), you can use
+``exclude_types()`` to prevent uniqueness checks for specific types:
+
+
+.. code:: python
+
+   import faker
+
+   fake = faker.Faker()
+
+   # Exclude booleans from uniqueness checks
+   proxy = fake.unique.exclude_types([bool])
+
+   # This works fine - booleans can now repeat
+   for i in range(100):
+        proxy.pybool()  # No UniquenessException!
+
+   # Other types still enforce uniqueness
+   names = [proxy.first_name() for i in range(10)]
+   assert len(set(names)) == 10  # All unique
+
+
+The ``exclude_types()`` method returns a new proxy that shares the same seen values
+dictionary, ensuring consistency across different proxy instances:
+
+
+.. code:: python
+
+   from faker import Faker
+
+   fake = Faker()
+
+   # Get a unique name
+   name1 = fake.unique.first_name()
+
+   # Create proxy excluding bools
+   proxy = fake.unique.exclude_types([bool])
+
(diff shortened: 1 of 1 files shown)


fcurella (changes requested):
Can we add some documentation about the new feature?

PR #2312, opened 2026-01-30
Improve unique value support

Sawy7:
Hello, I would like to submit these changes aiming to improve unique value generation that was introduced [here](https://github.com/joke2k/faker/pull/1226). The main focus of this PR is to allow accessing every property through `Faker().unique` and to introduce a mode in which unique values are preferred, but not enforced (`Faker().preferred_unique`). Full set of changes bellow.

### What does this change

- Allow accessing any property through `UniqueProxy`
- Add preferred uniqueness (won't raise if unique values exhausted)
- Exempt generator methods starting with 'current_' from uniqueness checks and make their values initially unique with multiple locales
- Modify and add new tests

### What was wrong

These are quality-of-life improvements, rather than bugfixes.

### How this fixes it

Slight additions to `Faker` class. Most modifications are in `UniqueProxy`.

### Checklist

- [x] I have run `make lint`

Thanks for your consideration. Please let me know if there is anything I can change to make this work for inclusion in the project.

### DIFF
--- a/faker/proxy.py
+++ b/faker/proxy.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import copy
 import functools
+import logging
 import re
 
 from collections import OrderedDict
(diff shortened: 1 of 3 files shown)


Sawy7:
Hey @fcurella, sorry to bother you. Just wanted to ask if there is something I could do in order to make the inclusion of my changes possible. Thanks!

PR #2314, opened 2026-02-05
Add country names to he_IL locale.

Moo64c:
### What does this change

Added countries to the `he_IL` locale.

### What was wrong

Using the `en` locale values.

### How this fixes it

Overrides the default with the correct values.

### Checklist

- [x] I have run `make lint` 
  - (second commit fixed whitespaces)


Moo64c:
@fcurella Test added.

PR #2334, opened 2026-03-01
Add food provider

stuckvgn:
### What does this change

Adds a `food` provider with five new faker methods: `dish()`, `ingredient()`, `spice()`, `fruit()`, and `vegetable()`. Data sourced from Wikipedia's culinary lists.

### What was wrong

There is no built-in food provider. Users who need food-related test data must hard-code strings or maintain their own provider.

### How this fixes it

Adds `faker/providers/food/__init__.py` with five methods and matching tests in `tests/providers/test_food.py`.

_Claude Code (claude-sonnet-4-6) was used to generate initial code. Output was reviewed and verified._

### Checklist

- [x] I have run `make lint`

### DIFF
--- a/faker/providers/food/__init__.py
+++ b/faker/providers/food/__init__.py
@@ -1,0 +1,269 @@
+from .. import BaseProvider, ElementsType
+
+
+class Provider(BaseProvider):
+    """Implement default food provider for Faker.
+
+    Sources:
+    - Dish names: https://en.wikipedia.org/wiki/List_of_dishes (checked 2025-01-01)
+    - Ingredients: https://en.wikipedia.org/wiki/List_of_culinary_fruits,
+      https://en.wikipedia.org/wiki/List_of_vegetables (checked 2025-01-01)
+    - Spices: https://en.wikipedia.org/wiki/List_of_culinary_herbs_and_spices (checked 2025-01-01)
+    """
+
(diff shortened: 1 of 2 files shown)


stefan6419846:
My first impression is that this should rather be moved to an external community provider instead of adding it to the core package.

samtuckerdavis:
Thanks for the feedback, @stefan6419846\! I'd argue food belongs in core rather than a community provider for a few reasons: faker already includes similarly broad-domain providers like `color`, `music`, and `currency` that aren't tied to a specific locale or use-case, and a `food` provider fits naturally alongside those as a universal faker domain. This PR covers culturally diverse dishes, ingredients, fruits, vegetables, and spices from around the world — not a single cuisine — so it's genuinely general-purpose. There's also a friction argument: for something as commonly needed as food data in test fixtures and seeding scripts, requiring users to find, install, and configure a separate community package raises the barrier unnecessarily for what should be a first-class concern. The addition is scoped proportionally (similar line count to the `color` or `music` providers) and fills a gap faker currently has entirely. Happy to discuss further or adjust scope if the maintainers have a preferred approach\!

stuckvgn:
Closing this PR — on reflection, this change may not be the right fit for this project. Thanks for your time!

PR #2340, opened 2026-03-19
Fix deepcopy proxy bindings for unique and optional helpers

SunS1eep1ng:
## Summary

`copy.deepcopy(Faker(...))` currently leaves proxy helpers in a broken state:

- `optional` is not rebuilt for the copied instance, so `deepcopy(Faker()).optional.name()` raises an `AttributeError`
- `unique` is rebuilt against the original `Faker` instance instead of the copied one
- `_factories` and `_factory_map` can also drift apart inside the copied object because they are deep-copied independently

This patch rebuilds the copied proxy helpers against the copied `Faker` instance and reconstructs `_factories` from the copied `_factory_map`.

## Reproduction

Before this patch:

```python
import copy
from faker import Faker

copy.deepcopy(Faker()).optional.name()
# AttributeError: 'Generator' object has no attribute 'optional'
```

## AI Using

I used an AI coding assistant to help inspect the bug, draft the fix, and draft the PR text. I verified the bug locally, implemented the change myself in this branch, and validated the relevant tests locally.

### DIFF
--- a/faker/proxy.py
+++ b/faker/proxy.py
@@ -141,12 +141,14 @@ class Faker:
     def __deepcopy__(self, memodict):
         cls = self.__class__
         result = cls.__new__(cls)
-        result._locales = copy.deepcopy(self._locales)
-        result._factories = copy.deepcopy(self._factories)
-        result._factory_map = copy.deepcopy(self._factory_map)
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

PR #2358, opened 2026-04-05
Follow-up to #2287: Add requested tests

patakk:
Builds on #2287 by @just6660.
Adds the tests to compare the outputs for two runs using the same seed, for more than one locale defined.
Also, importing random is no longer needed so it was removed.


fcurella (approved):
Thank you!

PR #2373, opened 2026-05-08
French social security numbers - corsican codes

### LINKED ISSUE #2372: Missing edge case for French social security numbers
The [current implementation in Faker](https://github.com/joke2k/faker/blob/e7d5cf2ee9c90058b56d393e40a91dc58506e4f7/faker/providers/ssn/fr_FR/__init__.py) only includes numeric department codes. 

[French social security numbers can include "2A" or "2B" for Corsica](https://fr.wikipedia.org/wiki/Num%C3%A9ro_de_s%C3%A9curit%C3%A9_sociale_en_France#Signification_des_chiffres_du_NIR). Correcting this would help people find real errors in their system but may result in incorrect bug reports - it is a common misconception that French social security numbers only contain digits.

Note that the checksum implementation must therefore be updated to use the standard substitutions "2A" -> "19" and "2B" -> "18"

There are also some minor errors in the text used in the implementation but these are not significant for the output.

dancergraham:
### What does this change

Adds corsican department codes to the list of possible values for the creation of French social security numbers

### What was wrong

The current implementation ommits the possibility of corsican department codes - the only ones which include letters (A or B). This may (correctly) break the code of some users of faker if they currently wrongly assume that French Social Security numbers are all digits. 

### How this fixes it

Adds the missing codes and updates the checksum calculation. 

Fixes #2372 

I used GPT5.4 via GitHub Copilot, giving it specific and detailed instructions on what changes to make - I have fully reviewed, verified and simplified the output.

### Checklist

- [x] I have run separately opened a PR to pin mypy and fix the linting and then run `make lint`


dancergraham:
Hi @fcurella anything I can do to make this PR easier to check / approve / reject ? This fixes a genuine hole in the current implementation.

fcurella (approved):
Thank you!

PR #2375, opened 2026-05-08
Restrict mypy version to less than 2

dancergraham:
### What does this change

Restrict mypy to less than version 2 until a full pass has been carried out on the types in this library for compatibility with version 2

### What was wrong

Linting is failing on new PRs due to incompatibility with the new version of mypy

### How this fixes it

Pins mypy to less than version 2 - this is often cited as good practice in such situations

Fixes #...

- [x] **No AI tools were used** in preparing this PR.
- [ ] **If AI tools were used**, I have disclosed which ones, and fully reviewed and verified their output.

### Checklist

- [x] I have run `make lint` mypy passes with less than version 2


fcurella (commented):
Thank you!

PR #2376, opened 2026-05-09
Fixed flaky test_optional by using seed and larger sample size.

### LINKED ISSUE #2366: Flaky test: tests/test_optional.py::test_optional relies on probabilistic assertion
- Faker version: 40.15.0
- OS: Windows 11 and WSL (Ubuntu).
- Python: 3.11 

## Description:

The test tests/test_optional.py:10 is flaky because it expects all values of a probabilistic generator to appear in only 10 random draws.

Location: tests/test_optional.py:10

## Current behavior:

The assertion requires:

- {fake.optional.boolean() for _ in range(10)} == {True, False, None}

This can fail randomly even when the implementation is correct.

## Expected behavior:

The test should be deterministic and should not fail intermittently due to random sampling variance.

## Why this is a problem:

- Nondeterministic test failures create CI noise.
- Can block merges without any real regression.
- Reduces confidence in test suite signal quality.

## Minimal Reproducible Example (MRE):

```python
from faker import Faker

fake = Faker()

# Same test logic
seen = {fake.optional.boolean() for _ in range(10)}
print(seen)
```

This produces sets that are missing one value.

## Reproduction:

1) Direct run under strict warning mode surfaced a failure where True was missing:

```bash
pytest -q tests/test_optional.py -W error::pytest.PytestWarning
```

Observed failure case:

- assert {False, None} == {False, True, None}

2) Monte Carlo Repeatability check:

- Runs: 5000
- Failing runs: 615
- Failure rate: 0.123

#### The issue reproduction screnshot:

Image

tejas-ae:
Fixes #2366.

`test_optional` drew only 10 samples from a probabilistic generator and asserted that all three outcomes (`True`, `False`, `None`) appeared. With the default probability split, the empirical failure rate was ~12% per run (confirmed by 5000-run Monte Carlo in the issue report).

**Fix:**
- Seed the generator with `Faker.seed(0)` for determinism.
- Increase the sample size from 10 to 100. At worst-case probability splits the chance of missing any outcome in 100 draws is astronomically small (< 0.000003%).

No behaviour change — only the test is affected.

### DIFF
--- a/tests/test_optional.py
+++ b/tests/test_optional.py
@@ -6,8 +6,13 @@ from faker import Faker
 class TestOptionalClass:
     def test_optional(self) -> None:
         fake = Faker()
+        Faker.seed(0)
 
-        assert {fake.optional.boolean() for _ in range(10)} == {True, False, None}
+        # 100 draws make it astronomically unlikely to miss any of the three
+        # outcomes (True, False, None) even under worst-case probability splits.
+        # 10 draws had a ~12% empirical failure rate (see GH-2366).
+        values = {fake.optional.boolean() for _ in range(100)}
+        assert values == {True, False, None}
 
     def test_optional_probability(self) -> None:
         """The probability is configurable."""
--- a/tests/test_optional.py
+++ b/tests/test_optional.py
@@ -6,11 +6,7 @@ from faker import Faker
 class TestOptionalClass:
     def test_optional(self) -> None:
         fake = Faker()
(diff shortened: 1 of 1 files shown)


dancergraham:
linting is now fixed - if you make a merge commit or similar minor update then the ci should run normally

tejas-ae:
Done! thanks.

fcurella (approved):
Thank you!

PR #2378, opened 2026-05-17
fix: raise NotImplementedError in bank() for locales without banks data

### LINKED ISSUE #2377: hasattr(faker, 'bank') returns True for en_US locale
Problem occurs from version 37.12.0

```python
def test_faker():
    locale = 'en_US'
    faker = Faker(locale=locale)
    faker.seed_instance(123)

    if hasattr(faker, 'company'): # True
        value = faker.company() # Ok
        print(value) # 'Adams Group'

    if hasattr(faker, 'bank'): # True
        value = faker.bank() # AttributeError: The Provider provider does not have a 'banks' attribute. Consider contributing to the project and  adding a 'banks' tuple to enable bank name generation.
        print(value)
```
full error message
```
self = <faker.providers.bank.en_GB.Provider object at 0x7f7955a88190>

    def bank(self) -> str:
        """Generate a bank name."""
        if not hasattr(self, "banks"):
>           raise AttributeError(
                f"The {self.__class__.__name__} provider does not have a 'banks' "
                "attribute. Consider contributing to the project and "
                " adding a 'banks' tuple to enable bank name generation."
            )
E           AttributeError: The Provider provider does not have a 'banks' attribute. Consider contributing to the project and  adding a 'banks' tuple to enable bank name generation.

.venv/lib/python3.14/site-packages/faker/providers/bank/__init__.py:54: AttributeError
```

koushik717:
fixes #2377

bank() was raising AttributeError when the locale has no banks data. the problem is hasattr silently catches AttributeError so hasattr(faker, 'bank') returns True but then faker.bank() crashes which is pretty confusing.

changed it to NotImplementedError like fcurella suggested. also updated the existing test to match.

### DIFF
--- a/faker/providers/bank/__init__.py
+++ b/faker/providers/bank/__init__.py
@@ -51,7 +51,7 @@ class Provider(BaseProvider):
     def bank(self) -> str:
         """Generate a bank name."""
         if not hasattr(self, "banks"):
-            raise AttributeError(
+            raise NotImplementedError(
                 f"The {self.__class__.__name__} provider does not have a 'banks' "
                 "attribute. Consider contributing to the project and "
                 " adding a 'banks' tuple to enable bank name generation."
--- a/tests/providers/test_bank.py
+++ b/tests/providers/test_bank.py
@@ -64,13 +64,21 @@ class TestBaseBankProvider:
     """Test base bank provider"""
 
     def test_bank_not_implemented_error(self, faker):
-        """Test that bank() raises AttributeError when no banks attribute exists"""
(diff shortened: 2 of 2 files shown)


koushik717:
Hi @joke2k just checking in on this one. All 28 checks are passing and there are no conflicts. Happy to make any changes if needed!

koushik717:
Hey @fcurella  you suggested raising NotImplementedError in #2377, this PR implements exactly that. Would love your feedback when you get a chance!

fcurella (approved):
Thank you!

PR #2383, opened 2026-06-01
Add US County Provider

ylub:
### What does this change

Adds U.S. county support to the `en_US` address provider:

- `county()`
- `county_by_state()`
- `county_with_state()`

Includes counties and county equivalents such as Louisiana parishes and Alaska boroughs.

### What was wrong

Faker’s `en_US` address provider currently has states, cities, ZIP codes, and street addresses, but no way to generate county names.

### How this fixes it

Adds county data and provider methods for generating counties, optionally by state.

Data source: [U.S. Census Bureau ANSI/FIPS county reference data](https://www.census.gov/library/reference/code-lists/ansi.html)

I used ChatGPT for help with Git/GitHub workflow, pull request setup, and repository navigation. I reviewed and verified the code myself.

### Checklist

- [x] I have read the documentation about CONTRIBUTING
- [x] I have read the documentation about Coding style
- [x] I have run `make lint` Note: `make lint` requires `python3.11` locally; I ran Ruff on the changed Python files, and all checks passed.
- [x] I have run the relevant tests


ylub:
@joke2k If you like this idea and it is approved for Faker, I would also look into adding other countries and various levels of administrative regions, as they differ from one country to the next, such as states and districts for `de_DE` and `de_AT`, and regions and departments for `fr_FR`.

fcurella:
I'm worried this level of detail, multiplied for each country, would make the size of the package huge. We're already at 2MB as it is. If we really want to go down that road, we should explore solutions to ship data in a compressed format.

ylub:
Yes, I definitely understand the size concern. Do you think this would be better suited as an external provider? Also, for other countries with significantly smaller administrative datasets, would you generally prefer those to live in an external provider as well, or would they be reasonable candidates for inclusion in Faker itself?

For example, there are around 400 districts for the 16 "states" in Germany, compared to over 3,000 counties in the 50 states for the United States (not including territories). I am almost finished with the de_DE version and I would appreciate your guidance.

Thank you!

Issue #1926, opened 2023-10-11
RecursionError: maximum recursion depth exceeded in comparison


jonykalavera:
* Faker version: >= 19.6.2
* OS: Linux

Generating a float with max_value=1.0 and left_digits=1 raises a recursion error. This was not the behavior of previous releases.

### Steps to reproduce

```
>>> from faker import Faker
>>> Faker().pyfloat(left_digits=1, max_value=1.0)
Traceback (most recent call last):
  File "/home/luis/.pyenv/versions/3.10.4/lib/python3.10/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<console>", line 1, in <module>
  File "/home/luis/.cache/pypoetry/virtualenvs/csi-q5090HDc-py3.10/lib/python3.10/site-packages/faker/providers/python/__init__.py", line 217, in pyfloat
    positive_number = self._safe_random_int(
  File "/home/luis/.cache/pypoetry/virtualenvs/csi-q5090HDc-py3.10/lib/python3.10/site-packages/faker/providers/python/__init__.py", line 282, in _safe_random_int
    return self._safe_random_int(orig_min_value, orig_max_value, positive)
  File "/home/luis/.cache/pypoetry/virtualenvs/csi-q5090HDc-py3.10/lib/python3.10/site-packages/faker/providers/python/__init__.py", line 282, in _safe_random_int
    return self._safe_random_int(orig_min_value, orig_max_value, positive)
  File "/home/luis/.cache/pypoetry/virtualenvs/csi-q5090HDc-py3.10/lib/python3.10/site-packages/faker/providers/python/__init__.py", line 282, in _safe_random_int
    return self._safe_random_int(orig_min_value, orig_max_value, positive)
  [Previous line repeated 978 more times]
  File "/home/luis/.cache/pypoetry/virtualenvs/csi-q5090HDc-py3.10/lib/python3.10/site-packages/faker/providers/python/__init__.py", line 279, in _safe_random_int
    min_value = max(min_value, 0)
```

### Expected behavior

Should return a float value  < 1.0

### Actual behavior

Raises RecursionError

jonykalavera:
This PR seems to be the origin of this issue https://github.com/joke2k/faker/pull/1908

jonykalavera:
Wow @fcurella !! you work fast!!

Issue #1152, opened 2020-04-11
Bug in internal method _safe_random_int


anentropic:
I guess maybe it is never called in an unsafe way or this would have come up already, but looking at the code for internal method `_safe_random_int` I think there is a potential problem:

https://github.com/joke2k/faker/blob/0b1f3d8cb14ee2ef051685b06589ee5238e4a084/faker/providers/python/__init__.py#L73
```python
    def _safe_random_int(self, min_value, max_value):
        orig_min_value = min_value
        orig_max_value = max_value

        if min_value is None:
            min_value = max_value - self.random_int()
        if max_value is None:
            max_value = min_value + self.random_int()
        if min_value == max_value:
            return self._safe_random_int(orig_min_value, orig_max_value)
        else:
            return self.random_int(min_value, max_value - 1)
```

It appears from this code that both `min_value` and/or `max_value` could be `None`

But if `max_value` is `None` then this will fail with a `TypeError`:
```python
        if min_value is None:
            min_value = max_value - self.random_int()
```

towr:
Looking at purely the function itself, I think the unlimited recursion when min_value == max_value is worse; it leads to a `RecursionError`.
However, this function only seems to be used in one place, and there checks have been done to ensure they're not both None, or otherwise equal. Adding a docstring with some explanation could help make it clear that's a precondition.

PR #2396, opened 2026-06-25
fix: prevent infinite RecursionError in pyfloat() for sub-integer ranges

### LINKED ISSUE #2082: Intermittent RecursionError in `_safe_random_int`
* Faker version: 26.2.0
* OS: Ubuntu 22.04

I have a Django application that has tests using factory-boy. factory-boy bundles faker so random examples can be generated for various fields on my models. One of my factories uses Faker to generate random floats that are no less than 100:

```python
from factory import Faker

...

historical_min_time_between_samples = Faker("pyfloat", min_value=100)
```

However, this is causing intermittent `RecursionError`s to be raised. Looking into faker's source code, finding [this comment](https://github.com/joke2k/faker/issues/1152#issuecomment-616383723), and using loguru's `logger.catch` on the `pyfloat` function, I believe the issue to be in how `_safe_random_int` handles the case where `max_value` and `min_value` are the same.

I've written the below patch for `_safe_random_int` and can see "Returning min_value to prevent RecursionError: 100" every so often.

```python
def _safe_random_int(self, min_value: float, max_value: float, positive: bool, retry: int = 0) -> int:
    orig_min_value = min_value
    orig_max_value = max_value

    if min_value is None:
        min_value = max_value - self.random_int()
    if max_value is None:
        max_value = min_value + self.random_int()
    if positive:
        min_value = max(min_value, 0)

    if min_value == max_value:
        if retry >= 10:
            print("Returning min_value to prevent RecursionError:", min_value)
            return int(min_value)  # Directly return min_value (or max_value since they are equal) to avoid recursion
        else:
            return self._safe_random_int(orig_min_value, orig_max_value, positive, retry + 1)

    min_value = int(min_value)
    max_value = int(max_value - 1)
    if max_value < min_value:
        max_value += 1
    return self.random_int(min_value, max_value)
```

However, this exposes or introduces another bug in `pyfloat` in the final statements where `result` is set. Specifically, when the recursion block above is hit, `max_value` is None for some reason when evaluating whether `result` ended up > than max_value or < than min_value.

```python
@no_type_check
def pyfloat(
    self,
    left_digits: Optional[int] = None,
    right_digits: Optional[int] = None,
    positive: Optional[bool] = None,
    min_value: Optional[Union[float, int]] = None,
    max_value: Optional[Union[float, int]] = None,
) -> float:
    if left_digits is not None and left_digits < 0:
        raise ValueError("A float number cannot have less than 0 digits in its integer part")
    if right_digits is not None and right_digits < 0:
        raise ValueError("A float number cannot have less than 0 digits in its fractional part")
    if left_digits == 0 and right_digits == 0:
        raise ValueError("A float number cannot have less than 0 digits in total")
    if min_value is not None and max_value is not None:
        if min_value > max_value:
            raise ValueError("Min value cannot be greater than max value")
    if None not in (min_value, max_value) and min_value == max_value:
        raise ValueError("Min and max value cannot be the same")
    if positive and min_value is not None and min_value <= 0:
        raise ValueError("Cannot combine positive=True with negative or zero min_value")
    if left_digits is not None and max_value and math.ceil(math.log10(abs(max_value))) > left_digits:
        raise ValueError("Max value must fit within left digits")
    if left_digits is not None and min_value and math.ceil(math.log10(abs(min_value))) > left_digits:
        raise ValueError("Min value must fit within left digits")

    # Make sure at least either left or right is set
    if left_digits is None and right_digits is None:
        needed_left_digits = max(1, math.ceil(math.log10(max(abs(max_value or 1), abs(min_value or 1)))))
        right_digits = self.random_int(1, sys.float_info.dig - needed_left_digits)

    # If only one side is set, choose #digits for other side
    if (left_digits is None) ^ (right_digits is None):
        if left_digits is None:
            left_digits = max(1, sys.float_info.dig - right_digits)
        else:
            right_digits = max(1, sys.float_info.dig - left_digits)

    # Make sure we don't ask for too many digits!
    if left_digits + right_digits > sys.float_info.dig:
        raise ValueError(
            f"Asking for too many digits ({left_digits} + {right_digits} == {left_digits + right_digits} > "
            f"{sys.float_info.dig})",
        )

    sign = ""
        # Copy values to ensure we're not modifying the original values and thus going out of bounds
        left_min_value = min_value
        left_max_value = max_value
        # Make sure left_digits still respected
        if left_digits is not None:
            if max_value is None:
                left_max_value = 10**left_digits  # minus smallest representable, adjusted later
            if min_value is None:
                left_min_value = -(10**left_digits)  # plus smallest representable, adjusted later

        if max_value is not None and max_value < 0:
            left_max_value += 1  # as the random_int will be generated up to max_value - 1
        if min_value is not None and min_value < 0:
            left_min_value += 1  # as we then append digits after the left_number
        left_number = self._safe_random_int(
            left_min_value,
            left_max_value,
            positive,
        )
    else:
        if positive is None:
            sign = self.random_element(("+", "-"))
        elif positive is True:
            sign = "+"
        else:
            sign = "-"

        left_number = self.random_number(left_digits)

    result = float(f"{sign}{left_number}.{self.random_number(right_digits)}")
    if positive and result == 0:
        if right_digits:
            result = float("0." + "0" * (right_digits - 1) + "1")
        else:
            result += sys.float_info.epsilon

    if right_digits:
        result = min(result, 10**left_digits - float(f'0.{"0" * (right_digits - 1)}1'))
        result = max(result, -(10**left_digits + float(f'0.{"0" * (right_digits - 1)}1')))
    else:
        result = min(result, 10**left_digits - 1)
        result = max(result, -(10**left_digits + 1))

    # It's possible for the result to end up > than max_value or < than min_value
    # When this happens we introduce some variance so we're not always the exactly the min_value or max_value.
    # Which can happen a lot depending on the difference of the values.
    # Ensure the variance is bound by the difference between the max and min
    if max_value is not None:
        if result > max_value:
            if min_value is not None:
                result = result - (result - max_value + self.generator.random.uniform(0, max_value - min_value))
            else:
                result = max_value
    if min_value is not None:
        if result < min_value:
            if max_value is not None:
                result = result + (min_value - result + self.generator.random.uniform(0, max_value - min_value))
            else:
                result = min_value

    return result
```

I know this is not the correct approach to solving this issue as the issue is actually in `pyfloat`, but I unfortunately don't have the time to really sit down and debug this. Hopefully my patches will at least get you in the right direction. :smile_cat: 

For the record, the values that cause the RecursionError from the original code:

```python
left_number = self._safe_random_int(
    left_min_value,  # 100
    left_max_value,  # 100
    positive,  # None
)
```

gaoflow:
## What

`pyfloat(min_value=-1, max_value=0)` (and any range where both bounds truncate to the same integer, such as `(-0.9, 0.1)`) raises an unbounded `RecursionError` at runtime.

## Root cause

`_safe_random_int` adjusts the caller-supplied bounds before selecting an integer.
For a range like `(-1, 0)` the adjustment shifts `min_value` from −1 to 0, leaving
both bounds equal to 0.  The method then recurses with the *original* arguments,
which produce the same collapsed bounds, which recurse again — indefinitely.

The recursive call was intended for the case where *one* bound was `None` (randomly
chosen), so that the retry would pick a different random value.  When *both* bounds
are explicit, retrying always gives the same result.

## Fix

One-line guard: only recurse when at least one original bound was `None`.
If both were explicitly supplied and collapse to the same integer, return that
integer directly and let the downstream clamping in `pyfloat` keep the result
within the requested range.

## Reproducer (before fix)

```python
from faker import Faker
Faker().pyfloat(min_value=-1, max_value=0)  # RecursionError
Faker().pyfloat(min_value=-0.9, max_value=0.1)  # RecursionError
```

## Related issues

Closes #2082, #1926, #1152 (all filed at different times for the same underlying bug, then closed as stale without a code fix).

---

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -264,7 +264,13 @@ class Provider(BaseProvider):
             min_value = max(min_value, 0)
 
         if min_value == max_value:
-            return self._safe_random_int(orig_min_value, orig_max_value, positive)
+            # Only retry if at least one bound was randomly chosen (None originally);
+            # when both bounds were explicitly supplied and still collapse to the same
+            # integer after adjustment, retrying would recurse infinitely with the same
+            # inputs.  Return the collapsed value directly in that case.
+            if orig_min_value is None or orig_max_value is None:
+                return self._safe_random_int(orig_min_value, orig_max_value, positive)
+            return int(min_value)
         else:
             min_value = int(min_value)
             max_value = int(max_value - 1)
--- a/tests/providers/test_python.py
+++ b/tests/providers/test_python.py
@@ -519,6 +519,27 @@ class TestPyfloat(unittest.TestCase):
             self.assertLessEqual(result, 0.1)
             self.assertGreater(result, 0)
 
+    def test_pyfloat_min_neg_one_max_zero(self):
+        """
+        pyfloat(min_value=-1, max_value=0) previously caused an infinite
+        RecursionError inside _safe_random_int when both adjusted bounds
+        collapsed to the same integer (0).  Regression test.
+        """
+        for _ in range(20):
+            result = self.fake.pyfloat(min_value=-1, max_value=0)
+            self.assertGreaterEqual(result, -1)
+            self.assertLessEqual(result, 0)
+
+    def test_pyfloat_min_neg_max_zero_general(self):
+        """
+        Any range where both int-truncated bounds coincide (e.g. (-0.9, 0.1))
+        should not raise RecursionError.
+        """
+        for _ in range(20):
+            result = self.fake.pyfloat(min_value=-0.9, max_value=0.1)
+            self.assertGreaterEqual(result, -0.9)
+            self.assertLessEqual(result, 0.1)
+
 
 class TestPyint(unittest.TestCase):
     def setUp(self):


fcurella (approved):
Thank you!

Issue #2401, opened 2026-06-29
pt_BR: support new alphanumeric CNPJ format (Serpro, effective July 2026)


moohbr:
The pt_BR provider's `cnpj()` and `company_id()` still generate the old numeric-only CNPJ. Brazil's new alphanumeric CNPJ format is not supported. The first 12 characters can now be alphanumeric (`0-9`, `A-Z`); only the 2 check digits remain numeric.

### Steps to reproduce

1. `from faker import Faker`
1. `fake = Faker("pt_BR")`
1. `fake.cnpj()`

### Expected behavior

`cnpj()` should be able to produce alphanumeric CNPJs following the official Serpro algorithm, e.g. `12.ABC.345/01DE-35`. Check digits still computed via mod-11 over each character's value (`ASCII - 48`).

Reference: ["Cálculo dos dígitos verificadores de CNPJ alfanumérico"](https://www.gov.br/receitafederal/pt-br/centrais-de-conteudo/publicacoes/documentos-tecnicos/cnpj)

### Actual behavior

`cnpj()` only ever returns numeric values, e.g. `12.345.678/0001-95`. There is no way to generate the new alphanumeric format.

PR #2402, opened 2026-06-29
feat(pt_BR): support alphanumeric CNPJ via flag

moohbr:
### What does this change

Adds a `use_alphanumeric` flag (default `False`) to `company_id()` and `cnpj()` in the `pt_BR` provider. When enabled, they generate Brazil's new alphanumeric CNPJ: 12 alphanumeric base characters (`0-9`, `A-Z`) plus 2 numeric mod-11 check digits (e.g. `12.ABC.345/01DE-35`). Default behavior is unchanged (numeric, matriz `0001`).

### What was wrong

The `pt_BR` provider could only generate the old numeric-only CNPJ. Brazil's new alphanumeric CNPJ format was not supported, so there was no way to produce valid alphanumeric CNPJs.

### How this fixes it

Introduces a `_company_id_base()` helper that builds either the numeric base (8 random digits + `0001`) or an alphanumeric base (12 chars from `0-9A-Z`). The mod-11 check digits are computed from each character's `ord - 48` value, per the official Serpro algorithm, and validated against its `12ABC34501DE -> 35` example. Existing callers keep the previous output since the flag defaults to `False`. Tests cover both modes and verify the check digits.

Fixes #...

- [x] **If AI tools were used** I have disclosed which ones, and fully reviewed and verified their output.
(used Copilot and some to review PR TEXT, 'cause English isn't my mother language)

### Checklist

- [x ] I have read the documentation about [CONTRIBUTING](https://github.com/joke2k/faker/blob/master/CONTRIBUTING.rst)
- [x ] I have read the documentation about [Coding style](https://github.com/joke2k/faker/blob/master/docs/coding_style.rst)
- [x ] I have run `make lint`


moohbr:
#2401 Solves

tests/providers/test_company.py
```diff
@@ -479,17 +478,33 @@ class TestPtBr:
     def test_company_id_checksum(self):
         assert company_id_checksum([9, 4, 9, 5, 3, 4, 4, 1, 0, 0, 0, 1]) == [5, 1]
         assert company_id_checksum([1, 6, 0, 0, 4, 6, 3, 9, 0, 0, 0, 1]) == [8, 5]
+        assert company_id_checksum([ord(c) - 48 for c in "12ABC34501DE"]) == [3, 5]
 
     def test_company_id(self, faker, num_samples):
+        # https://www.gov.br/receitafederal/pt-br/centrais-de-conteudo/publicacoes/documentos-tecnicos/cnpj/manual-dv-cnpj.pdf/view
```
fcurella: can you move this comment to the provider itself?

fcurella (changes requested):

tests/providers/test_company.py
moohbr: ofc!

moohbr: should this be placed at the module level or in the class or module docstring?

moohbr: I followed the same approach used in `faker/providers/bank/es_AR/__init__.py`

fcurella: you can put it as dosctring for the `cnpj` method

moohbr:
@fcurella done!

fcurella (approved):
Thank you!

moohbr:
Need more something?

Issue #2037, opened 2024-05-01
BE Vat Numbers should have a modulo 97 check and start wit 1 or 0


JorisSpruyt:
* Faker version: 25
* OS: OS. X Monterery

VAT Numbers generated in Belgium (currently set-up in the provider ssn for locale nl_BE should start with 1 or 0 and have a controle number with a module 97 check.

Refer to https://en.wikipedia.org/wiki/VAT_identification_number

### Steps to reproduce

1. Generate vat_id wit locale nl_BE

### Expected behavior

Vat numbers should be starting with BE0 or BE1 and have a module 97 check as the last two numbers

JorisSpruyt:
Suggested re-write:

`def vat_id(self) -> str:

        vat_id_random_section = (
            '#######'
        )

        vat_id_possible_initial_numbers = (
            '0',
            '1'
        )
        """
        http://ec.europa.eu/taxation_customs/vies/faq.html#item_11
        https://en.wikipedia.org/wiki/VAT_identification_number
        :return: A random Belgian VAT ID starting with 0 or 1 and has a correct checksum with a modulo 97 check
        """
        generated_initial_number = self.random_element(vat_id_possible_initial_numbers)
        vat_without_check = self.bothify(generated_initial_number + vat_id_random_section)
        vat_as_int = int(vat_without_check)
        vat_check = 97 - (vat_as_int % 97)
        vat_check_str = f"{vat_check:0>2}"

        return "BE" + vat_without_check + vat_check_str`

stefan6419846:
You are of course invited to propose a corresponding PR to fix this issue.

JorisSpruyt:
Thanks @stefan6419846 , I was not sure about the workflow on github but I think i was able to do it.
https://github.com/joke2k/faker/pull/2038
This issue should be closed as duplicate or something?

Issue #864, opened 2018-10-31
Method .vat_id() is not working as documented


mindflayer:
As I mentiond in the title, `.vat_id()` is not working properly

What I get from the lib:
```>>> import faker
>>> f = faker.Faker(locale='it_IT')
>>> f.vat_id()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'Generator' object has no attribute 'vat_id'
>>> f = faker.Faker(locale='en_GB')
>>> f.vat_id()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'Generator' object has no attribute 'vat_id'
```

What the documentation says:
```
# https://faker.readthedocs.io/en/latest/locales/it_IT.html#faker-providers-ssn
fake.vat_id()
# 'IT32863982442'
# https://faker.readthedocs.io/en/latest/locales/en_GB.html#faker-providers-ssn
fake.vat_id()
# 'GBHA571'
```

fcurella:
Hi @mindflayer 

The `vat_id` method is only available on master, and hasn't been released to PyPI yet.

The issue is that the PyPI page points to the `latest` docs, when it should be pointing to `stable`: https://faker.readthedocs.io/en/stable/locales/it_IT.html#faker-providers-ssn.

PR #2403, opened 2026-06-30
fix(de_DE): compute valid ISO 7064 Mod 11,10 check digit for vat_id (#2399)

### LINKED ISSUE #2399: de_DE vat_id() produces VAT numbers with an invalid checksum (ISO 7064)
### Bug

`Faker('de_DE').vat_id()` returns `DE` + 9 random digits. The German USt-IdNr's last digit is an **ISO 7064 MOD 11,10** check digit, so the generated numbers fail validation ~90% of the time.

### Reproduce

```python
from faker import Faker
from stdnum.de import vat

f = Faker('de_DE')
valid = sum(vat.is_valid(f.vat_id()) for _ in range(1000))
print(valid, "/ 1000 valid")   # ~90 / 1000
```

Example invalid outputs: `DE428071508`, `DE237594599`, `DE466109352`.

### Expected

`vat_id()` for `de_DE` should compute the ISO 7064 MOD 11,10 check digit so the value validates (cf. #2037 for the Belgian VAT mod-97 check, #864).

CedricConday:
## Problem

`Faker('de_DE').vat_id()` returns `DE` + 9 **random** digits. The German USt-IdNr's last digit is an **ISO 7064 Mod 11,10** check digit, so the generated numbers fail validation ~90% of the time (see #2399).

```python
from faker import Faker
from stdnum.de import vat
f = Faker('de_DE')
sum(vat.is_valid(f.vat_id()) for _ in range(1000))   # ~90 / 1000 before this change
```

## Fix

Compute the ISO 7064 Mod 11,10 check digit over the first 8 digits and force a non-zero first digit (the spec — and `python-stdnum` — reject a leading `0`). After this change the same loop returns **1000 / 1000**, verified against `python-stdnum` (2000/2000 over a larger sample locally).

This mirrors the approach used for the Belgian VAT mod-97 check referenced in the issue (#2037, #864).

## Test

Strengthens the existing `TestDeDe.test_vat_id` to verify the check digit (the old assertion `^DE\d{9}$` matched the buggy random output too), recomputing the ISO 7064 Mod 11,10 digit independently — no new test dependency.

Fixes #2399

---
*AI-assisted, human-reviewed — I'm an AI engineer; I find, fix, and test with AI, then review and verify before opening it under my name. The fix was validated against `python-stdnum` as an independent oracle.*

### DIFF
--- a/faker/providers/ssn/de_DE/__init__.py
+++ b/faker/providers/ssn/de_DE/__init__.py
@@ -48,14 +48,26 @@ class Provider(BaseProvider):
         # get the check digit by summing up the digit sums and calculating the modulo of 10
         return str(sum(digit_sums) % 10)
 
+    def __get_vat_checkdigit(self, digits: str) -> str:
+        # ISO 7064 Mod 11,10 check digit for the German USt-IdNr.
+        product = 10
+        for digit in digits:
+            digit_sum = (int(digit) + product) % 10
+            if digit_sum == 0:
+                digit_sum = 10
+            product = (2 * digit_sum) % 11
+        return str((11 - product) % 10)
+
     def vat_id(self) -> str:
         """
         http://ec.europa.eu/taxation_customs/vies/faq.html#item_11
 
-        :return: A random German VAT ID
+        :return: A random German VAT ID with a valid ISO 7064 Mod 11,10 check digit
         """
-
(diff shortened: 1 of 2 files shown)


fcurella (approved):
thank you!

PR #2404, opened 2026-06-30
fix(es_ES): generate valid Spanish CCC control digits in iban() (#2398)

### LINKED ISSUE #2398: es_ES iban() produces IBANs with invalid Spanish national check digits
### Bug

`Faker('es_ES').iban()` returns IBANs whose ISO 13616 (mod-97) check is correct, but whose **Spanish national check digits** (the two control digits of the CCC inside the BBAN) are filled randomly. Strict validators therefore reject them.

### Reproduce

```python
from faker import Faker
from stdnum.es import iban as es_iban

f = Faker('es_ES')
valid = sum(es_iban.is_valid(f.iban()) for _ in range(1000))
print(valid, "/ 1000 valid")   # ~12 / 1000
```

Example output: `ES1266048764759382421948` — passes the IBAN mod-97 check, but `stdnum.es.iban.validate(...)` raises `InvalidChecksum` because the national CCC control digits are wrong.

### Expected

The generated BBAN should also carry valid Spanish CCC control digits, so the IBAN passes country-level validation (same fix that was done for Belgium in #2135 / #2165).

CedricConday:
## Problem

`Faker('es_ES').iban()` returns IBANs whose ISO 13616 (mod-97) check is correct, but whose **Spanish CCC control digits** (the two control digits inside the BBAN) are filled randomly. Strict validators reject them — `stdnum.es.iban` accepts only ~1% (see #2398).

```python
from faker import Faker
from stdnum.es import iban as es_iban
f = Faker('es_ES')
sum(es_iban.is_valid(f.iban()) for _ in range(1000))   # ~12 / 1000 before this change
```

## Fix

Override `bban()` for `es_ES` to compute the two CCC control digits (weights `2**i mod 11`) over `'00' + bank + branch` and over the account number, instead of leaving them random. The generic `iban()` then wraps the BBAN with the correct ISO 13616 check as before.

After this change the same loop returns **1000 / 1000**, verified against `python-stdnum` (2000/2000 over a larger sample locally).

## Test

Strengthens `TestEsEs.test_iban` to recompute and assert the CCC control digits (the previous assertion only checked the ISO mod-97 digits, which the buggy output already passed) — no new test dependency.

Fixes #2398

---
*AI-assisted, human-reviewed — I'm an AI engineer; I find, fix, and test with AI, then review and verify before opening it under my name. Validated against `python-stdnum` as an independent oracle.*

---

### DIFF
--- a/faker/providers/bank/es_ES/__init__.py
+++ b/faker/providers/bank/es_ES/__init__.py
@@ -6,3 +6,19 @@ class Provider(BankProvider):
 
     bban_format = "####################"
     country_code = "ES"
+
+    @staticmethod
+    def _ccc_control_digit(number: str) -> str:
+        # Spanish CCC control digit (weights 2**i mod 11).
+        check = sum(int(n) * 2**i for i, n in enumerate(number)) % 11
+        return str(check if check < 2 else 11 - check)
+
+    def bban(self) -> str:
+        # Spanish CCC: bank (4) + branch (4) + 2 control digits + account (10).
+        # The control digits are computed, not random, so the BBAN passes
+        # country-level validation (ISO 13616 alone does not cover them).
+        bank = self.numerify("####")
+        branch = self.numerify("####")
+        account = self.numerify("##########")
+        control = self._ccc_control_digit("00" + bank + branch) + self._ccc_control_digit(account)
+        return bank + branch + control + account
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

PR #2409, opened 2026-07-01
fix(da_DK): correct BBAN length so iban() generates valid Danish IBANs

CedricConday:
Re-opens #2406, which was closed in error (an automated cleanup pass on my account closed a batch of valid PRs and deleted their branches — this restores the work).

**Problem:** `Faker("da_DK").iban()` produced IBANs that fail checksum/length validation.

**Fix:** correct the `bban_format` so generated IBANs are structurally valid.

**Verified against `python-stdnum` (the reference validator):** 300/300 generated IBANs now pass `stdnum.iban.validate()`. Example: `DK6255575141148089`.

---


fcurella (approved):
Thank you!

PR #2413, opened 2026-07-01
fix(en_IN, es_MX, zh_CN): inherit en_GB bank provider so GB iban() is valid

CedricConday:
These locales fall back to GB IBANs but lacked a correct bank provider, producing invalid IBANs. They now inherit the en_GB bank provider.

---

### DIFF
--- a/faker/providers/bank/__init__.py
+++ b/faker/providers/bank/__init__.py
@@ -163,3 +163,26 @@ class Provider(BaseProvider):
             branch_code = self.lexify("???", letters=string.ascii_uppercase + string.digits)
 
         return bank_code + self.country_code + location_code + branch_code
+
+
+class NonIbanProvider(Provider):
+    """Base bank provider for locales whose country is outside the IBAN system.
+
+    The BBAN/IBAN machinery in the default bank provider is driven by ``country_code`` and
+    ``bban_format``, both of which default to their British values. A locale
+    that inherits the default therefore hands back a British account number,
+    which is worse than no answer at all. Locales in countries that never
+    adopted IBAN inherit from this class instead, so ``bban()`` and ``iban()``
+    raise and only the methods that do apply locally remain available.
+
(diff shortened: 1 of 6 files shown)


fcurella:
these shouldn't inherit from en_GB, but from a more abstract BankIBAN provider. Also, if they don't support IBAn (like China and Mexico), the `iban` should either raise `NotImplemented`, or return en_GB as fallback.

CedricConday:
The three locales in this PR already emit GB IBANs today — the base provider defaults `country_code = "GB"`, so inheriting `en_GB` makes that explicit rather than changing it.

The same holds for `fil_PH`, `fr_CH`, `it_CH`, `ru_RU`, `tl_PH`, and every locale with no bank provider at all: `Faker('en_US').iban()` returns a GB IBAN too.

Happy to add the abstract IBAN provider and have non-IBAN locales raise, but that breaks `iban()` for every locale relying on the current default. Would you want that as its own PR against a major version, with this one closed or narrowed to just the type annotation?

CedricConday:
On reflection you're right that inheriting `en_GB` is the wrong shape even though it preserves current behaviour — it encodes "GB" as a default in places that have nothing to do with the UK.

Happy to do this properly instead: add an abstract bank/IBAN provider holding the format machinery with no country baked in, have the GB-emitting locales inherit from that, and for locales that genuinely have no IBAN system (`zh_CN`, `es_MX`) make `iban()` raise `NotImplementedError` rather than silently returning a British IBAN.

That last part is a behaviour change — today those locales return a GB IBAN, so anyone relying on `iban()` not raising would notice. Do you want that in this PR, or split into one PR for the abstract provider and a second for the raising locales so the breaking part lands on its own?

_AI-assisted, human-reviewed._

CedricConday:
Closing this for now to keep my open PR list tidy — no maintainer bandwidth has come free for it. The locale bank provider inheritance issue is unchanged; happy to reopen and rebase if it is still wanted.

PR #2414, opened 2026-07-01
fix(mk_MK): add bank provider with correct bban_format so iban() is valid

CedricConday:
mk_MK had no bank provider with a valid `bban_format`, so `Faker("mk_MK").iban()` produced invalid Macedonian IBANs. Adds the correct format.


CedricConday:
Closing this — while cleaning up the batch I saw that `mk_MK` now has a bank provider on master (added independently), so this change is redundant. Thanks!

PR #2415, opened 2026-07-01
fix(no_NO): compute Norwegian MOD11 check digit so iban() passes stdnum validation

CedricConday:
Norwegian account numbers end in a MOD11 check digit; the no_NO provider generated it randomly, failing validation. This computes it correctly.

---

### DIFF
--- a/tests/providers/test_bank.py
+++ b/tests/providers/test_bank.py
@@ -409,6 +409,14 @@ class TestNoNo:
             assert iban[:2] == NoNoBankProvider.country_code
             assert re.fullmatch(r"\d{2}\d{11}", iban[2:])
 
+    def test_iban_stdnum(self, faker, num_samples):
+        try:
+            from stdnum import iban as iban_validator
+        except ImportError:
+            pytest.skip("stdnum not available")
+        for _ in range(num_samples):
+            iban_validator.validate(faker.iban())
+
 
 class TestPlPl:
     """Test pl_PL bank provider"""
--- a/tests/providers/test_bank.py
+++ b/tests/providers/test_bank.py
@@ -2,6 +2,8 @@ import re
 
 import pytest
 
+from stdnum import iban as iban_validator
+
 from faker.providers.bank import Provider as BankProvider
 from faker.providers.bank.az_AZ import Provider as AzAzBankProvider
 from faker.providers.bank.cs_CZ import Provider as CsCZBankProvider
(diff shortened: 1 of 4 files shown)


tests/providers/test_bank.py
```diff
@@ -435,6 +409,14 @@ def test_iban(self, faker, num_samples):
             assert iban[:2] == NoNoBankProvider.country_code
             assert re.fullmatch(r"\d{2}\d{11}", iban[2:])
 
+    def test_iban_stdnum(self, faker, num_samples):
+        try:
+            from stdnum import iban as iban_validator
```
fcurella: I'd rather add `stdnum` to the [dev requirements](https://github.com/joke2k/faker/blob/master/dev-requirements.txt) and avoid the `try ... except` block

fcurella (changes requested):

CedricConday:
Done — `python-stdnum` is now in dev-requirements and the import is hoisted to the top of the module, so the IBAN validation test always runs instead of skipping silently when the package is absent.

One packaging note: the distribution is `python-stdnum` while the importable module is `stdnum`, so the requirement line reads `python-stdnum>=1.20`.

CedricConday:
The stdnum dev-dependency and the hoisted import are in — the IBAN validation test now always runs instead of skipping when the package is absent. Re-requesting review.

_AI-assisted, human-reviewed._

CedricConday:
@fcurella — done as suggested: `python-stdnum` is in `dev-requirements.txt` (and `tox.ini`), and the test imports it directly with no `try ... except`. CI is green.

I've also pushed one fix to my own code: the retry loop fell through to `first_10 + "0"` after 100 attempts, which produces exactly the invalid check digit this PR exists to prevent. Draws where the MOD11 remainder is 10 have no single-digit representation and are now simply discarded — the loop is a rejection sampler, so it terminates with probability 1. Re-verified over 20,000 generated BBANs (correct length, format and check digit) and 5,000 IBANs against the ISO 13616 mod-97 checksum.

AI disclosure as per CONTRIBUTING.rst: prepared with Claude Opus 5; all figures above are from an actual run, not asserted.

Anything else you'd like changed here?

CedricConday:
Thanks, @fcurella, and sorry for the slow follow-up. I made that change back in August: `python-stdnum` is in `dev-requirements.txt` and in `tox.ini`, and the test now imports it directly instead of guarding the import. GitHub has marked your thread outdated, so I think the review request is the only thing still holding this one. Would you mind taking another look when you have a moment?

fcurella (approved):
Thank you!

PR #2416, opened 2026-07-01
fix(ru_RU): add correct bban_format so iban() is valid

CedricConday:
ru_RU lacked a correct `bban_format`, producing invalid Russian IBANs. Adds the proper format.

---


fcurella (approved):
Thank you!

PR #2417, opened 2026-07-01
fix(uk_UA): correct bban_format to 6 digits + 19 chars so iban() is valid

CedricConday:
uk_UA used a wrong BBAN length, producing invalid Ukrainian IBANs. Corrects it to 6 bank digits + 19 account characters.

---


fcurella (approved):
Thank you!

PR #2423, opened 2026-07-08
Fix typo: seperator -> separator in file provider

maxtaran2010:
Fix a typo in `faker/providers/file/__init__.py`: the local variable `seperator` was misspelled and should be `separator`.

The variable is used in the `file_path` method to build filesystem paths. The misspelling was cosmetic (it still worked because the value came from `fs_rule["separator"]`), but it's cleaner to use the correct spelling.


fcurella (approved):
Thank you!

PR #2427, opened 2026-07-14
Make uuid1() and uuid7() reproducible under a fixed seed

### LINKED ISSUE #2426: uuid1() and uuid7() are not reproducible under a fixed seed (timestamp taken from time.time())
* Faker version: 40.28.1 (also present on current `master`, bb62df7a)
* OS: macOS 15 (not OS-specific)

The `uuid1()` and `uuid7()` providers added in #2344 build their timestamp bits from `time.time()` (`faker/providers/misc/__init__.py`, lines 198 and 263 on master), so their output changes between runs even with a fixed seed. The other random components are drawn from `self.generator.random`, but the wall-clock timestamp dominates the high bits, and for `uuid1` the 0-999999 ns perturbation is far smaller than the run-to-run clock drift.

### Steps to reproduce

```python
import time
from faker import Faker

def draw(method):
    Faker.seed(0)
    return getattr(Faker(), method)()

a4, a1, a7 = draw("uuid4"), draw("uuid1"), draw("uuid7")
time.sleep(2)
b4, b1, b7 = draw("uuid4"), draw("uuid1"), draw("uuid7")
print("uuid4", a4 == b4)  # True
print("uuid1", a1 == b1)  # False
print("uuid7", a7 == b7)  # False
```

The `uuid1`/`uuid7` values behind those two `False`s, from the same pair of runs:

```
uuid1: run1=c7af58d3-7f1c-11f1-b082-629fd82c07cd
uuid1: run2=c8e18555-7f1c-11f1-b082-629fd82c07cd
uuid7: run1=019f5e12-8d5c-7629-b8f9-c1a0c2094cac
uuid7: run2=019f5e12-9532-7629-b8f9-c1a0c2094cac
```

(`uuid7` can coincidentally match if two runs land in the same millisecond; across runs it drifts, as above.)

### Expected behavior

Per the README's seeding section, "A Seed produces the same result when the same methods with the same version of faker are called." `uuid1()` and `uuid7()` should return the same value on every seeded run, like `uuid4()` does (`test_uuid4_seedability` pins exactly this; the uuid1/uuid7 tests have no seedability counterpart).

### Actual behavior

Same seed, different `uuid1()`/`uuid7()` values on each run, because the timestamp comes from the wall clock rather than the seeded generator.

Happy to send a PR deriving the timestamp from the seeded generator (a `unix_time`-style draw, keeping the version/variant bit layout) and adding `test_uuid1_seedability`/`test_uuid7_seedability` mirroring the uuid4 test.

---

Per the contributing guidelines' AI policy: this report was drafted with AI assistance; I ran the reproduction above myself and verified the cited lines on master before filing.

vidigoat:
### What does this change

`uuid1()` and `uuid7()` now derive their timestamp from the seeded Faker generator instead of `time.time()`, so they return the same value on every run under a fixed seed, matching `uuid4()`. Adds `test_uuid1_seedability` and `test_uuid7_seedability`.

### What was wrong

Both methods built their timestamp bits from the wall clock:

- `uuid1`: `nanoseconds = int(time.time() * 1e9)`
- `uuid7`: `unix_ts_ms = int(time.time() * 1000) + ...`

The wall-clock value dominates the high bits, so the output changes between runs even with a fixed seed. This breaks the seeding guarantee documented in the README ("A Seed produces the same result when the same methods with the same version of faker are called") and contradicts these methods' own docstrings, which state the timestamp is chosen "to ensure seedability". `uuid4()` is reproducible; `uuid1()`/`uuid7()` are not:

```python
import time
from faker import Faker

def draw(m):
    Faker.seed(0)
    return getattr(Faker(), m)()

a1, a7 = draw("uuid1"), draw("uuid7")
time.sleep(2)
b1, b7 = draw("uuid1"), draw("uuid7")
print(a1 == b1, a7 == b7)  # False False on master
```

### How this fixes it

The timestamp is drawn from `self.generator.random` over a fixed range (up to 2035-01-01 UTC) instead of the wall clock. The RFC 4122 (v1) and RFC 9562 (v7) version/variant bit layout is unchanged, so the generated UUIDs are still structurally valid time-based UUIDs — they are now simply reproducible under a fixed seed. `import time` is no longer used and is removed.

Fixes #2426

I used an AI assistant to help draft this description and cross-check the RFC bit layouts. I reviewed and verified the change myself: reproduced the bug on `master`, confirmed the two new tests fail before the fix and pass after, and ran the full test suite (`2393 passed`) plus `make lint` (isort/black 24.4.0/flake8 clean on the changed files; the pre-existing mypy findings in `json()` are unrelated).

### Checklist

- [x] I have run `make lint`

### DIFF
--- a/faker/providers/misc/__init__.py
+++ b/faker/providers/misc/__init__.py
@@ -6,7 +6,6 @@ import os
 import re
 import string
 import tarfile
-import time
 import uuid
 import zipfile
 
@@ -25,6 +24,13 @@ csv.register_dialect("faker-csv", csv.excel, quoting=csv.QUOTE_ALL)  # type: ign
 
 ColumnSpec = Union[Tuple[int, str], Tuple[int, str, Dict[str, Any]]]
 DataColumns = List[ColumnSpec]
+
+# Upper bound for the seeded timestamps used by ``uuid1``/``uuid7``: 2035-01-01 UTC.
+# Drawing the timestamp from this fixed range (rather than the wall clock) keeps the
+# generated UUIDs reproducible under a fixed seed while still spanning plausible dates.
+_MAX_UNIX_TIME_S = 2051222400
+_MAX_UNIX_TIME_MS = _MAX_UNIX_TIME_S * 1_000
+_MAX_UNIX_TIME_NS = _MAX_UNIX_TIME_S * 1_000_000_000
 
 
 class Provider(BaseProvider):
@@ -181,7 +187,7 @@ class Provider(BaseProvider):
         """Generate a random UUID1 (time-based) object and cast it to another type using a callable ``cast_to``.
 
         Uses the Faker random generator for the clock sequence and node fields to ensure seedability,
-        while the timestamp is derived from the current time with random perturbation for uniqueness.
+        while the timestamp is drawn from the seeded generator so the result is reproducible.
 
         By default, ``cast_to`` is set to ``str``.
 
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

PR #2428, opened 2026-07-16
Validate positive=True against a non-positive max_value in pydecimal

Sreekant13:
### What does this change

Adds a validation guard so that `pydecimal(positive=True, ...)` rejects a non-positive `max_value`, instead of silently returning negative values.

### What was wrong

`pydecimal` already rejected `positive=True` combined with a negative or zero `min_value`, but it had no equivalent guard for `max_value`. In the sign logic, the negative sign is selected whenever `max_value <= 0`, and that branch runs before `positive` is taken into account. As a result, `positive=True` was silently ignored:

```python
>>> from faker import Faker
>>> f = Faker()
>>> f.pydecimal(positive=True, max_value=-3)
Decimal('-85374957039153.3649')   # a negative value, despite positive=True
```

### How this fixes it

It adds the symmetric guard directly after the existing `min_value` check: when `positive` is set together with a `max_value` that is not positive, it raises `ValueError("Cannot combine positive=True with negative or zero max_value")`. Valid cases are not affected, for example `pydecimal(positive=True, max_value=5)` and `pydecimal(positive=True)` still return positive values. A regression test is added.

Fixes # (no linked issue; small self-contained bug fix)

- [X] **If AI tools were used**, I have disclosed which ones, and fully reviewed and verified their output.
Claude was used to understand the code structure and review the documentations.

### Checklist

- [x] I have run `make lint`

### DIFF
--- a/faker/providers/python/__init__.py
+++ b/faker/providers/python/__init__.py
@@ -316,6 +316,8 @@ class Provider(BaseProvider):
             raise ValueError("Min and max value cannot be the same")
         if positive and min_value is not None and min_value <= 0:
             raise ValueError("Cannot combine positive=True with negative or zero min_value")
+        if positive and max_value is not None and max_value <= 0:
+            raise ValueError("Cannot combine positive=True with negative or zero max_value")
         if (
             left_digits is not None
             and max_value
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

Sreekant13:
Thanks for reviewing and merging this.

Sreekant13:
Thanks for merging this. Good to have `pydecimal` reject the impossible combination up front rather than quietly returning a negative value.

PR #2431, opened 2026-07-19
fix(ro_RO): generate valid CNP birth dates in ssn()

chuenchen309:
## Problem

`Faker("ro_RO").ssn()` builds the CNP birth-date field with `day = random_int(1, 31)` regardless of the month or leap year, so a small fraction of generated CNPs encode an impossible date (e.g. month 09 day 31, or Feb 30) and are rejected by the independent validator `python-stdnum` (`stdnum.ro.cnp`). The CNP checksum is already computed correctly — only the day is wrong.

```python
from faker import Faker
from stdnum.ro import cnp
f = Faker("ro_RO"); Faker.seed(0)
sum(cnp.is_valid(f.ssn()) for _ in range(20000))   # 19641 / 20000 before this change (359 fail, all "InvalidComponent" dates)
# example bad value: 7240931410866  ->  month=09, day=31
```

## Fix

Clamp the day to the month's length with `calendar.monthrange`, deriving the century from the gender digit — the CNP convention `1/2 → 1900s, 3/4 → 1800s, 5/6 → 2000s` — so February 29 only appears in real leap years. This matches exactly how `stdnum.ro.cnp` decodes the date segment, so validation lines up.

After the change the same loop returns `20000 / 20000`, and `50000 / 50000` across independent seeds.

## Test

Adds `TestRoRO::test_ssn_birth_date_is_valid`, which decodes the `S YY MM DD` digits and asserts they form a real calendar date via `datetime` (no new test dependency). It fails without the fix (`ValueError: day is out of range for month`) and passes with it. Full `tests/providers/test_ssn.py`: 159 passed; `black`/`isort`/`flake8` clean.

---

This PR was authored by an AI coding agent (Claude Code) running on this account: the AI found the bug, ran the repro against `python-stdnum` as an independent oracle, wrote the fix and the test, and wrote this description. The human account holder reviews every change and is accountable for it. The verification above is real and re-runnable from the diff. If this isn't the kind of contribution you want, say so and I'll close it.

### DIFF
--- a/faker/providers/ssn/ro_RO/__init__.py
+++ b/faker/providers/ssn/ro_RO/__init__.py
@@ -1,3 +1,5 @@
+import calendar
+
 from .. import Provider as BaseProvider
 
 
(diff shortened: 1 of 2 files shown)


fcurella (approved):
Thank you!

PR #2439, opened 2026-07-29
test: add boundary tests for text(max_nb_chars)

saimoon504:
Related to #2389

Brief summary: adds three tests to tests/providers/test_lorem.py
pinning down the boundary behavior of text(max_nb_chars).

### What was wrong

The existing lorem tests exercise text() per locale but do not test
its boundary behavior. While testing Faker for a university course we
noticed this gap when investigating #2389.

### How this fixes it

Adds a TestTextBounds class with three tests:

- test_text_respects_upper_bound - output never exceeds max_nb_chars,
  checked around the internal mode switches at 25 and 100 characters
- test_text_minimum_length_raises - values below 5 raise ValueError
  (enforced in the code but previously untested)
- test_text_ext_word_list_only - with ext_word_list, every word in the
  output comes from the given list

All three pass locally: 3 passed in 0.67s on Python 3.10.12.

### DIFF
--- a/tests/providers/test_lorem.py
+++ b/tests/providers/test_lorem.py
@@ -19,6 +19,28 @@ from faker.providers.lorem.uk_UA import Provider as UkUaLoremProvider
 from faker.providers.lorem.vi_VN import Provider as ViVNLoremProvider
 
 
+class TestTextBounds:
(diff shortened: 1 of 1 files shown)


fcurella (commented):
Thank you!

